PR #5, opened 2021-01-18
Accounts list

### LINKED ISSUE #3: Display list of accounts
## Goal

Display list of accounts taken from the extension in an simplified form:
- account name
- account address
- display balances as 0
- no UI design applied

jodator:
Closes #3.

I had some troubles with the setup, namely:

1. The ` global` is not defined for some of the polkadot-js modules (for instance: `global.TextEncoder` polyfill from `@polkadot/x-textencoder`.

    ```html
    <!-- added in index.html -->
    <script>
      var global = window
    </script>
    ```

1. Additionally, I feel that tests are too poor, but at the moment I don't see more to add.

1. The `<SubstrateContextProvider>` is IMO doing too much, so it might require splitting into two. However, it was taken from the substrate-front-end sample. I'd like to refactor it later to have it in line with our approach.


packages/ui/src/providers/substrate/provider.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+let loadAccts = false
+
+const loadAccounts = (state: State, dispatch: Dispatch<Action>) => {
+  const asyncLoadAccounts = async () => {
+    dispatch({ type: 'LOAD_KEYRING' })
+    try {
+      await web3Enable('Pioneer')
+      let allAccounts = await web3Accounts()
```
krzysztof-jelski, line 38: For loading accounts we may want to consider some sort of subscription, as in [apps](https://github.com/polkadot-js/apps/blob/master/packages/react-hooks/src/useAccounts.ts#L21). That way when account is added in the extension it is reflected on the page without the need for reloading. Can come as next PR though.

packages/ui/src/providers/substrate/provider.tsx
```diff
@@ -0,0 +1,77 @@
...
+      allAccounts = allAccounts.map(({ address, meta }) => ({
+        address,
+        meta: { ...meta, name: `${meta.name} (${meta.source})` },
+      }))
+      keyring.loadAll({ isDevelopment: true }, allAccounts)
```
krzysztof-jelski, line 43: I'd love to explore the possibility of having not-global keyring. I believe it should be possible to just inject `new Keyring()` through a context+hook.

packages/ui/src/providers/substrate/provider.tsx
```diff
@@ -0,0 +1,77 @@
...
+  apiPromise.on('connected', () => {
+    dispatch({ type: 'CONNECT', payload: apiPromise })
+    // `ready` event is not emitted upon reconnection and is checked explicitly here.
+    apiPromise.isReady.then(() => dispatch({ type: 'CONNECT_SUCCESS' }))
+  })
+  apiPromise.on('ready', () => dispatch({ type: 'CONNECT_SUCCESS' }))
+  apiPromise.on('error', (err) => dispatch({ type: 'CONNECT_ERROR', payload: err }))
+}
+
+let loadAccts = false
```
krzysztof-jelski, line 31:
```suggestion
let loadAccounts = false
```

krzysztof-jelski (approved):
Looks like a good start.

packages/ui/src/providers/substrate/provider.tsx
jodator: I see if this will be required for #10 as well.

packages/ui/src/providers/substrate/provider.tsx
jodator: To be done in #10

PR #11, opened 2021-01-18
Provide hook api for keyring and accounts

### LINKED ISSUE #10: Provide hook API for loading keyring
As suggested in #5:

> I'd love to explore the possibility of having not-global keyring. I believe it should be possible to just inject new Keyring() through a context+hook.

I think that having this earlier would be beneficial for quicker development of other features. Right not the code is messy.

Additionally take into account other PR comments: https://github.com/Joystream/pioneer/pull/5#discussion_r559578477

jodator:
Closes #10.

Wow, I've removed most of the previous code here.

The thing that is still unresolved is that I can't provide proper keying in `it.skip('Shows loading', () => {` test. maybe I'm doing something wrong here when providing context value. I've skipped that test for now. I suspect that I should better mock the keyring as it _might_ be a singleton there (I didn't test that though).


jodator:
> I suspect that I should better mock the keyring as it _might_ be a singleton there (I didn't test that though).

So it looks like I we can't have two `Keyring` instances because `KeyringOptions` can be initialized only once on page: https://github.com/polkadot-js/ui/blob/83fcb1fa7b516bb110263011e8121eccb735a0d2/packages/ui-keyring/src/options/index.ts#L49

So the only option I see for now is to mock whole `Keyring` or split the tests (however I don't recall how the mocha runner handles multiple tests).

jodator:
Not the best solution, but somewhat working: https://github.com/Joystream/pioneer/compare/provide-hook-api-for-loading-keyring...stub-useaccounts-hook?expand=1 (Still, I might be missing something there on how to test React app).

jodator:
We'll use #12 to close #10.

PR #13, opened 2021-01-20
👛 Display total account balance

jodator:
This PR introduces:

- `useBalances` hook that returns balances for given accounts
- `useApi` hook to interact with API

I've used mock API in tests to mock balances response.


jodator:
@krzysztof-jelski I've updated the test to mock hooks instead of polkadot objects. I've also refactored the `<Accounts>` component a bit since they did not load on gh-pages when localhost was not running.

packages/ui/src/hooks/useBalances.ts
```diff
@@ -0,0 +1,55 @@
...
+import { AccountInfo } from '@polkadot/types/interfaces'
+import { Account } from './useAccounts'
+import { useApi } from './useApi'
+
+interface Balance {
+  total: string
+}
+
+interface AddressToBalanceMap {
+  [key: string]: Balance
```
krzysztof-jelski, line 11: How about introducing `type Address = string`? Could be used here and in the `Account` type as well.

packages/ui/src/pages/Profile/Accounts.tsx
```diff
@@ -17,14 +19,14 @@ export function Accounts({ accounts }: Props) {
...
-        {accounts.map((account) => (
+        {allAccounts.map((account) => (
           <BodyRow key={account.address}>
             <BodyCell>
               <h3>{account.name}</h3>
               <p>{account.address}</p>
             </BodyCell>
             <BodyCell>
-              <p>0 JOY</p>
+              <p>{balances.map[account.address] ? balances.map[account.address].total : '-'}</p>
```
krzysztof-jelski, line 29: This seems to complex here. On a language level I guess we can do sth like
```suggestion
              <p>{balances.map[account.address].total || '-'}</p>
```
But perhaps our hook can return a more convenient object that would handle rendering? Or we could have a small component for balance that would have the view logic.

krzysztof-jelski (approved):

packages/ui/src/pages/Profile/Accounts.tsx
jodator: For sure more atomic components are welcome. However as this is still work in progress I could introduce it later when implementing the rest of balances. I'm also thinking about introducing a component for the whole account as it will have some "complex" logic.

> This seems to complex here. On a language level I guess we can do sth like

👌 I still forgot that I can use this syntax in TS

PR #23, opened 2021-01-21
Account accounts table init layout

ClumsyVlad:


ClumsyVlad:
Check all skipped tests before merging.

PR #43, opened 2021-01-26
💸 Token transfer amount handling

jodator:
In this PR:

- user can select arbitrary value in token transfer modal
  - set half/max balance value
- few UI components were refactored as some were duplicated

However, I still face the problem with the development setup as I wan't able to set working environment other than default substrate build on my Apple M1 :/

### DIFF
--- a/packages/ui/src/components/AccountInfo.tsx
+++ b/packages/ui/src/components/AccountInfo.tsx
@@ -1,0 +1,89 @@
+import { Account } from '../hooks/types'
+import Identicon from '@polkadot/react-identicon'
+import React from 'react'
+import styled from 'styled-components'
+import { BorderRad, Colors } from '../constants'
+import { CopyButton } from './buttons/CopyButton'
+
+export const AccountInfo = ({ account }: { account: Account }) => {
+  return (
+    <AccountInfoWrap>
+      <AccountPhoto>
+        <Identicon size={40} theme={'beachball'} value={account.address} />
+      </AccountPhoto>
+      {/*<AccountType>Root account</AccountType>*/}
+      <AccountName>{account.name}</AccountName>
+      <AccountCopyAddress>
+        <AccountAddress>{account.address}</AccountAddress>
+        <AccountCopyButton />
+      </AccountCopyAddress>
+    </AccountInfoWrap>
+  )
+}
+const AccountInfoWrap = styled.div`
+  display: grid;
+  grid-template-columns: 40px 1fr;
+  grid-template-rows: min-content, 24px 18px;
+  grid-column-gap: 12px;
+  grid-template-areas:
+    'accountphoto accounttype'
+    'accountphoto accountname'
+    'accountphoto accountaddress';
+  align-items: center;
+  width: 100%;
+  justify-self: start;
+`
+const AccountPhoto = styled.div`
+  display: flex;
+  grid-area: accountphoto;
+  justify-content: flex-end;
+  align-items: center;
+  align-content: center;
+  align-self: center;
+  height: 40px;
+  width: 40px;
+  background-color: ${Colors.Blue[500]};
+  border-radius: ${BorderRad.full};
+  overflow: hidden;
+`
+const AccountName = styled.h5`
+  grid-area: accountname;
+  margin: 0;
+  padding: 0;
+  font-size: 16px;
+  line-height: 24px;
+  font-weight: 700;
+  color: ${Colors.Black[900]};
+`
+const AccountCopyAddress = styled.div`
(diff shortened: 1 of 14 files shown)


packages/ui/test/hooks/useNumberInput.test.ts
```diff
@@ -0,0 +1,61 @@
+import { expect } from 'chai'
+import { renderHook, act } from '@testing-library/react-hooks'
+import { useNumberInput } from '../../src/hooks/useNumberInput'
+
+describe('useNumberInput', () => {
+  function render() {
+    let decimals = 18
+    const result = renderHook(() => useNumberInput(decimals))
```
krzysztof-jelski, line 8: Propagate to destructuring, should look cleaner.

packages/ui/src/utils/formatters.ts
```diff
@@ -0,0 +1,11 @@
+import BN from 'bn.js'
+import { formatBalance } from '@polkadot/util'
+
+const decimals = 12
```
krzysztof-jelski, line 4:
```suggestion
const DECIMALS = 12
```
(aka convention to have constants all caps)

packages/ui/src/hooks/useNumberInput.ts
```diff
@@ -0,0 +1,43 @@
...
+  }, [decimals])
+
+  return [value, set] as const
+}
+
+function stripLeadingZeroes(value: string) {
+  if (value === '') {
+    return value
+  }
+  const leadingZeroes = value.match(/^0*/)?.[0].length
```
krzysztof-jelski, line 25:
```suggestion
  const leadingZeroes = value.search(/^0*/) >= 0
```

packages/ui/src/hooks/useNumberInput.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (value.startsWith('.') || value === '') {
+    value = '0' + value
+  }
+  return value
+}
+
+function truncateDecimals(value: string, decimals: number) {
+  if (value.includes('.')) {
+    const [integer, decimal] = value.split('.')
+    const trimmed = decimal.substring(0, decimals)
```
krzysztof-jelski, line 38: Shouldn't we have some rounding here? Eg. trim(1.29, 1) -> 1.3

packages/ui/src/components/AccountInfo.tsx
```diff
@@ -0,0 +1,89 @@
...
+//   width: fit-content;
+//   margin: 0;
+//   padding: 0 8px;
+//   font-size: 10px;
+//   line-height: 16px;
+//   border-radius: 8px;
+//   color: ${Colors.White};
+//   background-color: ${Colors.Blue[200]};
+//   text-transform: uppercase;
+// `
```
krzysztof-jelski, line 89: Remove, I guess

packages/ui/src/components/TransferModal.tsx
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export function TransferModal({ from, to, onClose }: Props) {
+  const { api } = useApi()
+  const { keyring } = useKeyring()
+  const balances = useBalances([from, to])
+  const [isSending, setIsSending] = useState(false)
+  const [amount, setAmount] = useNumberInput(0)
+
+  if (!balances.hasBalances)
```
krzysztof-jelski, line 28: I suggest adding braces.

packages/ui/src/components/TransferModal.tsx
```diff
@@ -0,0 +1,212 @@
...
+  const setHalf = () => setAmount(transferableBalance.div(new BN(2)).toString())
+  const setMax = () => setAmount(transferableBalance.toString())
+  const signAndSend = async () => {
+    setIsSending(true)
+
+    if (!api || !keyring) {
+      return
+    }
+
+    await api.tx.balances
```
krzysztof-jelski, line 50: Consider adding a UI test for this modal. I'm not insisting on it, but could trigger some healthy refactoring (eg. not depending on api directly).

krzysztof-jelski (approved):

packages/ui/src/hooks/useNumberInput.ts
jodator: I've based this on dapp-boilerplate. The number imput is used to validate user input, so rounding might not be desired here.

packages/ui/src/components/TransferModal.tsx
jodator: That makes sense. I'll add them in the next task.

PR #45, opened 2021-01-27
📝 Transaction details

jodator:
At this moment transaction fee is `0` AFAICS.


packages/ui/src/components/TokenValue.tsx
```diff
@@ -0,0 +1,24 @@
...
+    {formatTokenValue(value)}&nbsp;<TokenSymbol>JOY</TokenSymbol>
+  </TokenValueWrap>
+))
+
+const TokenValueWrap = styled.span`
+  font-weight: 700;
+`
+
+const TokenSymbol = styled.span`
+  color: ${Colors.Black['400']};
```
krzysztof-jelski, line 22: Not sure if we plan on having dark/light themes, but I'm used to having colors in the style components theme definition.

krzysztof-jelski (approved):
Read through the code, didn't play it.

packages/ui/src/components/TokenValue.tsx
jodator: That's sound reasonable. I'll track it here: https://github.com/Joystream/pioneer/issues/59.

PR #52, opened 2021-01-28
Drop down select

ClumsyVlad:
Template for drop-down select. Must be edited and redesigned.


jodator:
Waiting for #45 to be merged.

PR #70, opened 2021-02-01
🗃 Choose account transfer view

jodator:


krzysztof-jelski (approved):
Found some minor CSS gaps (some properties are defined twice; IDE complains about that). The account selection list is not scrollable, as well it does not allow to enter arbitrary address, but I guess that is to come in the future. React complains about state updates to unmounted components -- expect it to be fixed with the rxjs PR.

jodator:
> Found some minor CSS gaps (some properties are defined twice; IDE complains about that).

Thanks! I've installed the styled-components support for PHP Storm

> The account selection list is not scrollable, as well it does not allow to enter arbitrary address, but I guess that is to come in the future.

Yep, I've left that as follow-ups for Vlad

> React complains about state updates to unmounted components -- expect it to be fixed with the rxjs PR.

Ouch, I'll review this with the Rx API update.

PR #71, opened 2021-02-01
🐟 Use ApiRx

jodator:


krzysztof-jelski:
This looks definitely better IMO.

packages/ui/src/modals/TransferModal/SignTransferModal.tsx
```diff
@@ -67,14 +60,20 @@ export function SignTransferModal({ onClose, from, amount, to }: Props) {
       onClose()
     }
 
-    if (keyringPair.meta.isInjected) {
-      const { signer } = await web3FromAddress(from.address)
+    let subscription: Subscription
 
-      await submittableExtrinsic.signAndSend(from.address, { signer: signer }, statusCallback).catch(console.error)
+    if (keyringPair.meta.isInjected) {
+      web3FromAddress(from.address).then(({ signer }) => {
```
krzysztof-jelski, line 66: I guess it's still ok to use `await` here.

packages/ui/src/providers/api/context.tsx
```diff
@@ -3,6 +3,5 @@ import { UseApi } from './provider'
 
 export const ApiContext = createContext<UseApi>({
   isConnected: false,
-  state: 'CONNECTING',
```
krzysztof-jelski, line 6: Can't we detect connecting state with RxApi?

packages/ui/src/hooks/useObservable.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function useObservable<T>(observable: Observable<T> | undefined, deps: readonly any[]) {
+  const [data, setData] = useState<T | undefined>(undefined)
+
+  useEffect(() => {
+    const subscription = observable?.subscribe(setData)
+    return () => subscription && subscription.unsubscribe()
+  }, deps)
+
+  return [data]
```
krzysztof-jelski, line 12: Why an array?

jodator: Hmmm, I've copied that part from the other project, but looks like I didn't give a second look at that. I don't see a reason actually.

packages/ui/src/providers/api/context.tsx
jodator: Probably everything between starting the connection and `isConnected` observable emits change is a connecting state. For now, it was not used so I've removed the `state` variable.

packages/ui/src/modals/TransferModal/SignTransferModal.tsx
jodator: I'm not sure if I can return Promise from the `useEffect()` hook callback (using the `async` keyword). I wanted to return the `subscription` callback so I've used the synchronous arrow funciton.

krzysztof-jelski (approved):

packages/ui/src/providers/api/context.tsx
jodator: I'd go with a holistic approach to connection status improvements: https://github.com/Joystream/pioneer/issues/76. For now, I'm OK with a current connection state handling.

krzysztof-jelski: OK

Issue #33, opened 2021-01-22
Token transfer: handle transaction error


jodator:
Handle transaction errors:

- read [transfer events](https://github.com/orgs/Joystream/projects/38) to see if the transaction was successful
- decide if the wait screen should be closed on "finalized" or "inBlock" state
- see potential workflow improvements (deduplicate code for similar action)

PR #78, opened 2021-02-03
🎉 Transaction summary

### LINKED ISSUE #31: Waiting screen
Add screen to be used when:

- waiting for the extension sign
- waiting for the transaction result

### LINKED ISSUE #32: Token transfer: transaction succes
Implement token transfer transaction success modal.

jodator:
Closes #31. Closes #32.

This PR handles:

- waiting screen (for extension signing)
- waiting screen (for the transaction to be in block)
- transaction success with a transaction summary

Transaction failure is not handled as it requires a better understanding of possible results (#33)


packages/ui/src/components/selects/AccountSelectTemplate/OptionAccount.tsx
```diff
@@ -3,8 +3,8 @@ import styled from 'styled-components'
 import { Account } from '../../../hooks/types'
 import { useBalance } from '../../../hooks/useBalance'
 import { BorderRad, Colors, Transitions } from '../../../constants'
+import { BalanceInfo, InfoTitle, InfoValue } from '../../../modals/common'
```
krzysztof-jelski, line 6: Consider setting up IDE or an eslint plugin for import sorting.

krzysztof-jelski (approved):
Looks good. Some random thoughts, not sure if anything should go into this PR:
- `You have just successfully transferred balance from:` feels like missing a `to` somewhere further ;-). Also, perhaps not `balance`, but `funds` or `tokens`
- If I cancel in the extension, the UI still waits
- Adding/deducting the transferred amount from balances ignores the sender's fee. Maybe it's ok.

jodator:
Thanks for tips, I've updated the success screen.

> Adding/deducting the transferred amount from balances ignores the sender's fee. Maybe it's ok.

That's a bug as the fee is `0` but should be incorporated in the amounts. 

> If I cancel in the extension, the UI still waits

This and the above will be fixed by a follow up then.

PR #91, opened 2021-02-05
Transaction success and failure modals, style fixes

ClumsyVlad:


jodator:
@ClumsyVlad could you check also Firefox? I have this result 😢 :

Screenshot 2021-02-05 at 13 53 06

It works OK on Chrome, though :/

The second thing is that the gap here is different than on designs:
Screenshot 2021-02-05 at 14 20 42

PR #159, opened 2021-02-22
🐍 Add query node mock server

### LINKED ISSUE #156: Add Hydra query node mock server
Add hydra mock server to

- get data from the server
- use the mock server in UI tests

jodator:
Closes #156 

The mock server setup includes:

- Miragejs to mock requests responses
- ApolloClient to access query node graphql API
- Initial members queries and mocks – rest will be added when working on particular features

There's some amount of generated code but it looks like it's better than writing them by hand anyway.


packages/ui/webpack.config.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
krzysztof-jelski, line 1: This looks strange here. Typescript rule for js file?

packages/ui/src/mocks/data/mockMembers.ts
```diff
@@ -0,0 +1,11 @@
+import { MemberFieldsFragment } from '../../api/queries'
+import rawMembers from './raw/members.json'
+
+export type MockVideo = MemberFieldsFragment
```
krzysztof-jelski, line 4: This name (`MockVideo`) does not seem right here. `MockMember`?

krzysztof-jelski (approved):

packages/ui/src/mocks/data/mockMembers.ts
jodator: Yep, my bad :+1:

packages/ui/webpack.config.js
jodator: It looks that just my IDE complains about it.

PR #181, opened 2021-03-02
📉 Add stable descending sort

p-sad:


jodator (approved):
It behaves better now 👍 

I've added few things:
- extracted the `filterAccounts()` util as its implementation wasn't needed here and it looks a bit nicer now alongside your util :)
- 0938f658cd8019ec56358ea5b929c58e94ad22f8 There's `String.localCompare()` - I've forgot about that.
- bf0d782053e242ad65f51ba3d6ad8adaf965d6e1 After above we can have order reversing in one place.
BTW, I love the tests here as I could verify quickly my refactorings doesn't break anything

PR #209, opened 2021-03-04
🧑‍⚕️Handle transaction errors

jodator:
Closes #33. Closes #190.

*   I've added error & success checks based on the [events from the blockchain](https://polkadot.js.org/docs/api/examples/promise/transfer-events).
*   The odd thing that bugged me was that extension was not loading the accounts every time (see this PR: #210)

Issue #64, opened 2021-01-29
Subscribe to the extension accounts list


jodator:
The application reads accounts from the extension on the page load. However, it is possible to subscribe to the accounts from the extension. This would enable to update the views dynamically without a need to reload the page.

PR #210, opened 2021-03-04
⏲ Make keys from extension more reliable

jodator:
As described in #209 – the extension was not always available (on Firefox at least) and the keys were not loaded. The library did not detect that extension was loaded later because of a hardcoded check that was evaluated when enabling the extension.

### DIFF
--- a/packages/ui/src/hooks/useAccounts.ts
+++ b/packages/ui/src/hooks/useAccounts.ts
@@ -1,6 +1,6 @@
-import { useEffect, useState } from 'react'
 import { useKeyring } from './useKeyring'
 import { Account } from '../common/types'
+import { useObservable } from './useObservable'
 
 interface UseAccounts {
   allAccounts: Account[]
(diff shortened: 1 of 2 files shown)


packages/ui/src/providers/keyring/provider.tsx
```diff
@@ -15,21 +15,46 @@ function isKeyringLoaded() {
...
+
+  if (!isKeyringLoaded()) {
+    keyring.loadAll({ isDevelopment: true }, allAccounts)
+  }
+}
+
+// Extensions is not always ready on application load, hence the check
+const onExtensionLoaded = (callback: () => void) => () => {
+  const intervalId = setInterval(() => {
+    if (Object.keys((window as any).injectedWeb3).length) {
```
krzysztof-jelski, line 34: Can this check be simplified perhaps?
```suggestion
    if ((window as any).injectedWeb3 !== undefined) {
```

packages/ui/src/providers/keyring/provider.tsx
```diff
@@ -15,21 +15,46 @@ function isKeyringLoaded() {
...
-        if (!isKeyringLoaded()) {
-          keyring.loadAll({ isDevelopment: true }, allAccounts)
-        }
-      })
-  }, [])
+    if (!isLoaded) {
+      return
+    }
+
+    loadKeysFromExtension()
```
krzysztof-jelski, line 56: It's async, so perhaps:
```suggestion
    loadKeysFromExtension().catch(console.error)
```

packages/ui/src/providers/keyring/provider.tsx
jodator: I'll double check but AFAIR the `window.injectedWeb3` object is available immediately and the `Object.keys()` heuristic is used to detect the moment when the extension is ready.  
  
I did not found anything better in the API :/

packages/ui/src/hooks/useAccounts.ts
```diff
@@ -9,25 +9,20 @@ interface UseAccounts {
...
+  }
 
-    return () => {
-      subscription.unsubscribe()
-    }
-  }, [keyring])
+  const hasAccounts = allAccounts.length !== 0
 
-  return state
+  return { allAccounts, hasAccounts }
```
krzysztof-jelski, line 27: I'm a bit concerned that we return `allAccounts`, which a) is not a state and b) is appended with keyring accounts. But apparently the code in this hook is invoked only once, so it works fine...

My expectation would be for `accounts` to be updated whenever something is changed in the extension (eg. an account added or deleted), but looks like it does not work like this.

packages/ui/src/providers/keyring/provider.tsx
krzysztof-jelski: Fair enough

packages/ui/src/hooks/useAccounts.ts
jodator: I'm planning to refactor this together with #64 :) I've already looked at it and it looks like we could use that cheaply.  
I've added this short PR as I saw that, at least on my machine, accounts were not always added from the extension.

krzysztof-jelski (approved):

PR #225, opened 2021-03-05
Members about modal

### LINKED ISSUE #192: Display member details
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=591%3A30154)

### Requiremens

- Display "My Profile" side bar with user's informations.
- Stub data that are not currently fetch from the query-node mock API.
- Stub Accounts / Roles tabs (disable switching between tabs).

ClumsyVlad:
Closes #192


jodator (approved):
Hi, I've made some adjustements, mostly related to code style (shorter imports :wink: ). Other than that, good that you've introduced enums for sizes – it's better to track values instead of plain string. I've changed that to string literal type union because it's more compact and AFAIR there's some issues with string enums.

PR #226, opened 2021-03-08
Allow entering unknown addresses in account select

p-sad:


packages/ui/src/components/selects/SelectAccount/helpers.ts
```diff
@@ -1,7 +1,19 @@
...
   return accounts.filter(
     (item) => item.name?.toLocaleLowerCase().includes(text.toLocaleLowerCase()) || item.address.includes(text)
   )
 }
+
+export function isValidAddress(address: Address, keyring: KeyringInstance | KeyringStruct) {
+  try {
+    keyring.encodeAddress(keyring.decodeAddress(address))
+  } catch (e) {
+    console.log(e)
```
jodator, line 15: We could live without that log.

jodator (approved):

PR #227, opened 2021-03-08
📬 Lay down a skeleton for Transfer Invites modal

p-sad:


packages/ui/src/modals/TransferInviteModal/TransferDetailsModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <Label>From</Label>
+          <SelectMember onChange={stubHandler} enable={!member} selected={member} />
+        </Row>
+        <TransactionAmount>
+          <AmountInputBlock>
+            <Label htmlFor={'amount-input'}>Number of tokens</Label>
+            <NumberInput
+              id="amount-input"
+              value={formatTokenValue(new BN(amount))}
+              onChange={(event) => setAmount(cleanInputValue(event.target.value))}
```
jodator, line 42: Here we have a basic numeric input, so the formatting and even using BN shouldn't be necessary.

If you remove the format/clean loop the value should be OK (numeric like).

jodator: On second though I think that this is OK. However, we could have `cleanInputValue` in the `useNumberInput()` hook itself. Could you refactor this so we don't have `cleanInputValue()` here and in other usages?

jodator (changes requested):
The PR is on point – assuming that validation and button enabling will be done later.

The only thing is to refactor the `useNumberInput()`

PR #254, opened 2021-03-12
🦥 Add accounts tab to member profile

p-sad:


jodator (changes requested):
I think that we could improve this a bit. The thing that stood out for me from the details tab is:

- different padding
- different labels

I think that the below UI would be closer to the details tab:

Screenshot 2021-03-12 at 14 27 39

Could you give it a bit more time to make the UI a bit closer to the above? But don't spend too much time on this. Re-using `<MembershipLabel>` and `<AboutColumn>` should be enough.

Refactoring `<MembershipLabel text="foo" />` to `<MembershipLable>Foo</MembershipLabel>` would be also welcome as a companion refactoring :)

The last thing is that this tab will be displaying accounts by address only, so we need to think of it. ATM, creating `Account` object is enough as this page might be redesigned and it might be changed to something different:

```js
const rootAccount = {
    name: 'Root account',
    address: member.rootAccount,
}
```

jodator (approved):
👍 LGTM

PR #259, opened 2021-03-15
🔦 Member search component

### LINKED ISSUE #202: Add members search to member select component.
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=562%3A27038)

### Requirments

- query Hydra for existing members
- update in a realiable way the list upon responses from the query node

jodator:
Closes #202


jodator:
As for the solution of unifying select's – the good is that we have this logic inside one component as it was duplicated previously. However, I'm not sure if the `renderFoo()` is the best solution for rendering "external" parts. However, I've found ["render props"](https://reactjs.org/docs/render-props.html) pattern described in React docs, so maybe it's not something uncommon.

p-sad (approved):

PR #271, opened 2021-03-18
Add flow for inviting new members

p-sad:


jodator (changes requested):
@p-sad Not a thorough code review, but AFAICS:

1. Something wrong with AvatarURI field – I can't input anything. It looks like either `onchange` event is not attached or the `value` is always empty.
2. The root/controller account should be text only inputs (so like `handle` or `name`) – see [wireframe for that](https://www.figma.com/file/J2KUqGu97lNQC2iKsxJeVH/My-Profile?node-id=192%3A15867).

However, for 2 I like the idea of displaying nice account info if a proper address is found. Let's discuss this on standup. For the PR please change it to text inputs.

The idea is to have `string` as input value, but then create `Account` instance and pass it to other parts of the modal.

PR #286, opened 2021-03-24
🦧 Open Select Membership dialog on login

### LINKED ISSUE #201: Select member on login
Handle log in member flow.

### Requirements

- Display membership switch modal when user reopens app.
- If no memberships for the user do nothing.

@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=994%3A63372)

p-sad:
Closes #201

### DIFF
--- a/packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
+++ b/packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
@@ -18,6 +18,8 @@ export const CurrentMember = () => {
     return <AddMembershipButton>Create a membership</AddMembershipButton>
   }
 
+  !active && !isOpen && toggleOpen()
+
   return (
     <>
       <MembershipsCount />
--- a/packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
+++ b/packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
@@ -13,15 +13,15 @@ export const CurrentMember = () => {
   const { count, members, active, setActive } = useMyMemberships()
   const [isOpen, toggleOpen] = useToggle()
   const [isCreateOpen, toggleCreateOpen] = useToggle()
+  const forceInitialSelect = () => count > 1 && !active && !isOpen && toggleOpen()
   useEffect(() => {
     count == 1 && setActive(members[0])
-  }, [count])
+    forceInitialSelect()
+  }, [count, isOpen])
 
   if (count < 1) {
     return <AddMembershipButton>Create a membership</AddMembershipButton>
   }
-
-  count > 1 && !active && !isOpen && toggleOpen()
 
   return (
     <>
--- a/packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
+++ b/packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import styled from 'styled-components'
 import { BorderRad, Colors, Transitions } from '../../../../constants'
 import { useMyMemberships } from '../../../../hooks/useMyMemberships'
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
```diff
@@ -10,19 +10,24 @@ import { AddMembershipButton } from '../../../membership/AddMembershipButton'
...
   const [isCreateOpen, toggleCreateOpen] = useToggle()
+  useEffect(() => {
+    count == 1 && setActive(members[0])
+  }, [count])
 
   if (count < 1) {
     return <AddMembershipButton>Create a membership</AddMembershipButton>
   }
 
+  count > 1 && !active && !isOpen && toggleOpen()
```
jodator, line 24: This one looks odd. I think that we could do something better:

*   I'd move that out of "render" zone, so right after the last `useEffect()`.
*   The other thing is that I'm not sure if we have to call this on every render – I'm thinking about `useEffect()` hook for that
*   Anything more complex than single `&&` deserves own variable or own `if` especially in "logic" part of code.

packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
```diff
@@ -10,19 +10,24 @@ import { AddMembershipButton } from '../../../membership/AddMembershipButton'
...
+
   return (
     <>
       <MembershipsCount />
       <SwitchMember onClick={toggleOpen}>
-        <MemberInfo member={active || members[0]} isOnDark={true} />
+        {active && <MemberInfo member={active} isOnDark={true} />}
```
jodator, line 30: To shed more light on my previous comment.

This notation is OK as this is both

*   single `&&`
*   and outside "logic" part of the component (layout part)

packages/ui/test/components/CurrentMember.test.tsx
```diff
@@ -51,17 +43,40 @@ describe('UI: CurrentMember component', () => {
...
+      fireEvent.click(button)
+
+      fireEvent.click(within(getByRole('modal')).getByText(/alice_handle/i))
+
+      expect(queryByText(/bob_handle/i)).toBeFalsy()
+      expect(getByText(/alice_handle/i)).toBeDefined()
+    })
+  })
+
+  describe('with one membership', () => {
```
jodator, line 70: Cool :+1:

jodator (changes requested):
Works nicely :) I have only one comment as the check for opening the modal by toggle is a bit too complex. As for the `useEffect()` there — I'm not 100% sure if we must use it there, but I'm feeling anxious when calling state-changing action from within the component itself.

packages/ui/src/components/page/Sidebar/CurrentMember/CurrentMember.tsx
```diff
@@ -13,16 +13,16 @@ export const CurrentMember = () => {
   const { count, members, active, setActive } = useMyMemberships()
   const [isOpen, toggleOpen] = useToggle()
   const [isCreateOpen, toggleCreateOpen] = useToggle()
+  const forceInitialSelect = () => count > 1 && !active && !isOpen && toggleOpen()
```
jodator, line 16:
```suggestion
  const forceInitialSelect = () => {
    if (count > 1 && !active && !isOpen) {
      toggleOpen()
    }
  }
```

jodator: @p-sad I'd go with something like that. The reason here is that a long construct like this is harder to read later on.

Also being more fluent about the intention makes it is easier to quickly grab what's the function is doing. Especially if there are multiple checks before the function call. 

The `toggleOpen()` call itself is "hidden" behind checks.

jodator (changes requested):
One last thing 🙃

jodator (approved):

PR #317, opened 2021-03-25
☔️ Add created members to mock query node

### LINKED ISSUE #166: Fetch created member details from query-node
The member creation process should fetch data from query-node on success.

p-sad:
Closes #166


packages/ui/src/modals/TransferInviteModal/SignTransferModal.tsx
```diff
@@ -13,14 +13,15 @@ import { Text } from '../../components/typography'
...
+import { EventRecord } from '@polkadot/types/interfaces'
 
 interface Props {
   onClose: () => void
   sourceMember: BaseMember
   targetMember: BaseMember
   signer: Account
   amount: BN
-  onDone: (result: boolean, fee: BN) => void
+  onDone: (result: boolean, events: EventRecord[], fee: BN) => void
```
jodator, line 24: I've just noticed that we copy this in multiple places, so it looks like a good candidate to be extracted as own type.

packages/ui/src/modals/AddMembershipModal/AddMembershipModal.tsx
```diff
@@ -38,12 +41,29 @@ export const AddMembershipModal = ({ onClose }: MembershipModalProps) => {
     [JSON.stringify(transactionParams)]
   )
 
-  const onDone = (result: boolean) => setStep(result ? 'SUCCESS' : 'ERROR')
-
   if (step === 'PREPARE' || !transactionParams) {
     return <MembershipFormModal onClose={onClose} onSubmit={onSubmit} membershipPrice={membershipPrice} />
   }
 
+  const onDone = (result: boolean, events: EventRecord[]) => {
```
jodator, line 48: I'm thinkig if we couldn't move that as before. 

I find the seperation of render function better to read and cleaner. The downside is that you have to make add type sefety checks for `transactionParams` here as well.

packages/ui/test/components/SelectAccount.test.tsx
```diff
@@ -67,7 +67,7 @@ describe('UI: SelectAccount component', () => {
       const textBox = getByRole('textbox')
       act(() => {
         fireEvent.change(textBox, { target: { value: 'bob' } })
-        jest.runAllTimers()
+        jest.runOnlyPendingTimers()
```
jodator, line 70: Cool :+1:

packages/ui/src/hooks/useMyMemberships.ts
```diff
@@ -15,7 +15,7 @@ interface UseMembership {
 export function useMyMemberships(): UseMembership {
   const { allAccounts } = useAccounts()
   const addresses = allAccounts.map((account) => account.address)
-  const options = { variables: { rootAccount_in: addresses, controllerAccount_in: addresses } }
+  const options = { variables: { rootAccount_in: addresses, controllerAccount_in: addresses }, pollInterval: 5000 }
```
jodator, line 18: Let's move `5000` to own const. Just to avoid magic numbers.

packages/ui/src/modals/AddMembershipModal/AddMembershipModal.tsx
```diff
@@ -38,12 +41,29 @@ export const AddMembershipModal = ({ onClose }: MembershipModalProps) => {
...
+    const memberId = events.find((event) => event.event.method === 'MemberRegistered')?.event.data[0].toString()
```
jodator, line 49: We have the created memberId so we can use it later in the `SUCESS` step.

This can be separate PR, though as it would require to:

- pass the `memberId` to the success modal
- use that `memberId` to invoke "view my profile" button

jodator: Actually it deserves its own ticket – could you create one?

Tricky part to research: hide current modal and display arbitrary one.

jodator (changes requested):
Look OK :) Two things to add:

1. The duplicated callback might be extracted as TS type
2. Let's keep logic together if possible.

PR #343, opened 2021-04-01
💥 Add modal for failed balance check

### LINKED ISSUE #198: Insufficient funds check for Transfer Invites modal
@[figma](https://www.figma.com/file/J2KUqGu97lNQC2iKsxJeVH/My-Profile?node-id=192%3A17548)

### Requirements

- Validate transaction fees before creating modal
- Display modal with information.

p-sad:
Closes #198

### DIFF
--- a/packages/ui/src/components/TransferButton.tsx
+++ b/packages/ui/src/components/TransferButton.tsx
@@ -2,10 +2,10 @@ import React from 'react'
 import styled from 'styled-components'
 import { Account } from '../common/types'
 import { Colors } from '../constants'
-import { useToggle } from '../hooks/useToggle'
-import { TransferModal } from '../modals/TransferModal/TransferModal'
+import { useModal } from '../hooks/useModal'
+import { TransferModalCall } from '../modals/TransferModal'
 import { ButtonGhost, ButtonPrimary } from './buttons'
-import { ArrowInsideIcon, ArrowOutsideIcon, TransferIcon } from './icons'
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/TransferButton.tsx
```diff
@@ -14,33 +14,38 @@ interface Props {
 }
 
 export function TransferButton({ from, to, disabled }: Props) {
-  const [isOpen, toggleOpen] = useToggle()
+  const { showModal } = useModal()
   const isTransfer = !from && !to
   const isSend = !!from && !isTransfer
-  const icon = isTransfer ? <TransferIcon /> : isSend ? <ArrowOutsideIcon /> : <ArrowInsideIcon />
+  const iconName: TransferIconName = isTransfer ? 'TransferIcon' : isSend ? 'SendIcon' : 'ReceiveIcon'
```
jodator, line 20: Let's extract this logic to have:

```
type TransferType: 'Send' | 'Receive' | 'Transfer'
const getType( from: boolean, to: boolan ) => TransferType
```

Having this we don't have to push `icon` to the modal and we can have the same component that renders different icons depending on the type.

```
<TransferIcon type={getIconType(true, false)} />
```

Another thing is to update the button accordingly.

jodator (changes requested):
The modal looks superb :) After reading the code I came to mind that we don't have to push this icon in the data.
Let's change that.

jodator (approved):

Issue #288, opened 2021-03-25
Overview – working groups


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=199%3A1252)

### Requirements

* Display list of all working groups
*   Display working group details in list:
    *   [x] title, 
    *   [x] description,
    *   [x] number of workers, 
    *   [x] budget, 
    *   [x] working group leader avatar and name

jodator:
Let's pause with this one until we have other data for budget, openings, and leader.

PR #389, opened 2021-04-13
👷 Mock working group data

jodator:
See #288


jodator:
@p-sad Please review changes (maybe commit by commit) to check what we did wrong yesterday. I think that I didn't do something differently here (maybe just using `number` instead of `string` to fetch membership data). If you find something let me know :)

jodator:
@p-sad Please review changes (maybe commit by commit) to check what we did wrong yesterday. I think that I didn't do something differently here (maybe just using `number` instead of `string` to fetch membership data). If you find something let me know :)

p-sad (approved):

Issue #390, opened 2021-04-13
Working groups openings dashboard


jodator:
Requirements:

*   setup dashboard data
*   see if we can fetch data from query-node already

Sub-ticket of: #287

PR #391, opened 2021-04-13
📊 Statistics components

jodator:
Required for #390.

The main change is to make the `<Statistics />` component composable so we're not tied to a top-level interface. The change is to allow developer to compose stats dashboard as:

```tsx
<Statistics>
   <TokenValueStat title="Budget" value="200_000" />
   <StatisticItem title="My Roles">
      <Roles roles={[ 'F', 'BL' ]} />
   <StatisticItem>
</Statistics>
```


jodator:
FYI @ClumsyVlad :point\_up:

p-sad (approved):

Issue #294, opened 2021-03-25
Overview – My applications


jodator:
@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1639%3A28879)

### Requirements

- Setup My Applications tab
- Add "no applications" state
- Display list of applications

PR #421, opened 2021-04-16
💀 Add My Applications skeleton

p-sad:
See #294

### DIFF
--- a/packages/ui/src/working-groups/hooks/useApplications.ts
+++ b/packages/ui/src/working-groups/hooks/useApplications.ts
@@ -1,0 +1,32 @@
+import BN from 'bn.js'
+
+import { useMyMemberships } from '../../memberships/hooks/useMyMemberships'
+import { WorkingGroupOpening } from '../types'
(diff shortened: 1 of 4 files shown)


packages/ui/src/working-groups/hooks/useApplications.ts
```diff
@@ -0,0 +1,32 @@
+import BN from 'bn.js'
+
+import { useMyMemberships } from '../../memberships/hooks/useMyMemberships'
+import { WorkingGroupOpening } from '../types'
+import { WorkingGroupApplication } from '../types/WorkingGroupApplication'
+
+export function useApplications(): WorkingGroupApplication[] {
```
jodator, line 7: Hint: The usage pattern for all query node data will be:

```javascript
const { isLoading, foos } = useFoos()
```

So in the next iteration, you can change that as well.

For the layout setup phase, either is OK, but if you use the `isLoading + returnedData` pattern the fewer changes will be later.

jodator (approved):
@p-sad The PR size and step are on point.

Issue #416, opened 2021-04-16
Stepper component for modals


jodator:
sub-task for #293 

### Requirements:

*   Implement modal with steps containing application steps and role description

PR #423, opened 2021-04-19
🪜 Stepper component

jodator:
See #416

### DIFF
--- a/packages/ui/src/common/components/Stepper.stories.tsx
+++ b/packages/ui/src/common/components/Stepper.stories.tsx
@@ -1,0 +1,27 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+
+import { Modal, ModalBody, ModalHeader } from './Modal'
+import { Stepper } from './Stepper'
+import { TemplateBlock } from './storybookParts/previewStyles'
+
+export default {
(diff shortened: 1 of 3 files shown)


packages/ui/src/common/components/Stepper.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    { title: 'General parameters' },
+    { title: 'Working Group title & limits', isBabyStep: true },
+    { title: 'Starting date & duration', isBabyStep: true },
+    { title: 'Description' },
+    { title: 'Short & opening description', isBabyStep: true },
+    { title: 'Application process', isBabyStep: true },
+    { title: 'Reward' },
+    { title: 'Stake' },
+    { title: 'Application form' },
+  ],
```
p-sad, line 53: It would be cool to have a story with the active step in the middle, so we can see what past steps display like
```suggestion
  ],
  active: 3,
```

p-sad (approved):

packages/ui/src/common/components/Stepper.stories.tsx
jodator: That's a good point, but you can control the steps using "controls":

PR #428, opened 2021-04-20
ℹ️ Add more informative states to MyApplications

### LINKED ISSUE #294: Overview – My applications

p-sad:
Closes #294.

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/MyApplications.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/MyApplications.tsx
@@ -7,10 +7,10 @@ import { PageTitle } from '../../../common/components/page/PageTitle'
 import { Label, TextBig } from '../../../common/components/typography'
 import { ApplicationsList } from '../../../working-groups/components/ApplicationsList'
 import { useMyApplications } from '../../../working-groups/hooks/useMyApplications'
-import { WorkingGroupApplication } from '../../../working-groups/types/WorkingGroupApplication'
 import { AppPage } from '../../components/AppPage'
 
 import { WorkingGroupsTabs } from './components/WorkingGroupsTabs'
+import { isPending } from './helpers'
 
 export const MyApplications = () => {
   const crumbs = useMemo(
@@ -21,9 +21,8 @@ export const MyApplications = () => {
     []
   )
   const { applications, isLoading } = useMyApplications()
-  const isPending = (a: WorkingGroupApplication) => a.status == 'ApplicationStatusPending'
-  const currentApplications = useMemo(() => applications?.filter(isPending), [applications])
-  const pastApplications = useMemo(() => applications?.filter((a) => !isPending(a)), [applications])
+  const currentApplications = useMemo(() => applications?.filter(isPending), [applications, isLoading])
+  const pastApplications = useMemo(() => applications?.filter((a) => !isPending(a)), [applications, isLoading])
   const displayState = () => {
     if (isLoading) return <Loading />
     return applications?.length ? null : <TextBig>No applications found</TextBig>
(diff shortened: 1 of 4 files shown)


packages/ui/src/app/pages/WorkingGroups/MyApplications.tsx
```diff
@@ -18,20 +19,35 @@ export const MyApplications = () => {
...
+  }
 
   return (
     <AppPage crumbs={crumbs}>
       <PageHeader>
         <PageTitle>Working Groups</PageTitle>
         <WorkingGroupsTabs />
       </PageHeader>
-      <TextBig>My Applications</TextBig>
+      <TextBig>{pickText()}</TextBig>
```
jodator, line 37: The "My applications" text was a placeholder to have _something_ on the tab displayed. You can remove it.  
  
For the loading state you can use existing component + have:

```
{loading ? (
	<Loading />
	) : (
	{currentApplications?.length ? (
          <>
            <Label>Current applications</Label>
            <ApplicationsList applications={currentApplications} />
          </>
        ) : null}
	// ... rest
) }
```

packages/ui/src/app/pages/WorkingGroups/MyApplications.tsx
```diff
@@ -18,20 +19,35 @@ export const MyApplications = () => {
...
+            <Label>Current applications</Label>
+            <ApplicationsList applications={currentApplications} />
+          </>
+        ) : null}
+        {pastApplications?.length ? (
+          <>
+            <Label>Past applications</Label>
+            <ApplicationsList applications={pastApplications} />
+          </>
+        ) : null}
```
jodator, line 50:
```suggestion
        {pastApplications?.length && (
          <>
            <Label>Past applications</Label>
            <ApplicationsList applications={pastApplications} />
          </>
        ) }
```

should be shorter, but I don't recall if that would work (guessing on the `null` use).

packages/ui/src/app/pages/WorkingGroups/MyApplications.tsx
```diff
@@ -18,20 +19,35 @@ export const MyApplications = () => {
     ],
     []
   )
-  const { applications } = useMyApplications()
+  const { applications, isLoading } = useMyApplications()
+  const isPending = (a: WorkingGroupApplication) => a.status == 'ApplicationStatusPending'
```
jodator, line 23: Two things:

1.  Short names are cryptic the only good place for them is in for loops or in sort functions IMO. Having `application => application.status === '...'` is not that long :wink:. The other thing is destructuring should work: `( { status } ) => status === '...'`
2.  Re-creating helper functions on every render has its cost related to creating a new Function every time. This allocates memory, takes time, etc. This helper can be safely moved outside the component. This way it will be created once on module read.

packages/ui/src/app/pages/WorkingGroups/MyApplications.tsx
```diff
@@ -18,20 +19,35 @@ export const MyApplications = () => {
...
+  const currentApplications = useMemo(() => applications?.filter(isPending), [applications])
+  const pastApplications = useMemo(() => applications?.filter((a) => !isPending(a)), [applications])
```
jodator, line 25: Keep in mind that React checks the dependency list by reference comparison which will not work with complex objects. So passing an object/array (even with the same values) will cause this to trigger re-render. You can use primitives to prevent that.

For now, we don't have to change that. The `useMyApplications()` uses memo internally.

I'd suspect, however, that the `useMemo()` dependency list inside `useMyApplications()` might be not ideal there.

jodator (changes requested):
As a draft the PR is OK, but I've added some general remarks.

Issue #291, opened 2021-03-25
Working group – History


jodator:
@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1728%3A813)

### Requirements

- Display tab with past openings
- Display tab with past workers
- Display activities

PR #438, opened 2021-04-21
🏺 Add working group openings history

p-sad:
Part 1/3 of #291


packages/ui/src/app/pages/WorkingGroups/WorkingGroup/OpeningsTab.tsx
```diff
@@ -22,6 +23,7 @@ export const OpeningsTab = ({ workingGroup }: Props) => {
   })
   const { member: leader } = useMember(workingGroup.leaderId)
   const { workers } = useWorkers(workingGroup.id ?? '')
+  const currentOpenings = useMemo(() => openings.filter(isOpeningOpen), [isLoading, openings])
```
jodator, line 26: Let's move filtering to the hook by either

1.  Hook parameter `useOpenings( { type: 'PAST' | 'CURRENT' } )`
2.  Different hook, like `useCurrentOpenings()` and `usePastOpenings()`

packages/ui/src/app/pages/WorkingGroups/WorkingGroup/HistoryTab.tsx
```diff
@@ -1,17 +1,40 @@
...
+        <Tabs tabs={tabs} />
+        {currentTab === 'OPENINGS' && <OpeningsHistory groupId={id} />}
+        {currentTab === 'WORKERS' && 'workers'}
+      </MainPanel>
       <SidePanel>side</SidePanel>
     </ContentWithSidepanel>
   )
 }
+
+function OpeningsHistory({ groupId }: { groupId: string }) {
```
jodator, line 36: For consistency - could you keep using arrow functions as components?

jodator (changes requested):
I'd add a few changes:

1. hook or parameter to get upcoming / past openings
2. The "past" opening should not have "Apply now" button.
3. Please link the issue in PR description 😉

packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpenings.tsx
```diff
@@ -85,7 +83,7 @@ export const WorkingGroupsOpenings = () => {
           </Statistics>
           <ContentWithTabs>
             <Tabs tabsSize="xs" tabs={openingsTabs} />
-            {isLoading ? <Loading /> : <OpeningsList openings={activeTab === 'OPENINGS' ? currentOpenings : []} />}
+            {isLoading ? <Loading /> : <OpeningsList openings={activeTab === 'OPENINGS' ? openings : []} />}
```
jodator: You can add past openings here as you've added the API for that.

packages/ui/src/working-groups/hooks/useOpenings.ts
```diff
@@ -1,17 +1,34 @@
+import { useMemo } from 'react'
+
 import { useGetWorkingGroupOpeningsQuery } from '../queries'
-import { asWorkingGroupOpening } from '../types'
+import { asWorkingGroupOpening, WorkingGroupOpening } from '../types'
 
 interface UseOpeningsParams {
   groupId?: string
+  type?: 'open' | 'past'
```
jodator, line 8: I don't recall a part where we would need both, past and open, together. I'd go with `type: 'open' | 'past'`.

packages/ui/src/working-groups/hooks/useOpenings.ts
```diff
@@ -1,17 +1,34 @@
...
+  const typeFilter: (opening: WorkingGroupOpening) => boolean = useMemo(() => {
+    switch (type) {
+      case 'open':
+        return (opening) => isOpeningOpen(opening)
+      case 'past':
+        return (opening) => !isOpeningOpen(opening)
+      default:
+        return () => true
+    }
+  }, [type])
```
jodator, line 23: This looks complex:
- there's `useMemo()` where it rather should be `useCallback()`
- because of the optionality of the `type` you had to use a switch where `default` is not used now.

```suggestion
  const typeFilter: getTypeFilter( type )
```

where:

```ts
const getTypeFilter = (type: 'open' | 'past') => {
  if (type === 'open' ) {
    return isOpeningOpen
  }
  return (opening: WorkingGroupOpening) => !isOpeningOpen(opening)
}
```

jodator: And there the `isOpeningOpen` could be optionally inlined there.

packages/ui/src/working-groups/hooks/useOpenings.ts
```diff
@@ -1,17 +1,34 @@
...
+
   const groups = data?.workingGroupOpenings ?? []
+  const openings = useMemo(() => groups.map(asWorkingGroupOpening).filter(typeFilter), [loading, data])
```
jodator, line 26: After the above, the `typeFIlter` can be moved to the `useMemo()` so it will be invoked only after the data changes.

```suggestion
  const openings = useMemo(() => groups.map(asWorkingGroupOpening).filter(getTypeFiler(type)), [loading, data])
```

jodator (changes requested):
A small set of updates still needed ;)

IMO making `type` optional enforced complex structure of the `typeFilter`.

The second thing is that you missed a part where we could add cheaply "upcoming" openings to the working groups openings tab.

ps.: The tabs name might be confusing so it might be easier to miss that.

PR #450, opened 2021-04-22
⚙️ Add group workers history

p-sad:
See #291


jodator (changes requested):
One, small change because of boolean API.

jodator (approved):
Thanks! 👍

PR #454, opened 2021-04-23
Fraction component, typography fixes

ClumsyVlad:


packages/ui/src/common/components/typography/Fraction.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import styled, { css } from 'styled-components'
+
+import { Colors } from '../../constants'
+
+import { TextInlineMedium } from './Text'
+
+interface FractionProps {
+  num: string | number
+  den: string | number
```
jodator, line 10: Those props are too cryptic. Having full names in props helps to understand what they are for.

Here, it should be rather: `numerator`  & `denominator` as you've named components.

packages/ui/src/common/components/typography/Fraction.tsx
```diff
@@ -0,0 +1,48 @@
...
+  className?: string
+  sameSize?: boolean
+}
+
+export const Fraction = React.memo(({ num, den, className, sameSize }: FractionProps) => {
+  return (
+    <FractionBlock className={className} sameSize={sameSize} lighter value>
+      <FractionNumerator>{num ? num : 0}</FractionNumerator>
```
jodator, line 18:
```suggestion
      <FractionNumerator>{num ?? 0}</FractionNumerator>
```

For undefined check, you can use nullish coalescing operator `??` - does the same but nicer syntax :)

packages/ui/src/common/components/typography/Fraction.tsx
```diff
@@ -0,0 +1,48 @@
...
+      {!sameSize && <span> </span>}/{!sameSize && <span> </span>}
+      <FractionDenominator>{den ? den : 0}</FractionDenominator>
```
jodator, line 20:
```suggestion
      <FractionDenominator>{den ?? 0}</FractionDenominator>
```

As above :point\_up:

jodator (changes requested):
Component looks OK 😄 

I've added some code-style remarks to update.

jodator (approved):

PR #464, opened 2021-04-27
🧐 Storybook fix

jodator:


jodator:
@thesan FYI: 3de3af0 - when adding SVG as a React component you have to use camel case for properties. I've missed that as both will work (when not using camel case React will complain in console).

PR #465, opened 2021-04-27
👤 UnknownAccountInfo component

p-sad:


jodator:
@p-sad Looks OK. Please take a look at the tests I've added. The change from other tests in the project is that I've used context provider instead of mocking the `useAccounts()` hook. I've previously abused hook mocks. After changes in some of the hooks, we should be able to use a context provider for mocking hook data.

packages/ui/src/accounts/components/UnknownAccountInfo.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+import { useAccounts } from '../hooks/useAccounts'
+import { accountOrNamed } from '../model/accountOrNamed'
+
+import { AccountInfo } from './AccountInfo'
+
+interface Props {
+  address: string
```
jodator:
```suggestion
  address: Address
```

For some basic types, we should use better typing. The address has its own defined type.

jodator (approved):
Other than that one comment, LGTM 👍

Issue #432, opened 2021-04-20
Members list


jodator:
## Wireframe / design

![](https://user-images.githubusercontent.com/247363/115417604-094e7f00-a1f9-11eb-9562-f9e04dc45e57.png)

## Requirements

*   Create members list component together with an accompanying storybook story.
*   Assume all fields came from query-node GraphQL.
*   Click on the member profile should open the existing modal component.
*   Members list should display paginated results.
*   Create react hook or ask for assistance if you're not experienced with writing custom hooks.

## Technical details

*   Most of the building blocks components are available:
    *   List wireframe
    *   list heading (with sorting)
    *   MemberInfo with avatar and icons
    *   Member roles
    *   Token (JOY) value component
*   The list requires integration with Apollo Client `useQuery` hook to display paginated results. In later steps sorting and filtering will be required.
*   To open member details use `useModal()` hook:

```typescript
import { useModal } from '../../../common/hooks/useModal'
import { MemberModalCall } from '../../components/MemberProfile'

showModal<MemberModalCall>({ modal: 'Member', data: { id: memberId } })
```

jodator:
Closed by [`3665ff0`](https://github.com/Joystream/pioneer/commit/3665ff08842cf445ab14d6a59c08c04ea9025266)

PR #468, opened 2021-04-27
🔩 Create a MemberList component

thesan:
Only adds the component to Storybooks for now
Part of #432


packages/ui/src/memberships/components/MemberListItem/MemberListItem.tsx
```diff
@@ -23,43 +25,77 @@ export const MemberListItem = ({ member }: Props) => {
...
       <MemberColumn>
-        <MemberInfo member={member} onClick={showMemberModal} showId />
+        <MemberInfo member={member} onClick={showMemberModal} showId={member.type !== 'Members'} />
       </MemberColumn>
+
+      {member.type === 'Members' && (
+        <MemberColumn>
+          <Info>{member.isConcilMember ? 'YES' : 'NO'}</Info>
+        </MemberColumn>
+      )}
```
jodator, line 43: Introducing multiple type checks may indicate that this component might be a new one. And, I think that this should be true also here. The reason is that it starts getting complicated to support different types in one component.

I think that it would be better to revert the changes made here and introduce a new component for memberships lists.

To help you with naming, as you might not be accustomed to the projects domain:

*   the existing one can be renamed to `<MyMembers />`  as it is used to displays user's memberships (members)
*   the new one can be named simply `<Members />` as it displays a list of all of the Pioneer's users

packages/ui/src/memberships/types/memberships.ts
```diff
@@ -27,7 +30,21 @@ export interface Member {
...
+  type: 'Members'
+  isConcilMember: boolean
+  totalBalanced: number
+  totalStacked: number
+}
+
+export type Member = Membership | MMember
+export type Members = Membership[] | MMember[]
+
+export interface DetailedMember extends Membership {
```
jodator, line 47: Initially, I wanted to push some changes to those types but after second thought I think that this change should also be reverted.

Sorry for that but the domain behind those fields wasn't properly described. You can assume that:

*   there will be a hook for getting member's balance and stake, you can use simply `0` for those values in the component for now
*   the same is with `isCouncil` I'm not 100% sure, but it looks like that part of the Joystream GraphQL queries was not yet designed. Having in mind that most members will not be council members please also use "no" here in the list.

packages/ui/src/app/App.tsx
```diff
@@ -29,7 +29,7 @@ export const App = () => (
         <Route exact path="/working-groups/my-applications" component={MyApplications} />
         <Route exact path="/working-groups/my-roles" component={MyRoles} />
         <Route exact path="/working-groups/grouppreview/:id" component={WorkingGroup} />
-        <Route exact path="/members" component={MemberList} />
+        <Route exact path="/members" component={Members} />
```
jodator, line 32: :+1:

packages/ui/src/memberships/components/MemberList/MemberList.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const roles = 200
+  const count = type === 'Members' ? 20 : 76
+  const total = 100
+
+  return type === 'Members'
+    ? `${id}px ${name}px ${concil}px ${roles}px ${count}px ${count}px ${total}px ${total}px`
+    : `${name}px ${roles}px ${count}px ${count}px 96px 76px 54px`
+}
+
+export const MemberList: FC<MembersSectionProps> = ({ type, members }) => (
```
jodator, line 31: Minor thing – we prefer the below notation as it is consistent with the rest of the code base :)

```suggestion
export const MemberList = ({ type, members }: MembersSectionProps) => (
```

jodator (changes requested):
Hi @thesan! Overall I like that you've created stories for both types of members lists. Both capture what the lists should look like :)

The domain for memberships might be hard to understand from the wireframes only, so all my comments reflect that. One Pioneer user can have multiple memberships (thus multiple members) which he or she can use to act at various parts of the system. Those are already implemented in thy "My Profile -> My Memberships" section.

Having this in mind I would like to ask you to:

- revert changes to the types system
- split the created members list into two (but re-using base building blocks if possible – similarly to what you did with ListsHeaders 👍 )

packages/ui/src/memberships/components/MemberListItem/Member.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <MemberColumn>
+        <Info>#{member.id}</Info>
+      </MemberColumn>
+
+      <MemberColumn>
+        <MemberInfo member={member} onClick={showMemberModal} showGroup={false} />
+      </MemberColumn>
+
+      <MemberColumn>
+        <Info>{member.isConcilMember ? 'YES' : 'NO'}</Info>
```
jodator, line 34: You can remove `member.isConcilMember` and always show "NO".

packages/ui/src/memberships/components/MemberListItem/Member.tsx
```diff
@@ -0,0 +1,55 @@
...
+      </MemberRolesColumn>
+      <MemberColumn>
+        <CountInfo count={0} />
+      </MemberColumn>
+      <MemberColumn>
+        <CountInfo count={0} />
+      </MemberColumn>
+
+      <MemberColumn>
+        <TokenValue value={member.totalBalance ?? 0} />
```
jodator, line 48: At this stage we can assume that this might come from another source, so please use here:

```suggestion
        <TokenValue value={0} />
```

packages/ui/src/memberships/components/MemberListItem/Member.tsx
```diff
@@ -0,0 +1,55 @@
...
+      </MemberColumn>
+      <MemberColumn>
+        <TokenValue value={member.totalStacked ?? 0} />
```
jodator, line 51:
```suggestion
        <TokenValue value={0} />
```

Similarly to the total balance, we can assume that this will come from another source, so for now using `0` will be OK.

packages/ui/src/memberships/components/MemberListItem/Fileds.tsx
```diff
@@ -0,0 +1,58 @@
...
+  display: grid;
+  grid-template-columns: ${({ member }: { member: Membership | OtherMember }) => colLayoutByType(member.kind)};
+  grid-template-rows: 1fr;
+  justify-content: space-between;
+  justify-items: start;
+  align-items: center;
+  width: 100%;
+  height: ${Sizes.accountHeight};
+  padding: 16px 0 16px 14px;
+`
```
jodator, line 41: Instead of adding a type for this, we can have either:

*   two components for this or
*   base component and two extensions for each list type which will define the gaps

This will duplicate some code but would remove a need to have two types, that are not necessary.

I'd go with `MemberItemWrap`:

```
const MemberItemWrap = styled.div`
  display: grid;
  grid-template-rows: 1fr;
  justify-content: space-between;
  justify-items: start;
  align-items: center;
  width: 100%;
  height: ${Sizes.accountHeight};
  padding: 16px 0 16px 14px;
`
const MyMemberItemWrap = styled(MemberItemWrap)`
   grid-template-columns: colLayoutByType('MyMember')
`
const MemberItemWrap = styled(MemberItemWrap)`
   grid-template-columns: colLayoutByType('Member')
`
```

packages/ui/src/memberships/components/MemberListItem/Fileds.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { Fonts, Sizes } from '../../../common/constants'
+import { Membership, OtherMember } from '../../types'
+
+export const colLayoutByType = (kind: (Membership | OtherMember)['kind']) => {
```
jodator, line 7:
```suggestion
export const colLayoutByType = (kind: 'Member' | 'MyMember') => {
```

This to be in line with the proposed changes below :point\_down:

packages/ui/src/memberships/components/MemberInfo.tsx
```diff
@@ -29,10 +29,21 @@ interface MemberInfoContainerProps {
   maxRoles?: number
 }
 
-export type MemberInfoProps = MemberInfoContainerProps & MemberInfoWrapProps
+export type MemberInfoProps = MemberInfoContainerProps & MemberInfoWrapProps & { showGroup?: boolean }
```
jodator, line 32: :+1:

jodator (changes requested):
The changes are in a good direction :)

The last thing that I will insist on here is to keep the current `Member` type unchanged. 

The rationale behind this is that we don't have those values yet and we shouldn't add them if not needed. For instance, the balances will probably come from the blockchain, not from the GraphQL query. We can safely assume that the same type will be used for both lists and the remaining fields will be added from other sources.

packages/ui/src/memberships/components/MemberListItem/Fileds.tsx
thesan: I ended up adding a kind prop like you suggested for the colLayoutByType function

jodator (approved):

PR #483, opened 2021-04-29
⇅ Add members sorting

### LINKED ISSUE #433: Add members sorting
## Requirements

*   Add sorting mechanism (by changing query parameters) to add sorting by:
    *   member id
    *   member handle
    *   is council member
    *   number of slashed times
    *   number of terminated times
    *   total balance
    *   total staked
    *   order by active roles is to be defined (but might not be needed)
*   Changing sort & order should reset to the first page of the results.

## Technical details

thesan:
Should close #433

I followed the `packages/ui/src/common/api/schemas/schema.graphql` but Apollo generates the `GetMembersQueryVariables` type with `orderBy` at the root (along with `rootAccount_in` and `controllerAccount_in`)


jodator (approved):
Thanks @thesan! I've made only one change to your PR.

I've extracted the reducer function outside the component so it will not be re-created each time the component is used (d5a523ec96ddc127cf51362cc81accc49e1545ec).

Moving it outside the `<Members />` component function will create this function once.

Issue #295, opened 2021-03-25
Application preview


jodator:
@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4830607)

### Requirements

- Display application preview general details under various stages of an application life-cycle
- Display application preivew form

PR #489, opened 2021-04-29
🌵 Display form answers in application details

p-sad:
See #295

### DIFF
--- a/packages/ui/src/common/api/schemas/schema.graphql
+++ b/packages/ui/src/common/api/schemas/schema.graphql
@@ -540,6 +540,10 @@ input MembershipWhereUniqueInput {
   handle: String
 }
 
+input ApplicationFormQuestionAnswerWhereInput {
+  application_eq: ID
+}
+
 enum Network {
   BABYLON
   ALEXANDRIA
@@ -578,6 +582,7 @@ type Query {
   workingGroupOpenings(offset: Int, limit: Int = 50, where: WorkingGroupOpeningsWhereInput): [WorkingGroupOpening!]
   workingGroupOpening(where: WorkingGroupOpeningWhereUniqueInput): WorkingGroupOpening
   workingGroupApplications(offset: Int, limit: Int = 50, where: WorkingGroupApplicationsWhereInput): [WorkingGroupApplication!]
+  applicationFormQuestionAnswers(offset: Int, limit: Int = 50, where: ApplicationFormQuestionAnswerWhereInput): [ApplicationFormQuestionAnswer!]
 }
 
 enum EventType {
--- a/packages/ui/src/common/api/schemas/schema.graphql
+++ b/packages/ui/src/common/api/schemas/schema.graphql
@@ -817,6 +817,7 @@ enum ApplicationFormQuestionType {
 }
 
 type ApplicationFormQuestion {
+  id: ID!
   "Related opening metadata"
   openingMetadata: WorkingGroupOpeningMetadata!
   "The question itself"
--- a/packages/ui/src/common/api/schemas/schema.graphql
+++ b/packages/ui/src/common/api/schemas/schema.graphql
@@ -516,7 +516,7 @@ input MembershipWhereInput {
 }
 
 input WorkingGroupOpeningsWhereInput {
-  group_eq: ID
+  groupId_eq: ID
 }
 
 input WorkingGroupApplicationsWhereInput {
@@ -524,7 +524,7 @@ input WorkingGroupApplicationsWhereInput {
 }
 
 input WorkersWhereInput {
-  group_eq: ID
+  groupId_eq: ID
 }
 
 input WorkingGroupWhereUniqueInput {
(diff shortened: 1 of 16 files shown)


packages/ui/src/common/api/schemas/schema.graphql
```diff
@@ -817,6 +822,7 @@ enum ApplicationFormQuestionType {
 }
 
 type ApplicationFormQuestion {
+  id: ID!
```
p-sad, line 825: Added this ID so that I could seed the mocked data more easily, but I see that it's still missing in the schema in the newest Olympia PR. Not sure why some types don't have IDs

jodator: Maybe not all of them are required for the Hydra to create GraphQL schema. The one in the PR is later transformed by Hydra to the GraphQL format. My second guess is that only specific IDs might need to be defined.

packages/ui/src/common/api/schemas/schema.graphql
```diff
@@ -540,6 +540,10 @@ input MembershipWhereUniqueInput {
   handle: String
 }
 
+input ApplicationFormQuestionAnswerWhereInput {
+  application_eq: ID
```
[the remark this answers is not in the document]
jodator: Could you double-check?

packages/ui/src/mocks/data/raw/applications.json
```diff
@@ -6,7 +6,16 @@
...
+    "answers": [
+      {
+        "questionId": "0",
+        "answer": "18"
+      },
+      {
+        "questionId": "1",
+        "answer": "I am the smartest"
+      }
+    ],
```
jodator, line 18: Let's review it after the schema upgrade. I worry that the mapping there is different but let's keep that in mind.

jodator: Or event better -> please create a follow-up to check this and assign yourself to it :)

packages/ui/src/working-groups/modals/ApplicationDetailsModal/FormDetails.tsx
```diff
@@ -0,0 +1,43 @@
...
+    return (
+      <SidePaneTable>
+        <SidePaneText>Loading...</SidePaneText>
+      </SidePaneTable>
+    )
+
+  return (
+    <SidePaneTable>
+      {answers?.map((answer) => (
+        <QuestionAnswerPair answer={answer} key={JSON.stringify(answer)} />
```
jodator, line 30: Maybe a better would be `question.index` here instead of `JSON.stringify()`. The `index` should uniquely identify questions from the set so it falls into `key` constraints. The other thing is that, although negligible, there is a cost for JSON.stringify compared to property access.

packages/ui/src/working-groups/modals/ApplicationDetailsModal/FormDetails.tsx
```diff
@@ -0,0 +1,43 @@
...
+  applicationId: string
+}
+
+export const FormDetails = React.memo(({ applicationId }: Props) => {
+  const { isLoading, answers } = useApplicationQuestionAnswers(applicationId)
+
+  if (isLoading)
+    return (
+      <SidePaneTable>
+        <SidePaneText>Loading...</SidePaneText>
```
jodator, line 23: Please use `<Loading />` component. There a chance that we add something fancy for loaders and then we will have one place to change :)

packages/ui/src/working-groups/types/ApplicationFormQuestionAnswer.ts
```diff
@@ -0,0 +1,13 @@
+import { ApplicationFormQuestionAnswerFieldsFragment } from '../queries'
+
+export interface ApplicationFormQuestionAnswer {
```
jodator, line 3: WDYT about `QuestionAnswer`? I like descriptive types but maybe we will not have many question answers.

Or maybe `ApplicationQuestion` and `ApplicationQuestionAnswer` -> hence vomiting `Form` there.

jodator: However, if you're fine with `ApplicationFormQuestionAnswer` you can leave it as is :)

jodator (changes requested):
Awesome 🤩 There are some small things to polish, but overall good job.

packages/ui/src/working-groups/modals/ApplicationDetailsModal/FormDetails.tsx
p-sad: I'm wondering if we shouldn't also sort them by index before rendering, to make sure they show up in the same order as  they appeared on the form

packages/ui/src/working-groups/types/ApplicationFormQuestionAnswer.ts
p-sad: It's actually a scalding satire of java enterprise
I dropped the "Form", I feel like shortening it further either makes it too generic/non-descriptive or makes it linguistically unsound

packages/ui/src/working-groups/modals/ApplicationDetailsModal/FormDetails.tsx
jodator: Yep, we can do that as well. I've done that in the application form for questions.

PR #491, opened 2021-04-29
Roles preslice

### LINKED ISSUE #481: Roles Components
Prepare the [membership role component](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1866%3A1011).

ClumsyVlad:
Closes #481


packages/ui/src/memberships/components/MemberProfile/MemberDetails.tsx
```diff
@@ -44,10 +44,12 @@ export const MemberDetails = React.memo(({ member }: Props) => {
...
   return (
     <SidePaneTable>
-      <SidePaneColumn>
-        <SidePaneLabel text="About" />
-        <SidePaneText>{memberDetails?.about || ''}</SidePaneText>
-      </SidePaneColumn>
+      {memberDetails?.about && (
+        <SidePaneColumn>
+          <SidePaneLabel text="About" />
+          <SidePaneText>{memberDetails?.about}</SidePaneText>
```
jodator, line 50:
```suggestion
          <SidePaneText>{memberDetails.about}</SidePaneText>
```

packages/ui/src/memberships/components/MemberProfile/MemberRoles.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  </ButtonsGroup>
+                </MemberRoleTable>
+              </MemberRoleTableContainer>
+            )}
+          </MemberRoleToggleContainer>
+        )
+      }}
+    </MemberRoleToggle>
+  )
+}
```
jodator, line 89: Could you extract this into a separate file and create a storybook story for it?

We can use that later to validate if the data we pass there is OK when we will implement member roles.

packages/ui/src/memberships/components/MemberProfile/MemberDetails.tsx
jodator: The check above (`memberDetails?.about &&`) should tell TypeScript that `.about` is not `undefined`.

jodator (changes requested):
Nice :) The component you've added is exactly how I would imagine it.

I have only one request about adding a story for the member role component.

jodator (approved):

Issue #292, opened 2021-03-25
Open position preview


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=259%3A1652)

### Requirements

- Prepare opening page
- Implment opening dashboard
- Display opnening description
- Display opening application process
- Display list of candidates
- DIsplay different state of application (upcoming, etc)

### Questions

- Do applicaiton process is text-only or follows a protobuf schema?

jodator:
@wyrdrender to help you out with the domain requirements for the opening screens we need. I've made a list below to track everything. It looks like almost every state of the opening has a bit different set of components.

1.  Upcoming opening (this one requires mocks, but let's wait for #548 PR as it should ease with mocks creation).
    *   [x] Notify button (no action for now)
    *   [ ] Empty applicants list (this opening hasn't started)
    *   [ ] dashboard - I think that it should have: 
        *   Time to start,
        *   Expected duration,
        *   reward per 3600 blocks
        *   applicants + hiring limit
    *   [ ] created at block time
2.  Opened (active) opening
    *   [x] copy/apply now
    *   [ ] when no applicants (no applicants info + apply now)
    *   [x] applicants list
    *   [ ] dashboard:
        *   time left
        *   minimal stake or my stake (when my application)
        *   reward per blocks
        *   current applicants + hiring
    *   [ ] created at block time
3.  Closed / cancelled
    *   [x] copy link
    *   [x] applicants list 
    *   [ ] info hiring complete if there's my application (currently it is always shown)
    *   [ ] dashboard:
        *   (first ?? -> I'll check that one),
        *   reward per blocks
        *   minimal stake or my stake (if my application)
        *   applicants + hired
    *   [ ] closed at block time

General information:

*   [x] badges
*   [ ] short description (AFAIR text-like)
*   [x] description (Markdown like)
*   [ ] application process (`details` field) (Markdown like)

jodator:
I think that we can close this one. The milestone needs a review anyway, so the remaining things are moved to: #647

PR #496, opened 2021-04-29
Open position preview

wyrdrender:
See #292

The basic layout of page was made, including new type of statistic component.

TODO:

- Integration with Opening updated model
- Display opnening description
- Display list of candidates
- DIsplay different state of application (upcoming, etc)
- Information about not started opening


jodator (approved):
I saw that you've fixed the storybook, thanks! ATM keeping webpack config in two places is annoying as I stumbled upon this error a few times already.

I think that we can fix that next time we will touch the webpack config.

packages/ui/src/common/components/BadgeViolet.ts
```diff
@@ -22,6 +24,20 @@ export const BadgeViolet = styled.span<BadgeVioletProps>`
...
+      margin: ${spacing(0, 0.5)};
+
+      &:last-child {
+        margin-right: 0;
+      }
+
+      &:first-child {
+        margin-left: 0;
+      }
+    `};
```
jodator, line 39: I think that I would go with adding dedicated parent and put CSS rules there. My thinking here is that we don't necessarily need to add a property for that.

Something like below:

```tsx
const BadgeRow = styled(Row)`
  ${BadgeViolet} {
    margin: ...
  }
  // or using inline grid maybe? 
`
```

PR #507, opened 2021-05-04
🥯 Change breadcrumbs logic

### LINKED ISSUE #490: Implement breadcrumbs logic
### Requirements

*   update the breadcrumbs logic
*   the last breadcrumb shouldn't be clickable

p-sad:
Closes #490

### DIFF
--- a/packages/ui/src/app/components/AppPage.tsx
+++ b/packages/ui/src/app/components/AppPage.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components'
 
 import { PageContent } from '../../common/components/page/PageContent'
 import { Breadcrumbs } from '../../common/components/page/Sidebar/Breadcrumbs/Breadcrumbs'
+import { breadcrumbsOptions } from '../constants/breadcrumbs'
 
 interface AppPageProps {
   lastBreadcrumb?: string
@@ -11,7 +12,7 @@ interface AppPageProps {
 
 export const AppPage = ({ children, lastBreadcrumb }: AppPageProps) => (
   <PageContent>
-    <Breadcrumbs lastBreadcrumb={lastBreadcrumb} />
+    <Breadcrumbs lastBreadcrumb={lastBreadcrumb} breadcrumbsOptions={breadcrumbsOptions} />
     <PageContainer>{children}</PageContainer>
   </PageContent>
 )
--- a/packages/ui/src/app/components/AppPage.tsx
+++ b/packages/ui/src/app/components/AppPage.tsx
@@ -5,13 +5,13 @@ import { PageContent } from '../../common/components/page/PageContent'
 import { Breadcrumbs } from '../../common/components/page/Sidebar/Breadcrumbs/Breadcrumbs'
 
 interface AppPageProps {
-  crumbs: { href: string; text: string }[]
+  objectName?: string
   children: ReactNode
 }
 
-export const AppPage = ({ children, crumbs }: AppPageProps) => (
+export const AppPage = ({ children, objectName }: AppPageProps) => (
   <PageContent>
-    <Breadcrumbs crumbs={crumbs} />
+    <Breadcrumbs objectName={objectName} />
     <PageContainer>{children}</PageContainer>
   </PageContent>
 )
--- a/packages/ui/src/app/components/AppPage.tsx
+++ b/packages/ui/src/app/components/AppPage.tsx
@@ -5,13 +5,13 @@ import { PageContent } from '../../common/components/page/PageContent'
 import { Breadcrumbs } from '../../common/components/page/Sidebar/Breadcrumbs/Breadcrumbs'
 
 interface AppPageProps {
-  objectName?: string
+  lastBreadcrumb?: string
   children: ReactNode
 }
 
-export const AppPage = ({ children, objectName }: AppPageProps) => (
+export const AppPage = ({ children, lastBreadcrumb }: AppPageProps) => (
   <PageContent>
-    <Breadcrumbs objectName={objectName} />
+    <Breadcrumbs lastBreadcrumb={lastBreadcrumb} />
     <PageContainer>{children}</PageContainer>
   </PageContent>
 )
--- a/packages/ui/src/app/pages/Profile/MyMemberships.tsx
+++ b/packages/ui/src/app/pages/Profile/MyMemberships.tsx
@@ -7,19 +7,12 @@ import { AppPage } from '../../components/AppPage'
 import { Memberships } from './components/Memberships'
 import { MyProfileTabs } from './components/MyProfileTabs'
 
-export const MyMemberships = () => {
-  const crumbs = [
-    { href: '#', text: 'My Profile' },
-    { href: '#', text: 'My Memberships' },
-  ]
-
-  return (
-    <AppPage crumbs={crumbs}>
-      <PageHeader>
-        <PageTitle>My Profile</PageTitle>
-        <MyProfileTabs />
-      </PageHeader>
-      <Memberships />
-    </AppPage>
-  )
-}
+export const MyMemberships = () => (
+  <AppPage>
+    <PageHeader>
+      <PageTitle>My Profile</PageTitle>
+      <MyProfileTabs />
+    </PageHeader>
+    <Memberships />
+  </AppPage>
+)
--- a/packages/ui/src/common/api/schemas/schema.graphql
+++ b/packages/ui/src/common/api/schemas/schema.graphql
@@ -7286,5 +7286,6 @@ input WorkingGroupWhereInput {
 }
 
 input WorkingGroupWhereUniqueInput {
-  name_eq: String
+  id: ID
+  name: String
 }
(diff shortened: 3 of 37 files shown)


packages/ui/src/common/model/breadcrumbs.ts
```diff
@@ -0,0 +1,11 @@
...
+  { path: '/profile', breadcrumb: 'My Profile' },
+  { path: '/profile/memberships', breadcrumb: 'My Memberships' },
+  { path: '/working-groups', breadcrumb: 'Working Groups' },
+  { path: '/working-groups/openings', breadcrumb: 'Openings' },
+  { path: '/working-groups/my-applications', breadcrumb: 'My Applications' },
+  { path: '/working-groups/my-roles', breadcrumb: 'My Roles' },
+  { path: '/members', breadcrumb: 'Members' },
+]
+
+export const excludePaths = ['/', '/working-groups/grouppreview']
```
p-sad, line 11: Not sure if this belongs in `model`, couldn't find a better place for it though

packages/ui/src/app/components/AppPage.tsx
```diff
@@ -5,13 +5,13 @@ import { PageContent } from '../../common/components/page/PageContent'
 import { Breadcrumbs } from '../../common/components/page/Sidebar/Breadcrumbs/Breadcrumbs'
 
 interface AppPageProps {
-  crumbs: { href: string; text: string }[]
+  objectName?: string
```
jodator, line 8: I'm not sure the meaning of `objectName` ;)

packages/ui/src/app/pages/Profile/MyMemberships.tsx
```diff
@@ -7,19 +7,12 @@ import { AppPage } from '../../components/AppPage'
...
-}
+export const MyMemberships = () => (
+  <AppPage>
+    <PageHeader>
+      <PageTitle>My Profile</PageTitle>
+      <MyProfileTabs />
+    </PageHeader>
+    <Memberships />
+  </AppPage>
+)
```
jodator, line 18: Nice :D

packages/ui/src/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsList.tsx
```diff
@@ -1,17 +1,27 @@
...
-  crumbs: Array<BreadcrumbsItemProps>
+  objectName?: string
 }
 
-export function BreadcrumbsList({ crumbs }: BreadcrumbsListProps) {
+export function BreadcrumbsList({ objectName }: BreadcrumbsListProps) {
+  const { location } = useHistory()
+  const routesList = objectName ? [...knownRoutes, { path: location.pathname, breadcrumb: objectName }] : knownRoutes
+  const crumbs = useBreadcrumbs(routesList, { excludePaths })
+  console.log(crumbs)
```
jodator, line 18:
```suggestion
  const crumbs = useBreadcrumbs(routesList, { excludePaths })
```

:oncoming\_police\_car:

packages/ui/src/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsList.tsx
```diff
@@ -1,17 +1,27 @@
...
 
 export interface BreadcrumbsListProps {
-  crumbs: Array<BreadcrumbsItemProps>
+  objectName?: string
 }
 
-export function BreadcrumbsList({ crumbs }: BreadcrumbsListProps) {
+export function BreadcrumbsList({ objectName }: BreadcrumbsListProps) {
+  const { location } = useHistory()
+  const routesList = objectName ? [...knownRoutes, { path: location.pathname, breadcrumb: objectName }] : knownRoutes
```
jodator, line 16: Now I see ;) So the `objectName` is `breadcrumb`. So the better name would be either `activeBreadcrumb`, `lastBreadcrumb` or just `breadcrumb` (ps. I still don't know what this does so I might shooting blind with the names).

packages/ui/src/common/model/breadcrumbs.ts
jodator, line 11: Could you also change the path for the working group?

Having in mind that the list of groups is finite and there shouldn't be a conflict with other routes we could have this route:

```
/working-groups/:groupName

so:

/working-groups/membership
/working-groups/storage
etc
```

packages/ui/src/common/components/page/Sidebar/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -6,12 +6,14 @@ import { HomeLink } from './HomeLink'
 
 export type BreadcrumbsNavProps = BreadcrumbsListProps
 
-export const Breadcrumbs = React.memo(({ crumbs }: BreadcrumbsNavProps) => (
-  <BreadcrumbsNavigation>
-    <HomeLink />
-    <BreadcrumbsList crumbs={crumbs} />
-  </BreadcrumbsNavigation>
-))
+export const Breadcrumbs = React.memo(({ objectName }: BreadcrumbsNavProps) => {
```
jodator, line 9: That type is unnecessary. Could you remove it?

jodator: The `BreadcrumbsNavProps` isn't needed.

packages/ui/src/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsList.tsx
```diff
@@ -1,17 +1,27 @@
...
+  const crumbs = useBreadcrumbs(routesList, { excludePaths })
+  console.log(crumbs)
   return (
     <BreadcrumbsListComponent>
-      {crumbs.map(({ href, text }, index) => (
-        <BreadcrumbsItem key={index} href={href} text={text} />
+      {crumbs.map(({ match: { url }, breadcrumb, key }) => (
+        <BreadcrumbsItem key={key} url={url}>
+          {breadcrumb}
+        </BreadcrumbsItem>
```
jodator, line 24: Could you also make the last breadcrumb text only? Right now it is disabled only but still renders as a link on the page.

packages/ui/src/common/model/breadcrumbs.ts
jodator: I don't have a better idea for the second level, but the first level (`common`) is IMO wrong.

The know paths are tied to the application, so they should be moved to the `src/app`.

This will make you change how the `<Breadcrumbs />` component is created in terms of inversion of control. The known paths should be injected as a dependency (prop) to the component.

When working on that please check if some memo isn't needed somewhere to prevent re-renders of the breadcrumbs when the page is changed (re-rendered) but the URL stays the same.

jodator (changes requested):
Nice :) Reading through the code I had some comments how we could improve this.

Looking that I asked you to change something in the component, I think that this is a good case to use the TDD approach for the changes :) 

Please start with a commit that adds tests for the current `<Breadcrumbs />` component. I don't see a need for other tests TBH.

A story for the breadcrumbs is also welcomed :)

packages/ui/src/common/api/schemas/schema.graphql
```diff
@@ -7286,6 +7286,5 @@ input WorkingGroupWhereInput {
 }
 
 input WorkingGroupWhereUniqueInput {
-  id: ID
-  name: String
```
jodator, line 7290: At least for now, please don't change the basic schema as this is what Hydra outputs.

packages/ui/src/working-groups/queries/workingGroups.graphql
```diff
@@ -101,8 +101,8 @@ query GetWorkingGroupOpeningQuestions($id: ID!) {
   }
 }
 
-query GetWorkingGroup($id: ID!) {
-  workingGroupByUniqueInput(where: { id: $id }) {
+query GetWorkingGroup($name_eq: String) {
+  workingGroupByUniqueInput(where: { name_eq: $name_eq }) {
     ... WorkingGroupFields
```
jodator, line 106: I think that this one will be to change so the other `name_eq` in the code may left intact

jodator (changes requested):
Quick R- as I think that we shouldn't remove things from the Schema as it is generated by Hydra tools.

packages/ui/test/page/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+describe('Breadcrumbs', () => {
+  it('Short path', () => {
+    location.pathname = '/profile'
+    renderCrumbs()
+    expect(screen.getByText('My Profile')).toBeDefined()
+  })
+
+  function renderCrumbs() {
+    render(
+      <HashRouter>
```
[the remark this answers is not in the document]
jodator: That's and old comment.

jodator (approved):
Thanks!

PR #511, opened 2021-05-05
🎉 Apply fore a role success

jodator:
Sorry for the changes in other components:

1. The `PercentageChart` caused React errors in the console
2. I've unified getting some mocks because it is easier to change one method rather than 5 stories :D

In this PR:

- I've added better types for substrate event data reading
- Implemented the "Success" step data retrieval that _should_ be consistent with Joystream event naming
- The stepper / form data typings needs some work in a follow-up PR


jodator:
Huh, I don't know why the tests fails :cry: I'll check it tomorrow morning.

jodator:
:crossed\_fingers: fixed hopefully

PR #521, opened 2021-05-06
🔖 New tooltip preslice

### LINKED ISSUE #494: Remake tooltip component

ClumsyVlad:
Closes #494


ClumsyVlad:
Check states and hover/focus troubles. I could not solve this normally :(

jodator:
@ClumsyVlad - take a look at the changes:

*   I've reworked how the tooltip show/hide delay should work
*   I made click on tooltip permanent (you have to unclick to hide it)
*   I've added a tooltip story
*   I've split the code because the file was too big

I think that the job here is almost ready at this stage. There's still room to improvement codewise:

1.  This is a general component, but there are some components named for external domains. I think that it should be possible to achieve the same from the outside.
2.  The `React.portal()` might be moved to a separate component as `TooltipPopup` or something like this.

Ad 1.:

I think that something similar to this:

```typescript
export const AvatarStarTooltipContainer = styled(DefaultTooltip)<DefaultTooltipProps & DarkTooltipInnerItemProps>`
  color: ${Colors.White};
  border-color: ${Colors.Blue[500]};
  background-color: ${Colors.Blue[500]};

  ${TooltipComponent}:hover > &,
  ${TooltipComponent}:focus > & {
    color: ${Colors.White};
    border-color: ${Colors.Blue[400]};
    background-color: ${Colors.Blue[400]};
  }
`
```

jodator:
@ClumsyVlad So the comments I had earlier:

1.  Separation of concerns - the base Tooltip container should not know nor care about Members tooltips. The member tooltip should override base tootlipt styles: [`426aaf4`](https://github.com/Joystream/pioneer/pull/521/commits/426aaf4d7815366192605dcce534d091670acd5d)
2.  Make Tooltip a bit easier to read (in JSX part at least) by creating a component with: [`84a96f8`](https://github.com/Joystream/pioneer/pull/521/commits/84a96f8077724f792ee1f9927278e1253c1a8aae)

PR #522, opened 2021-05-06
Filter the Member list

### LINKED ISSUE #435: Apply filters to the search results

thesan:
Closes #435

### DIFF
--- a/packages/ui/src/common/components/forms/InputComponent.tsx
+++ b/packages/ui/src/common/components/forms/InputComponent.tsx
@@ -137,24 +137,10 @@ export const InputComponent = React.memo(
   }
 )
 
-export const InputText = React.memo(
-  ({ id, value, required, validation, placeholder, disabled, onChange }: InputProps) => {
-    return (
-      <Input
-        id={id}
-        name={id}
-        type="text"
-        value={value}
-        required={required}
-        validation={validation}
-        placeholder={placeholder}
-        disabled={disabled}
-        onChange={onChange}
-        autoComplete="off"
-      />
-    )
-  }
-)
+type IputTextProps = React.InputHTMLAttributes<HTMLInputElement> & InputProps
+export const InputText = React.memo((props: IputTextProps) => {
+  return <Input name={props.id} type="text" autoComplete="off" {...props} />
+})
 
 export const InputNumber = React.memo(
   ({ id, value, required, validation, placeholder, disabled, onChange }: InputProps) => {
--- a/packages/ui/src/memberships/components/MemberList/Members.tsx
+++ b/packages/ui/src/memberships/components/MemberList/Members.tsx
@@ -1,3 +1,4 @@
+import { Loading } from '@/common/components/Loading'
 import React, { Dispatch, ReactNode } from 'react'
 
 import { List, ListItem } from '../../../common/components/List'
@@ -11,11 +12,12 @@ import { colLayoutByType } from '../MemberListItem/Fileds'
 type SortKey = MemberListOrder['sortBy']
 interface MemberListProps {
   members: Member[]
+  isLoading?: boolean
   order?: MemberListOrder
   onSort?: Dispatch<SortKey>
 }
 
-export const MemberList = ({ members, order, onSort }: MemberListProps) => {
+export const MemberList = ({ isLoading, members, order, onSort }: MemberListProps) => {
   const SortHeader =
     order && onSort && members.length > 1
       ? ({ children, sortKey }: { children: ReactNode; sortKey: SortKey }) => (
@@ -28,7 +30,9 @@ export const MemberList = ({ members, order, onSort }: MemberListProps) => {
         )
       : ListHeader
 
-  return (
+  return isLoading ? (
+    <Loading />
+  ) : (
     <div>
       <ListHeaders colLayout={colLayoutByType('Member')}>
         <SortHeader sortKey="id">Member ID</SortHeader>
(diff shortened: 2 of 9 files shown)


packages/ui/src/common/components/forms/InputComponent.tsx
```diff
@@ -137,24 +136,9 @@ export const InputComponent = React.memo(
...
-        disabled={disabled}
-        onChange={onChange}
-        autoComplete="off"
-      />
-    )
-  }
-)
+export const InputText = React.memo((props: InputProps) => {
+  return <Input name={props.id} type="text" autoComplete="off" {...props} />
+})
```
thesan, line 141: I changed this in order to add an onKeyDown to the input but it might be overkill (especially because of the type change)

jodator: Sure :) That makes sense.

packages/ui/test/membership/hooks/useMembers.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  beforeEach(() => {
+    mockedUseFilterMembersQuery.mockClear()
+  })
+
+  it('calls the useFilterMembersQuery', () => {
+    renderUseMembers({})
+    expect(useFilterMembersQuery).toBeCalled()
+  })
+
+  it('queries members by ascending ids (default)', () => {
```
jodator, line 40: Hey, we follow a different naming for tests. You can try to change the names to follow that, but that's not a requirement.

In this case, we would name the test: "Default order". (Starting with a capital letter).

The other test names could be:

*   Inverse order
*   Order by handle
*   Search by handle
*   Search by Id

This is to:

*   have the description of the test be more compact
*   describe the outcome/state rather than what the test does

It differs from the more broadly used "sentence test description". However, I find it better in terms of readability when a test fails.

packages/ui/src/memberships/components/MemberList/Members.tsx
```diff
@@ -28,7 +31,9 @@ export const MemberList = ({ members, order, onSort }: MemberListProps) => {
         )
       : ListHeader
 
-  return (
+  return isLoading ? (
+    <Loading />
+  ) : (
```
jodator, line 36: Could you unpack `?`. I think it will provide more readability:

```typescript
if (isLoading) {
	return <Loading />
}

return (
  <div>...
```

packages/ui/src/memberships/queries/members.graphql
```diff
@@ -23,6 +23,34 @@ query GetMembers($rootAccount_in: [String!], $controllerAccount_in: [String!], $
   }
 }
 
+query FilterMembers(
+  $id: ID
+  $search: String
+  $isVerified: Boolean
+  $isFoundingMember: Boolean
+  # $isCouncil: Boolean!
+  # role: ???
```
jodator, line 32: You can safely omit comments like this. If there's a TODO a better approach would be to create an issue on GH with things that were left to do. However, if this is a part of a ticket that you're working on, you don't have to create a separate one.

In this case, a follow-up is welcomed as we have to remember that in the feature when we get the `isCouncil` nad roles from the query node.

packages/ui/src/mocks/resolvers.ts
```diff
@@ -39,42 +44,63 @@ export const getMemberResolver = (obj: any, args: any, context: any, info: any)
...
   const authorizedKeys = [EntryAsc, EntryDesc, HandleAsc, HandleDesc]
   if (!orderBy || !authorizedKeys.includes(orderBy)) {
-    return
+    return members
   }
 
   const [key, direction] = orderBy.toLowerCase().split('_')
   const membersKey = (key === 'entry' ? 'id' : key) as 'id' | 'handle'
   const fact = direction === 'desc' ? 1 : -1
+  const sortFn = (a: T, b: T) => fact * (b[membersKey] ?? '').localeCompare(a[membersKey] ?? '')
```
jodator, line 62: Try to avoid abbreviations. I'd go with something like:

```suggestion
  const sortByKey = (a: T, b: T) => fact * (b[membersKey] ?? '').localeCompare(a[membersKey] ?? '')
```

jodator: Or even:  
 

```typescript
const sortByKey = (key: key of Member) => {
  return (a: T, b: T) => fact * (b[key] ?? '').localeCompare(a[key] ?? ''
}
```

then use it:

```typescript
members.sort(sortByKey(membersKey))
```

However, this is a code used for mocks. So, we don't need to spend much time on it ;) Renaming it to `sortByKey` is sufficient.

packages/ui/src/memberships/components/MemberListFilters/MemberListFilters.tsx
```diff
@@ -71,6 +71,12 @@ export const MemberListFilters = ({ roles, onApply }: MemberListFiltersProps) =>
           <InputText
             placeholder="Search"
             value={search}
+            onKeyDown={
+              onApply &&
+              (({ key }) => {
+                key === 'Enter' && onApply(filters)
+              })
+            }
```
jodator, line 79: Could you extract this one to a function? It should be more readable:

```typescript
<InputText ... onKeyDown={onSearchKeyDown} />
```

jodator, line 79: Could you extract this one to a function? It should be more readable:

```typescript
<InputText ... onKeyDown={onSearchKeyDown} />
```

jodator (changes requested):
Hey @thesan. The component works nice!

I have a few comments to align you with the desired code style so don't worry if you see longer comment :)

packages/ui/test/membership/hooks/useMembers.test.tsx
thesan: Sounds good. I prefer to have the test suite fitting with the others

packages/ui/src/memberships/queries/members.graphql
thesan: I removed the comments and created #525

PR #523, opened 2021-05-06
🚨 Open position preview

wyrdrender:
See #292

* Update openings graphql schema
* Implement missing graphql data
* Add opening types
* Add useCopyToClipboard hook

TODO:
* Add applications section
* Add 'created' footer


packages/ui/src/common/api/queries/__generated__/baseTypes.generated.ts
```diff
@@ -38,24 +38,6 @@ export type ApplicationFormQuestion = BaseGraphQlObject & {
   applicationformquestionanswerquestion?: Maybe<Array<ApplicationFormQuestionAnswer>>
 }
 
-export type ApplicationFormQuestionAnswer = BaseGraphQlObject & {
-  __typename: 'ApplicationFormQuestionAnswer'
-  id: Scalars['ID']
-  createdAt: Scalars['DateTime']
```
jodator, line 44: I might be missing something in the GH UI. The generated file should be an outcome of the `schema.graphql` or a `../queries/*.graphql` files. However, If you didn't change the `schema.graphql` file it is OK as you've generated the other queries.

Just to double-check :)

packages/ui/src/working-groups/constants.ts
```diff
@@ -0,0 +1,14 @@
+export enum OpeningStatuses {
+  UPCOMING = 'OpeningStatusUpcoming',
+  OPEN = 'OpeningStatusOpen',
+  FILLED = 'OpeningStatusFilled',
+  CANCELLED = 'OpeningStatusCancelled',
+}
```
jodator, line 6: I was thinking about this one. I rather not use enums (minor issues in the runtime). But here I think that this actually a really nice use case for string enum.

jodator (commented):
Looks good :) I have only one question about the schema, but we can discuss this later.

It looks like the generated files were changed without schema changes. I'd blame that on the fact that a tool changed. But if not, please make a follow-up PR.

Issue #517, opened 2021-05-06
Update GraphQL mocks


jodator:
### Requirements

*   provide a generic method for providing data
*   provide pagination
*   provide generic search

PR #536, opened 2021-05-10
🤖 Automate base resolvers

jodator:
See #517


jodator:
@wyrdrender - I've updated the resolvers. I see that most of the resolvers can be generalised as we did almost everything the same way.

ATM I've added two types of automatic resolvers:

*   where -> returns list of models
*   connection -> returns paginated list of models

The remaining one is for a single model (unique query).

wyrdrender (approved):

PR #542, opened 2021-05-11
🖇 Add a mapping for transactions common between working groups

p-sad:


packages/ui/src/working-groups/hooks/useWithdrawMembershipTransaction.ts
```diff
@@ -0,0 +1,15 @@
...
+export function useWithdrawMembershipTransaction(groupName: string) {
+  const { api } = useApi()
+
+  return useMemo(() => {
+    if (api && groupName in groupExtrinsics) {
+      const name = groupName as keyof typeof groupExtrinsics
+      return api.tx[groupExtrinsics[name]].withdrawApplication
+    }
+  }, [api, groupName])
+}
```
jodator, line 15: I'd say that this is too narrow for a hook -> There's ~10 extrinsic to handle so that would mean 10 atomic hooks.

I think that we might be better of with a pure function that will return proper extrinsic:

```typescript
const tx = getGroup(api, 'forum').withdrawAppliction()
```

If you manage to have proper type on the `tx` that would be an awesome bouns (I didn't check if that is not a case already ;) )

packages/ui/src/working-groups/types/groupExtrinsics.ts
```diff
@@ -0,0 +1,19 @@
+import { AugmentedSubmittables } from '@polkadot/api/types'
+
+enum groupNames {
+  forum,
+  storage,
+  content,
+  membership,
+}
+type groupSubmittableSet = keyof AugmentedSubmittables<'rxjs'> &
+  ('forumWorkingGroup' | 'storageWorkingGroup' | 'contentDirectoryWorkingGroup' | 'membershipWorkingGroup')
```
jodator, line 10: I looks like my above comment might be solved by this, so that's :+1:

jodator (changes requested):
Exactly, something like this, except that I would use a pure function instead of a hook.

packages/ui/src/working-groups/types/groupExtrinsics.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const groupExtrinsics: {
+  [key in keyof typeof groupNames]: groupSubmittableSet
+} = {
```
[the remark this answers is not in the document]
jodator: A bit simpler notation. We don't need Enum as the union type is OK for string ;)

packages/ui/src/working-groups/model/getGroup.ts
```diff
@@ -0,0 +1,10 @@
+import { ApiRx } from '@polkadot/api'
+
+import { groupExtrinsics } from '../types'
+
+export function getGroup(api?: ApiRx, groupName?: string) {
```
jodator, line 5: The `groupName` should be proper type IMO:

```suggestion
import { groupExtrinsics, GroupName } from '../types'

export function getGroup(api: ApiRx, groupName: GroupName) {
```

PR #544, opened 2021-05-12
👨🏻‍⚖️ Add My Roles page with current & past roles

DevNikita16:
👨🏻‍⚖️ Add My Roles page with current & past roles


yarn-error.log
```diff
@@ -0,0 +1,17518 @@
+Arguments: 
+  /usr/local/bin/node /usr/local/bin/yarn init
+
+PATH: 
+  /opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Apple/usr/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Users/nikita/Documents/pioneer/node_modules/.bin
+
+Yarn version: 
+  1.22.10
```
jodator, line 8: Oh that shouldn't land in the PR ;)

Can you remove this file and add `*.log` to gitignore?

packages/ui/src/working-groups/components/RolesList.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <TextInlineBig>
+            <TokenValue value={new BN(worker.stake)} />
+          </TextInlineBig>
+          <OACSubscriptionWide>Staked</OACSubscriptionWide>
+        </OpenItemSummaryColumn>
+      </OACItemSummary>
+      <ButtonGhost square size="medium">
+        <KebabMenuIcon />
+      </ButtonGhost>
+    </OACWrap>
```
jodator, line 77: I wonder what to do with this one. The components' names suggest that those are for "application and openings" components. However, it looks like we can re-use them elsewhere.

I think that we need some better abstraction for this.

ps.: I'll keep in mind that, for now it is OK.

jodator (changes requested):
Nice PR @DevNikita16 👍 

I have only one things to ask you: please remove the log :)

packages/ui/src/working-groups/components/RolesList.tsx
DevNikita16: Probably this should be some component in common, because it appears in a lot of places

yarn-error.log
DevNikita16: Done

PR #548, opened 2021-05-12
🏭 Add mocks generator

jodator:
I've added the `mocks:generate` script that creates stubs for `mocks/data/raw` JSON files.

Additionally, I've updated some data:

*   use `createdAtBlock.timetamp` instead of `createdAt` since we already fetch block data.
*   The way how workers are created is changed (no leader for now)
*   applications/answers/questions are easily tied together by the generate script
*   only two know members (Alice and Bob)


jodator:
@wyrdrender & @p-sad - I've added you both as reviewers mostly to show you the changes in the mocks generation.

Hopefully, we will spend less time on those and we could introduce changes in fewer places.

p-sad (approved):

PR #549, opened 2021-05-12
🦶 Open position preview footer

wyrdrender:
* Add footer component with created date and block info

### DIFF
--- a/packages/ui/src/common/components/AboutDateColumn.tsx
+++ b/packages/ui/src/common/components/AboutDateColumn.tsx
@@ -13,13 +13,12 @@ import { TextMedium, TextSmall } from './typography'
 
 interface Props {
   block: Block
-  time?: string
   horizontal?: boolean
 }
 
-export const AboutDateColumn = React.memo(({ time, block, horizontal }: Props) => (
+export const AboutDateColumn = React.memo(({ block, horizontal }: Props) => (
   <Wrapper horizontal={horizontal}>
-    <AboutText>{formatDateString(time ?? block.timestamp)}</AboutText>
+    <AboutText>{formatDateString(block.timestamp)}</AboutText>
     {horizontal && <Separator>|</Separator>}
     <BlockInfo>
       <BlockIcon />
--- a/packages/ui/src/common/components/AboutDateColumn.tsx
+++ b/packages/ui/src/common/components/AboutDateColumn.tsx
@@ -1,5 +1,7 @@
 import React from 'react'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
+
+import { spacing } from '@/common/utils/styles'
 
 import { Colors, Transitions } from '../constants'
 import { formatDateString, formatTokenValue } from '../model/formatters'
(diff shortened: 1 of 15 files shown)


packages/ui/src/common/queries/blocks.graphql
```diff
@@ -4,4 +4,5 @@ fragment BlockFields on Block {
   number
   timestamp
   network
+  createdAt
```
jodator, line 7: ~I'm sorry, I forgot about that. For block, it looks like we have to use `timestamp`. That's because the block's timestamp is related to the actual time a block was mined, while `createdAt` is the timestamp of the DB record.~ Fortunately, I'm wrong here :sweat\_smile: 

It looks like all the [timestamps are related to the block time](buy_membership_without_referrer), which makes more sense.

This however, makes a little mess in handling :) I'd prefer to use a Block's timestamp here as we already are going to use `Block` instance to display that block-date component.

packages/ui/src/working-groups/types/WorkingGroupOpening.ts
```diff
@@ -28,6 +35,12 @@ export interface WorkingGroupOpening {
     member: Member
     status: string
   }[]
+  createdAtBlock: {
+    id: string
+    number: number
+    network: Network
+    createdAt: string
+  }
```
jodator, line 43: Could you use `Block` from `@/common/types`? It looks like it also needs an update to have a `timestamp` field.

packages/ui/src/working-groups/types/WorkingGroupOpening.ts
```diff
@@ -52,6 +65,12 @@ export const asWorkingGroupOpening = (fields: WorkingGroupOpeningFieldsFragment)
         status: application.status.__typename,
       }))
     : [],
+  createdAtBlock: {
+    id: fields.createdAtBlock.id,
+    number: fields.createdAtBlock.number,
+    network: fields.createdAtBlock.network,
+    createdAt: fields.createdAtBlock.createdAt,
+  },
```
jodator, line 73: And for this we have a `asBlock` already.

packages/ui/src/common/components/AboutDateColumn.tsx
```diff
@@ -12,24 +14,41 @@ import { TextMedium, TextSmall } from './typography'
...
   time: string
+  horizontal?: boolean
 }
 
-export const AboutDateColumn = React.memo(({ time, block }: Props) => (
-  <Column>
+export const AboutDateColumn = React.memo(({ time, block, horizontal }: Props) => (
+  <Wrapper horizontal={horizontal}>
     <AboutText>{formatDateString(time)}</AboutText>
+    {horizontal && <Separator>|</Separator>}
```
jodator, line 23: After the changes in where we get the timestamp, I think that we should also use only `Block` here and use its timestamp.

jodator (commented):
@wyrdrender - I've left my initial comments in the PR. I see that all of the "date"-like fields duplicates block timestamp. This is good - it makes sense to tie all the dates to the block's timestamp. Because every change on the blockchain is related to a given block all other timestamps will reflect that something happened on that timestamp.

I'll make the quick changes as those are minimal and merge your PR :)

PR #558, opened 2021-05-13
Show component names in devtools

### LINKED ISSUE #556: Show component name in devtools

ClumsyVlad:
Closes #556


ClumsyVlad:
In this case looks like we will have same names on production. Do we need to change this?

jodator:
> In this case looks like we will have same names on production. Do we need to change this?

Yeah, you can parametrize that and only add this plugin in development mode:

```
argv.mode === 'development' // line 23 as well
```

Issue #296, opened 2021-03-25
Withdraw application


jodator:
### Requirements

- Implement working group withdraw opening application transaction

jodator:
See: https://github.com/Joystream/pioneer/pull/563#issuecomment-840615934

PR #563, opened 2021-05-13
↩️ Add WithdrawApplication modal

p-sad:
See #296


jodator:
The PR is Good, but I've got missed up with the requirements and it seems that the idea for withdrawing applications was withdrawn.

PR #572, opened 2021-05-14
Introduce opening layout

jodator:


jodator:
@ClumsyVlad you can take over this branch. I needed to make some quick fixes.

ClumsyVlad (approved):

jodator:
Since @DevNikita16 might need that one, I'm merging it blindly :see\_no\_evil:

packages/ui/src/memberships/components/MemberListFilters/MemberListFilters.tsx
```diff
@@ -117,46 +118,31 @@ export const MemberListFilters = ({ roles, onApply }: MemberListFiltersProps) =>
...
-  text-align: end;
+  display: flex;
+  position: relative;
 `
 
 const Fields = styled.div`
-  background: ${Colors.Black[100]};
-  display: flex;
+  display: grid;
+  grid-auto-flow: column;
```
thesan, line 142: Screenshot from 2021-05-17 20-35-41

@jodator with this display grid auto flow column I'm not sure how to align the apply button to the end (without using a position absolute). Which makes the layout harder to reuse for the past proposals filter box. Does it have to be a display grid ?

jodator: OH, I see that now :disappointed:

It doesn't have to be `grid` IMO. The problem previously was with smaller screen sizes as on my mac's screen I couldn't reach the Apply button.

However, our designer is already working on changing the filters section here to a similar one introduced in the Proposals filtering.

jodator: OH, I see that now :disappointed:

It doesn't have to be `grid` IMO. The problem previously was with smaller screen sizes as on my mac's screen I couldn't reach the Apply button.

However, our designer is already working on changing the filters section here to a similar one introduced in the Proposals filtering.

thesan: Right I was just looking at the wired frames so I had not seen the different design for the past proposal filter :facepalm: .

PR #580, opened 2021-05-14
🐁 Prepare side panel scroll

### LINKED ISSUE #528: Add scroll in activities-like sections in side panels
Troubles:
1) How high should it be?
2) Shadows clipping cause overflow

ClumsyVlad:
Closes #528


ClumsyVlad:
@jodator It's better if we update the neighborHeight value when the neighbor resizes. Could you please add such a listener inside SidePanel.tsx?

jodator (approved):
Let's see how this will work ;)

Issue #541, opened 2021-05-11
Current proposals


jodator:
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=533%3A10)

### Requirements

- display current / past tabs stubs
- prepare proposals list component (add storybook story for every new component)
- prepare Proposal interface
- add stub "proposal activities" component

### Implementaiton detials

We should start implementing the base components using storybook-first aproach and preparing the internal Proposal interface. All of the data should be fetched using hooks. Before we get the proper GraphQL schema we will stub data in hooks. The Hydra (GraphQL server) integration will be added later but the changes should be limited to the hooks internals. The hook API should be similar to the one created for the Working Groups or Memberships queries.

jodator:
@thesan This issue can be split into multiple PRs.

jodator:
You can also create a base hook to get the proposals, something like this:

```typescript
const useProposals = () => {
  const proposals: Proposal[] = [
    { id: 0, stage: 'DECIDING', type: 'Founding request', title: 'Lorem ipsum foo bar', proposer: { id: '0', ... } },
    { id: 1, stage: 'DORMANT', type: 'Founding request', title: 'Lorem ipsum foo bar', proposer: { id: '0', ... } },
  ]
  
  return {
    isLoading: false
    proposals
  }
}
```

This will make the upgrade to the GraphQL API a bit easier.

jodator:
Ps.: see the updated designs for the proposals list: ([design](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2599%3A109614)).

I think that the current proposals will have the same styles as the past. The only change is that the background of the current opening should be white. Similarly to [current](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=16%3A2266) / [past](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=257%3A16856) openings.

thesan:
> You can also create a base hook to get the proposals, something like this:
> 
> ```ts
> const useProposals = () => {
>   const proposals: Proposal[] = [
>     { id: 0, stage: 'DECIDING', type: 'Founding request', title: 'Lorem ipsum foo bar', proposer: { id: '0', ... } },
>     { id: 1, stage: 'DORMANT', type: 'Founding request', title: 'Lorem ipsum foo bar', proposer: { id: '0', ... } },
[...]

PR #581, opened 2021-05-14
💍 Add Proposals list

thesan:


packages/ui/src/proposals/constants/proposals.ts
```diff
@@ -0,0 +1,9 @@
+export const ProposalStages = {
+  deciding: { title: 'Deciding', desc: 'Lorem ipsum, dolor sit amet consectetur' },
+  dormant: { title: 'Dormant', desc: 'Lorem ipsum, dolor sit amet consectetur' },
+  gracing: { title: 'Gracing', desc: 'Lorem ipsum, dolor sit amet consectetur' },
+  succeded: { title: 'Execution Succeded', desc: 'Lorem ipsum, dolor sit amet consectetur' },
+}
+export const ProposalTypes = {
+  funding: { title: 'Funding Request', desc: 'Lorem ipsum, dolor sit amet consectetur' },
+}
```
jodator, line 9: We can keep it for now but in the `Proposal` format:

```
export const proposals: Proposal[] = [
   { id: 0, stage: 'DECIDING', type: 'Founding request', title: 'Lorem ipsum foo bar', proposer: { id: '0', ... } },
   { id: 1, stage: 'DORMANT', type: 'Founding request', title: 'Lorem ipsum foo bar', proposer: { id: '0', ... } },
]
```

For the helper text, that can be also the same for all and inlined.

jodator (changes requested):
Hi @thesan, thanks for the PR!

I would only like to ask you to change the `Proposal` type a bit and introduce the `ProposalStege` type.

Other than that it looks OK.

packages/ui/src/proposals/constants/proposals.ts
jodator: This can be also reused in a hook later on: [https://github.com/Joystream/pioneer/issues/541#issuecomment-841236089](https://github.com/Joystream/pioneer/issues/541#issuecomment-841236089).

Issue #299, opened 2021-03-25
Leave a position


jodator:
@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4838716)

@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1608%3A28479)

### Requirements

- Implement using modal system

jodator:
@p-sad could you:

*   add this modal to the "leave position" button?
*   see if we can get the "unstaking period" from the mocks?

jodator:
Closing this as the only thing is the unstaking period: #299

PR #614, opened 2021-05-17
🚪 Add a Leave Role modal

p-sad:
See #299

### DIFF
--- a/packages/ui/src/working-groups/modals/LeaveRoleModal/LeaveRolePrepareModal.tsx
+++ b/packages/ui/src/working-groups/modals/LeaveRoleModal/LeaveRolePrepareModal.tsx
@@ -1,16 +1,17 @@
 import React, { useState } from 'react'
 
-import { ButtonPrimary } from '../../../common/components/buttons'
-import { InputComponent, InputText } from '../../../common/components/forms'
-import { Modal, ModalFooter, ModalHeader, Row } from '../../../common/components/Modal'
-import { Label, TextMedium } from '../../../common/components/typography'
+import { ButtonPrimary } from '@/common/components/buttons'
+import { InputComponent, InputText } from '@/common/components/forms'
+import { Modal, ModalFooter, ModalHeader, Row } from '@/common/components/Modal'
+import { Label, TextMedium } from '@/common/components/typography'
 
 interface Props {
+  unstakingPeriod: number
   onClose: () => void
   onContinue: (rationale: string) => void
 }
 
-export const LeaveRolePrepareModal = ({ onClose, onContinue }: Props) => {
+export const LeaveRolePrepareModal = ({ onClose, onContinue, unstakingPeriod }: Props) => {
   const [rationale, setRationale] = useState('')
   return (
     <Modal onClose={onClose} modalSize="m">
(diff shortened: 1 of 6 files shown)


packages/ui/src/working-groups/modals/LeaveRoleModal/LeaveRoleSignModal.tsx
```diff
@@ -0,0 +1,53 @@
...
+  })
+  return (
+    <TransactionModal status={status} onClose={onClose}>
+      <ModalHeader onClick={onClose} title="Leaving a position?" />
+      <ModalBody>
+        <Row>
+          <TextMedium>The transaction can only be signed with the membership's controller account.</TextMedium>
+        </Row>
+        <InputComponent label="From" inputSize="l" disabled={true}>
+          {signer ? <SelectedAccount account={signer} /> : <TextMedium>Controller account not found.</TextMedium>}
```
jodator, line 40: I will be always the `controllerAccount`:

> Signer uses controller account of member corresponding to member identifier in worker.

p-sad: Not sure if I understand, the signer cannot be changed (`SelectedAccount` instead of `SelectAccount`) and it is picked a couple of lines above:
```ts
const signer = allAccounts.find((account) => account.address == worker.membership.controllerAccount)
```

jodator: For most of the extrinsics, you'll need to use the `controllerAccount` of the associated member.

We only display the signer and do not allow to change it as it would cause the transaction to fail.

packages/ui/test/working-groups/modals/LeaveRoleModal.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+    it("With member's controller account in keyring", async () => {
+      await renderWithValidSigner()
+      await fireEvent.click(screen.getByRole('button', { name: 'Leave the position anyway' }))
+      expect(
+        screen.findByText("The transaction can only be signed with the membership's controller account.")
+      ).toBeDefined()
+      expect(screen.getByRole('button', { name: 'Sign and leave role' })).not.toBeDisabled()
+    })
+
+    it("Without member's controller account in keyring", async () => {
```
jodator, line 93: That shouldn't be possible. You should only be allowed to leave a role for the worker that you control.

This will only be possible if:

*   you have the account key pair (from the extension)
*   the app loads a member for your keys
*   the app detects if that Worker is associated with a Member that is yours

jodator (changes requested):
Overall the PR is good. There's some domain use misconceptions, but that we can iron out quickly :)

packages/ui/src/working-groups/modals/LeaveRoleModal/LeaveRolePrepareModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface Props {
+  onClose: () => void
+  onContinue: (rationale: string) => void
+}
+
+export const LeaveRolePrepareModal = ({ onClose, onContinue }: Props) => {
+  const [rationale, setRationale] = useState('')
+  return (
+    <Modal onClose={onClose} modalSize={'m'}>
```
wyrdrender, line 16:
```suggestion
    <Modal onClose={onClose} modalSize="m">
```

jodator (approved):

PR #616, opened 2021-05-17
🖼️ Complete the proposals current page layout

thesan:
See #541.

### DIFF
--- a/packages/ui/src/common/components/List/ListHeader.tsx
+++ b/packages/ui/src/common/components/List/ListHeader.tsx
@@ -6,7 +6,6 @@ import { TableListRowStyle } from './List'
 
 export const ListHeaders = styled.div`
   display: grid;
-  grid-area: accountstablenav;
   grid-template-rows: 1fr;
   width: 100%;
   ${(props: { colLayout?: string }) => TableListRowStyle(props.colLayout)}
--- a/packages/ui/src/proposals/hooks/useProposals.ts
+++ b/packages/ui/src/proposals/hooks/useProposals.ts
@@ -8,7 +8,11 @@ const mock: { [k: string]: { proposals: Proposal[] } } = {
   past: { proposals: [] },
 }
 
-export const useProposals = ({ type }: { type: 'current' | 'past' }) => {
+interface Props {
+  type: 'current' | 'past'
+}
+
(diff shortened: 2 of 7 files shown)


packages/ui/src/common/components/List/ListHeader.tsx
```diff
@@ -6,7 +6,6 @@ import { TableListRowStyle } from './List'
 
 export const ListHeaders = styled.div`
   display: grid;
-  grid-area: accountstablenav;
```
jodator, line 9: :+1:

packages/ui/src/proposals/hooks/useProposals.ts
```diff
@@ -0,0 +1,22 @@
...
+import { useMockDelay } from '@/mocks/hooks/useMockDelay'
+
+import { Proposal } from '../types'
+
+const mock: { [k: string]: { proposals: Proposal[] } } = {
+  current: { proposals: mockProposals },
+  past: { proposals: [] },
+}
+
+export const useProposals = ({ type }: { type: 'current' | 'past' }) => {
```
jodator, line 11: A bit cleaner approach would be to move that to an interface.

jodator, line 11:
```suggestion
const mock: { [k: string]: { proposals: Proposal[] } } = {
  current: { proposals: mockProposals },
  past: { proposals: [] },
}

interface Props {
  type: 'current' | 'past'
}

export const useProposals = ({ type }: Props) => {
```

Like this: :point\_up: Then you can extend the interface if needed and the function is easier to read.

jodator (changes requested):
Thanks! One question though, I'd move the mentioned types definition to an interface so the hook function is easier to read.

jodator:
@thesan ps.: please remember to add `See #xxx` or `Closes #xxx` in the PR description to have a direct link to an issue.

If you see that this PR resolves all the requirements of the issue, please use `Closes #xxx`.

thesan:
> @thesan ps.: please remember to add `See #xxx` or `Closes #xxx` in the PR description to have a direct link to an issue.
> 
> If you see that this PR resolves all the requirements of the issue, please use `Closes #xxx`.

Got it, I will keep it in mind :ok_hand:

PR #619, opened 2021-05-17
🚓 Correct `Textarea` CSS syntax

p-sad:
Incorrect CSS syntax in Textarea component. Firefox powers through it and uses the default `resize` style, but in Chrome the site breaks completely as soon as the component tries to render.


jodator:
I didn't notice that on Brave, but anyway the resize does not work as intended there. The other thing is that textarea should be bigger IMO.

Issue #398, opened 2021-04-14
Working group - complete About tab


p-sad:
- [x] Status message label
- [x] Dashboard (spending, total hired, total fired)

jodator:
Looks like done, other things to be done separately.

PR #620, opened 2021-05-17
🍁 Display working group status in About tab

p-sad:
See #398


packages/ui/dev/scripts/generators/generateWorkingGroups.js
```diff
@@ -18,6 +18,8 @@ const generateWorkingGroups = () => {
       description: randomMarkdown(),
       setAtBlockId: randomFromRange(1, 50),
       setAtTime: '2021-03-09T10:28:04.155Z',
+      status: 'Scheduled downtime',
```
jodator, line 21: Let's have that randomized - not every group need to be in downtime :D

I'd go with either:

*   Math.random() > xxx
*   or `id % 3  === 0` to have every 3rd one have a status.

jodator (changes requested):
Good PR @p-sad :) I'd only change the generation script to not have every group have the status message and we can merge that.

packages/ui/dev/scripts/generators/generateWorkingGroups.js
jodator: Additionally, you can use `faker.lorem.words( randomFromRange(2, 3))`

PR #623, opened 2021-05-18
🧝🏼‍♀️ Role details page

DevNikita16:
Closes #623


packages/ui/src/working-groups/queries/workingGroups.graphql
```diff
@@ -25,6 +26,9 @@ fragment WorkerFields on Worker {
   hiredAtBlock {
     ...BlockFields
   }
+  application {
+    ...WorkingGroupApplicationFields
```
jodator, line 30: ~Maybe fetching application id & opening id would be enough?~

```
application {
  id
  opening {
    id
  }
}
```

jodator: However, it would require a bigger change in the modal system so for now it is OK.

jodator (approved):
I would change few things, but overall let's merge it 😄 

General remarks:

- please try to avoid multiple "Apply eslint fixes" in the git log ;) While working on the PR you can always use gti amend option (optionally with push force) to merge it with previous changes.
- no `console.log` in the PR so that's 👍

Issue #300, opened 2021-03-25
Display role details


jodator:
@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4838716)

Display role details:
- display role accounts, dashboard info 
- display role activities
- display role actions: see application, view opening, leave a postion

jodator:
The updated design is here: [https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1740%3A92222.](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1740%3A92222.)

This view might be tricky, so a little heads up:

1.  The account role should be probably very close to that displayed in the "My Accounts" view. There are hooks to display balances (at least total :shrug: as we didn't implement "locked", etc AFAIR).
2.  There are button links to:
    1.  Application (id should be fetched from the query node) -> you'll need to open an application preview modal
    2.  Opening (the same -> from the query node) -> use react-router to open the opening
    3.  Other buttons can be without action for now
3.  The badges are: group name | lead / regular (`isLead`) | worker id ;)

The dashboard can be stubbed for now with static stubs.

The role activities should display a stub as in other components.

jodator:
@DevNikita16 For the issue to be closed we still need:

1.  Better Application link with the Worker (ideally the worker should be generated together with the application). But that's a minor issue and can be skipped if work on that would consume much time.
2.  Buttons on the page should be:
    1.  Change Role account (:white\_check\_mark:)
    2.  A button to move funds (This can be removed and implemented together with #305 - so only remove the button for now)
    3.  Change reward account (:white\_check\_mark: )
3.  The dashboard is missing proper values (ie next payout in...)
4.  Different role states:
    1.  [unstaking](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1753%3A10723) (display that the account is in unstaking phase
    2.  [ended role](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1753%3A13189) (different badges, info that a role has ended) no leave role button 
    3.  [slashed role](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1753%3A11599) (we might need to check how to assert this state)

DevNikita16:
@jodator Do we already have possibility to handle unstaking phase?

jodator:
[...]

PR #626, opened 2021-05-18
🔄 Add move tokens button

DevNikita16:
See #300

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRole.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRole.tsx
@@ -110,9 +110,9 @@ export const MyRole = () => {
         </Statistics>
         <ContentWithSidepanel>
           <MainPanel>
-            <MyRoleAccount name="Role Account" address={worker.roleAccount} />
-            <MyRoleAccount name="Staking Account" address={worker.stakeAccount} />
-            <MyRoleAccount name="Reward Account" address={worker.rewardAccount} />
+            <MyRoleAccount account={{ name: 'Role Account', address: worker.roleAccount }} canChangeAccount={true} />
+            <MyRoleAccount account={{ name: 'Role Account', address: worker.stakeAccount }} canMoveTokens={true} />
+            <MyRoleAccount account={{ name: 'Role Account', address: worker.roleAccount }} canChangeAccount={true} />
           </MainPanel>
           <SidePanel>
             <ActivitiesBlock activities={activities} label="Role Activities" />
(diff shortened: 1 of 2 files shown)


packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRole.tsx
```diff
@@ -110,9 +110,9 @@ export const MyRole = () => {
         </Statistics>
         <ContentWithSidepanel>
           <MainPanel>
-            <MyRoleAccount name="Role Account" address={worker.roleAccount} />
-            <MyRoleAccount name="Staking Account" address={worker.stakeAccount} />
-            <MyRoleAccount name="Reward Account" address={worker.rewardAccount} />
+            <MyRoleAccount account={{ name: 'Role Account', address: worker.roleAccount }} canChangeAccount={true} />
```
jodator, line 113:
```suggestion
           <RoleAccountHeader>
             <Label>Role account</Label>
             {isActive && ( <ButtonsGroup>
                 <ButtonGhost size="small">Change Role account</ButtonGhost>}
             </ButtonsGroup> ) }
          </RoleAccountHeader>
           <MyRoleAccount name="Staking Account" address={worker.stakeAccount} />
```

Something like, above might be better to compose the page instead of making the `<MyRoleAccount />` aware of different actions that it could invoke.

jodator: OMG, I can't properly indent my comment above :roll\_eyes: But the idea is to make the MyRole account unaware of those buttons and compose the page using the components.

I've added `isActive` there to indicate that these buttons should be visible only when the worker is active.

ps.: The button can be also disabled in the unstaking period, but we don't have that one yet.

jodator (changes requested):
I think that the `<MyRoleAccount />` should be simpler. Then, you can compose the "my role page" using simpler components.

packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRole.tsx
```diff
@@ -44,6 +46,8 @@ export const MyRole = () => {
       data: { application: (worker && worker.application) as WorkingGroupApplication },
     })
   }, [worker && worker.application.id])
+  const isActive = worker && worker.status === 'WorkerStatusActive'
+  console.log(isActive)
```
jodator, line 50:
```suggestion
  const isActive = worker && worker.status === 'WorkerStatusActive'
```

:police\_car:

DevNikita16: oops, sorry 🙈

jodator (approved):

PR #633, opened 2021-05-19
Add links looks like buttons

### LINKED ISSUE #510: Create Navigation button component
### Intro

We use the `react-router` to provide routing to various parts of the app. In many places, we use custom `<buttons>` to invoke navigation. Using react router to navigate you can use:

1.  `<Link />` component
2.  `useHistory()` hook

The problem with the link is that it either enforces invalid HTML (you can't nest `<button>` inside `<a>` or it forces you to write more code.

You could use the `<Link component={ButtonPrimary} to="..." />` syntax, but then using props from `ButtonPrimary` is cumbersome.

### Requirements

Create a navigation button that will allow a syntax below:

```jsx
<NavButtonPrimary to="/working-groups" square>
  <Icon />
</NavButtonPrimary>
```

So it should allow us to:

*   use existing button components and their API
*   specify the path that the button should navigate as in `<Link to="" />`

This can be achieved either by updating our button with an optional property (which might be overkill) or by creating a new component that extends the existing buttons.

### LINKED ISSUE #610: Add link button styles
https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1382%3A66254

ClumsyVlad:
Closes #610 

Closes #510 

To avoid using buttons without onClick wrapped by Link to change location on Link click.

### DIFF
--- a/packages/ui/src/working-groups/components/OpeningsList.tsx
+++ b/packages/ui/src/working-groups/components/OpeningsList.tsx
@@ -1,8 +1,9 @@
 import React from 'react'
-import { Link } from 'react-router-dom'
+
+import { LinkButtonGhost } from '@/common/components/buttons/LinkButtons'
 
 import { BadgeViolet } from '../../common/components/BadgeViolet'
-import { ButtonGhost, ButtonPrimary, ButtonsGroup } from '../../common/components/buttons'
+import { ButtonPrimary, ButtonsGroup } from '../../common/components/buttons'
 import { Toggle } from '../../common/components/buttons/Toggle'
 import { Arrow } from '../../common/components/icons'
 import { List, ListItem } from '../../common/components/List'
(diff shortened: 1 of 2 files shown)


packages/ui/src/working-groups/components/OpeningsList.tsx
```diff
@@ -148,12 +149,10 @@ const OpeningDetails = ({ opening }: Props) => {
           </StatsBlock>
         </Statistics>
         <ButtonsGroup align="right">
-          <Link to={`/working-groups/openings/${opening.id}`}>
-            <ButtonGhost size="medium">
-              <Arrow direction="left" />
-              Learn more
-            </ButtonGhost>
-          </Link>
+          <LinkButtonGhost to={`/working-groups/openings/${opening.id}`}>
```
jodator, line 152: :+1:

Issue #304, opened 2021-03-25
Change role and reward accounts


jodator:
@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4838459)

### Requirements

- allow member to change role account
- allow member to change reward account
- use modal system to authorize transactions

jodator:
Closed in #636 & #644

PR #644, opened 2021-05-20
🥇Add change reward account modal

wyrdrender:
See #304 

* Create single modal component for all account changes actions
* Move change role account modal to common changes modal
* Add prepare and sign steps
* Add success and error steps
* Extend sign tests for reward account changing


jodator (approved):
I was thinking about using string enums, but maybe that's not needed ATM.

Other than that I like how you make those two modals in one :)

PR #646, opened 2021-05-20
📅 Implement the Datepicker

### LINKED ISSUE #584: Implement Calendar select
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1372%3A64930)

Prepare the calendar select component. This looks like that there might be some existing components to be used and restyled only. Check the existing solution and consult before implementation.

Only the Date select (without the Block select) for this task.

thesan:
Closes #584


thesan:
Currently you can't write the date manually (with the text input). Should I add this now or it can wait ?

jodator:
> Currently you can't write the date manually (with the text input). Should I add this now or it can wait ?

Yeah, we can move that as a next step. I'll do a review later today so I'll let you know if see anything else.

packages/ui/src/common/components/buttons/FilterButtons.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+import { ButtonGhost, ButtonPrimary } from '@/common/components/buttons'
+
+interface FilterButtonsProps {
+  onClear?: () => void
+  onApply?: () => void
+}
```
jodator, line 8: I think that it might be too early to introduce this component. That's because it is the only one we have for now.

Usually we should extract common parts when we see two or more occurrences of duplicated code. Up-front optimizations might not be needed.

However, if you see that this particular one will be used more let me know.

packages/ui/src/common/components/forms/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,97 @@
...
+import {
+  CalendarMonth,
+  CalendarWeekDay,
+  CalendarHeaderContainer,
+  CalendarDay,
+  CalendarTitle,
+  CalendarNav,
+} from './components'
+
+const DAYS = 6 * 7
```
jodator, line 17: I'm not sure what `DAYS` represents. Could you rename it to something more meaningful?

packages/ui/src/common/components/forms/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+interface CalendarBodyProps {
+  month: Date
+  within?: PartialDateRange
+  selection?: DateSelection
+  onChange?: (day: Date) => void
+}
+
+const CalendarBody = React.memo(({ month, within, selection, onChange }: CalendarBodyProps) => {
+  const firstday = useMemo(() => startOfWeek(addDays(month, -1), { weekStartsOn: 1 }), [month])
```
jodator, line 75:
```suggestion
  const firstDay = useMemo(() => startOfWeek(addDays(month, -1), { weekStartsOn: 1 }), [month])
```

It reads better and is consistent with the naming style.

packages/ui/src/common/components/forms/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <>
+      {['MO', 'TU', 'WE', 'TH', 'FR', 'SA', 'SU'].map((day) => (
```
jodator, line 79: I think that days array can be extracted as a `const`

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,182 @@
...
+  within?: PartialDateRange
+  onChange: (value: PartialDateRange) => void
+}
+const DatePickerCalendars = ({ value, within, onChange }: DatePickerCalendarsProps) => {
+  const { start, end } = fromRange(value)
+
+  const today = useMemo(() => startOfToday(), [])
+  const thisMonth = useMemo(() => startOfMonth(today), [])
+  const [monthL, setMonthL] = useState(start ? startOfMonth(start) : addMonths(thisMonth, -1))
+  const [monthR, setMonthR] = useState(end ? startOfMonth(end) : thisMonth)
```
jodator, line 79: If you could rename `mothL` & `monthR` -> the abbreviation is a bit cryptic. It will be better to be a bit more fluent with variable names, even if longer. I'd guess that is `monthLeft` and `monthRight` but if this has better meaning, like `startMonth` & `endMonth` the better. (Because Left & RIght is a visual aspect of those two variables, but their left/right placement has different meaning.

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,182 @@
...
+interface DatePickerCalendarsProps {
+  value: PartialDateRange
+  within?: PartialDateRange
+  onChange: (value: PartialDateRange) => void
+}
+const DatePickerCalendars = ({ value, within, onChange }: DatePickerCalendarsProps) => {
+  const { start, end } = fromRange(value)
+
+  const today = useMemo(() => startOfToday(), [])
+  const thisMonth = useMemo(() => startOfMonth(today), [])
```
jodator, line 77:
```suggestion
  const currentMonth = useMemo(() => startOfMonth(today), [])
```

Maybe a bit better?

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const [monthL, setMonthL] = useState(start ? startOfMonth(start) : addMonths(thisMonth, -1))
+  const [monthR, setMonthR] = useState(end ? startOfMonth(end) : thisMonth)
+
+  const selectionStrategy = (k1: keyof DateRange) => (day: Date) => {
```
jodator, line 81: The `k1` is again too cryptic. While reading I have no idea what this might mean ;)

However, I see that there's `as`  cast used a lot so maybe the `Date` type might be different :thinking: 

For now, let's keep it this way, I'll have another look when you split the PR to a smaller one.

jodator (changes requested):
First of all the calendar component works pretty well 👏 However, the PR is quite big and I had trouble focusing and properly review it. As a general rule when working on a bigger chunk try to divide your work into smaller chunks.

In this PR the parts with calendar select and a range select might easily be provided as two PRs.

For now, I've added some comments, mostly to provide more context into some variables or align the code style. However, I'd like to ask you first to split this PR into two:

- the date calendar picker (you can just delete all the range code from this PR)
- the range slider branch

I'm sorry for that, but it will be easier to review & merge the required part (calendar) and not throw out your work that was put into the range slider.

thesan:
> First of all the calendar component works pretty well clap However, the PR is quite big and I had trouble focusing and properly review it. As a general rule when working on a bigger chunk try to divide your work into smaller chunks.
> 
> In this PR the parts with calendar select and a range select might easily be provided as two PRs.
> 
> For now, I've added some comments, mostly to provide more context into some variables or align the code style. However, I'd like to ask you first to split this PR into two:
> 
>     * the date calendar picker (you can just delete all the range code from this PR)
> 
>     * the range slider branch
> 
> 
> I'm sorry for that, but it will be easier to review & merge the required part (calendar) and not throw out your work that was put into the range slider.

No problem, I understand. I will try to keep my PR a lot shorter.

I probably should have even split the `<Calendar />` and `<DatePicker />` in two PRs.

packages/ui/src/common/components/buttons/FilterButtons.tsx
thesan: I thought so but then I saw that these buttons are now used on the select filters in the new design [@figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=3034%3A2041). This is why I extracted this into a component.

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
thesan: My understanding from the design is that both calendars can be changed (to some extent) independently from each other and from the date range. But also you could use e.g the left calendar to set both the start and end dates so endMonth could be a bit confusing. In any case I think `leftCalendarMonth` and `rightCalendarMonth` are much clearer names than the previous ones :smile:

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
thesan: Sorry this one was a bit tricky to name, but with `k1` I didn't try very hard :facepalm: 

The multiple casts come from typescript not supporting union types (in this case `'start' | 'end'`) as object keys. Instead it will fallback to inferring { [k: string]: Date } but when casting the key to `'start' | 'end'` it outputs: `An index signature parameter type cannot be a union type`

packages/ui/src/common/components/forms/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,98 @@
+import { addDays, startOfMonth, startOfWeek } from 'date-fns'
+import React, { useMemo } from 'react'
```
wyrdrender, line 2:
```suggestion
import React, { useMemo, memo } from 'react'
```

packages/ui/src/common/components/forms/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,98 @@
...
+)
+
+interface CalendarHeaderProps {
+  month: Date
+  within?: PartialDateRange
+  withinMonths?: PartialDateRange
+  onChange?: (month: Date) => void
+}
+
+const CalendarHeader = React.memo(({ month, within, withinMonths, onChange }: CalendarHeaderProps) => {
```
wyrdrender, line 47:
```suggestion
const CalendarHeader = memo(({ month, within, withinMonths, onChange }: CalendarHeaderProps) => {
```

packages/ui/src/common/components/forms/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,98 @@
...
+const DAY_NAMES = ['MO', 'TU', 'WE', 'TH', 'FR', 'SA', 'SU']
+
+interface CalendarBodyProps {
+  month: Date
+  within?: PartialDateRange
+  selection?: DateSelection
+  onChange?: (day: Date) => void
+}
+
+const CalendarBody = React.memo(({ month, within, selection, onChange }: CalendarBodyProps) => {
```
wyrdrender, line 75:
```suggestion
const CalendarBody = memo(({ month, within, selection, onChange }: CalendarBodyProps) => {
```

packages/ui/src/common/components/forms/Calendar/components.tsx
```diff
@@ -0,0 +1,174 @@
...
+import { DateSelection, PartialDateRange } from '@/common/types/Dates'
+import { fromRange } from '@/common/utils/dates'
+
+interface CalendarNavProps {
+  month: Date
+  direction: -1 | 1
+  min?: Date
+  max?: Date
+  onClick?: (month: Date) => void
+  children?: React.ReactNode
```
wyrdrender, line 16: This is fine but there is no need to explicit declare `children` in Functional Component. Example below is equal and in my opinion more DRY because you don't have to redeclare `children` in every component:
```
import React, { FC } from 'react'
export const CalendarNav: FC<CalendarNavProps> = ({ month, direction, min, max, children, onClick }) => {})
```

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,181 @@
...
+  onClear?: () => void
+  onChange: (value: PartialDateRange) => void
+}
+
+export const DatePicker = ({ value, withinDates, onApply, onClear, onChange }: DatePickerProps) => {
+  const placeholder = '__/__/__'
+  const { start, end } = fromRange(value)
+  const dateString = `${toDDMMYY(start) ?? placeholder} - ${toDDMMYY(end) ?? placeholder}`
+
+  const container = useRef<HTMLDivElement>(null)
```
wyrdrender, line 26: WDYT? This way you don't need to worry if it's `null` or not.
```suggestion
  const container = useRef<HTMLDivElement>(document.createElement('div'))
```

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const [isOpen, toggleOpen] = useState(false)
+
+  useEffect(() => {
+    if (!isOpen) return
+
+    const closePopup = (evt: MouseEvent) => {
```
wyrdrender, line 32: Use meaningful names
```suggestion
    const closePopup = (event: MouseEvent) => {
```

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,181 @@
...
+  display: inline-block;
+`
+
+const DatePickerPopup = styled.div`
+  position: absolute;
+  left: 0;
+  top: calc(100% + 4px);
+  box-shadow: ${Shadows.light};
+  padding: 16px;
+
```
wyrdrender, line 153: Remove empty line
```suggestion
```

wyrdrender (changes requested):

packages/ui/src/common/components/forms/Calendar/components.tsx
thesan: Hi @wyrdrender,

Thanks for the review.

Just for clarifying I also have a slight preference for using `FC` but @jodator told me that [it is less consistent with rest of the project](https://github.com/Joystream/pioneer/pull/468#discussion_r621876190). That said, that other component wasn't rendering any children... So I can be clear on this point, is it consistent with the project code base to use the `FC` type only when declaring a functional component which renders children ?

packages/ui/src/common/components/forms/DatePicker/DatePicker.tsx
thesan: I think this would be fine here. But I generally would prefer being aware that a reference might be `null` (at least on the first render) than having the code potentially interact with a `div` which will get discarded. In my opinion it might be especially confusing in some cases when the `ref` is passed down other components.

packages/ui/src/common/components/forms/Calendar/components.tsx
jodator: > So I can be clear on this point, is it consistent with the project code base to use the `FC` type only when declaring a functional component which renders children ?

I'm for this :+1: 

This way we have a clear definition of whether the component receives children or not.

wyrdrender (approved):

jodator (approved):

Issue #301, opened 2021-03-25
Display activity details


jodator:
@[figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1636%3A2158308)

Display activity detail

jodator:
Comments on which activity to handle:

*   https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1405%3A69471

p-sad:
Designs distinguish between budget increase and decrease, but that can't be discerned from a single event.
It is possible to infer it from a collection of past events, but not sure how to implement it for activities in an efficient matter.

p-sad:
* Approach I took so far (events 1:1 activities) may actually be insufficient, seeing that a `StatusTextChangedEvent` can contain both a Status Changed activity _and_ zero or more Upcoming Openings.
* The design of "Reward amount changed" activities is vague and I'm wondering if it's on purpose, considering that the event specifies the monthly reward of a specific worker of a specific group, meaning that the generic design doesn't communicate any useful information — unless it will only be seen in Role preview (of a specific role)

Other than "Reward amount changed" and "Upcoming Opening", the Working Group activities are pretty much laid out

PR #651, opened 2021-05-21
🧨 Set up activities logic

p-sad:
Paving the way for #301 
This is a proof of concept for how activities could be represented in the application.


packages/ui/src/common/components/Activities/ActivityContent.tsx
```diff
@@ -0,0 +1,34 @@
...
+    case 'AppliedOnOpeningEvent':
+      return (
+        <>
+          <Link to="#" onClick={() => showModal({ modal: 'Member', data: { id: activity.membership.id } })}>
+            {activity.membership.handle}
+          </Link>{' '}
+          has applied on the opening{' '}
+          <Link to={`/working-groups/openings/${activity.opening.id}`}>{activity.opening.title}</Link>.
+        </>
+      )
```
p-sad, line 24: Might be worth wrapping these in separate components and returning only the extracted component, good/bad idea?

jodator (changes requested):
@p-sad - just to change the PR state on GH. A short summary of a F2F review:

The activities component has a great potential to grow, so we need to prepare base activity components for an extension without modifying the component's code.

packages/ui/src/common/components/Activities/ActivityContent.tsx
```diff
@@ -0,0 +1,27 @@
...
+      return <AppliedOnOpeningContent activity={activity} />
+    case 'BudgetSpendingEvent':
+      return <BudgetSpendingContent activity={activity} />
+    case 'ApplicationWithdrawnEvent':
+      return <ApplicationWithdrawnContent activity={activity} />
+    case 'BudgetSetEvent':
+      return <BudgetSetContent activity={activity} />
+    default:
+      return <div />
+  }
```
jodator, line 26: Not great not terrible ;) Let's refactor that later when we have more activities from other modules.

packages/ui/src/working-groups/types/WorkingGroupActivity.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface ApplicationWithdrawnActivity extends BaseActivity {
+  eventType: 'ApplicationWithdrawnEvent'
+  membership: {
+    id: string
+    handle: string
+  }
+  opening: {
+    id: string
+    title: string
+  }
```
jodator, line 20: Let's start with `Pick<Member, 'id' | 'handle'>`.

jodator (changes requested):
Last small batch of changes before merge.

Issue #613, opened 2021-05-17
Allow linking to member profile


jodator:
See #50.

The copy button was added in #602.

We need to allow to link to a member profile. Because this sidebar can be shown from various places the link should always link to the members subsection of the App.

jodator:
There's `MemberModalLink` added in #662. I think that it should be removed by this.

PR #662, opened 2021-05-24
🪑 Add convenience components for activities

p-sad:
See: #301 
The main feature of this one is the ModalLink component. I tried to enforce decent type control in it - it works, but doesn't look very elegant IMO

### DIFF
--- a/packages/ui/src/common/components/ModalLink.tsx
+++ b/packages/ui/src/common/components/ModalLink.tsx
@@ -1,11 +1,12 @@
 import React from 'react'
 import { ReactNode } from 'react-markdown'
-import { Link } from 'react-router-dom'
 
 import { MemberModalCall } from '@/memberships/components/MemberProfile'
 
 import { useModal } from '../hooks/useModal'
 import { AnyModalCall } from '../providers/modal/types'
+
+import { Link } from './Link'
 
 interface ModalLinkProps<Call> {
   call: Call extends AnyModalCall ? Call : never
@@ -16,7 +17,6 @@ const ModalLink = ({ call, children }: ModalLinkProps<AnyModalCall>) => {
   const { showModal } = useModal()
   return (
     <Link
-      to="#"
       onClick={(evt) => {
         evt.preventDefault()
         showModal(call)
--- a/packages/ui/src/common/components/ModalLink.tsx
+++ b/packages/ui/src/common/components/ModalLink.tsx
@@ -1,0 +1,30 @@
+import React from 'react'
+import { ReactNode } from 'react-markdown'
+import { Link } from 'react-router-dom'
+
+import { MemberModalCall } from '@/memberships/components/MemberProfile'
+
+import { useModal } from '../hooks/useModal'
+import { AnyModalCall } from '../providers/modal/types'
+
+interface ModalLinkProps<Call> {
+  call: Call extends AnyModalCall ? Call : never
+  children: ReactNode
+}
+
+const ModalLink = ({ call, children }: ModalLinkProps<AnyModalCall>) => {
+  const { showModal } = useModal()
+  return (
+    <Link
+      to="#"
+      onClick={(evt) => {
+        evt.preventDefault()
+        showModal(call)
+      }}
+    >
+      {children}
+    </Link>
+  )
+}
+
+export const MemberModalLink = (props: ModalLinkProps<MemberModalCall>) => ModalLink(props)
--- a/packages/ui/src/common/components/ModalLink.tsx
+++ b/packages/ui/src/common/components/ModalLink.tsx
@@ -1,8 +1,6 @@
 import React from 'react'
 import { ReactNode } from 'react-markdown'
 
-import { MemberModalCall } from '@/memberships/components/MemberProfile'
-
 import { useModal } from '../hooks/useModal'
 import { AnyModalCall } from '../providers/modal/types'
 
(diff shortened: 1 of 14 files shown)


packages/ui/src/common/components/ModalLink.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { ReactNode } from 'react-markdown'
+import { Link } from 'react-router-dom'
```
jodator, line 3: This should be "our" link (`<a>` element, not the one from the react router):

```jsx
import { Link } from '@/common/components/Link'
```

However, the `Link` type must be upgraded to:

```jsx
import React, { AnchorHTMLAttributes } from 'react'
import styled from 'styled-components'

import { Colors, Transitions } from '../constants'

interface LinkProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
  dark?: boolean
  size?: 'm' | 'l'
}
```

packages/ui/src/common/components/ModalLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface ModalLinkProps<Call> {
+  call: Call extends AnyModalCall ? Call : never
+  children: ReactNode
+}
+
+const ModalLink = ({ call, children }: ModalLinkProps<AnyModalCall>) => {
+  const { showModal } = useModal()
+  return (
+    <Link
+      to="#"
```
jodator, line 19: After the above changes :point\_up: you can remove `to`.

packages/ui/src/common/components/ModalLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+        evt.preventDefault()
+        showModal(call)
+      }}
+    >
+      {children}
+    </Link>
+  )
+}
+
+export const MemberModalLink = (props: ModalLinkProps<MemberModalCall>) => ModalLink(props)
```
jodator, line 30: We shouldn't introduce domain-specific components inside general ones. This should be removed from this file.

jodator (changes requested):
I've also seen that the storybook fails to load since there's no React Router (might be dummy MemoryRouter) in the Activities stories.

The other thing is to improve ModalLiknk types a bit.

packages/ui/src/common/components/ModalLink.tsx
jodator: You can also have something like this:

```jsx
const MemberModalLink: IModalLink<MemberModalCall> = ModalLink
```

jodator: The reason to have an "alias" as above is that it does not create another function ([TS playground](https://www.typescriptlang.org/play?#code/AQSwdgLgpgTgZgQwMZWAWQPYBMEBsDCeuAPGgHzADeAsAFDDAC22eAXOnQL512iSyIU6FrgDqICAAsAIgggJCuEmgA0waRSgAPaGCwBnYTgJFSFGvWA557aVx70IATwAOqAIJgnmY4tzAAXiM8P2J9CBhwAHMKAB9gsQkZOQVTcMiwKLUwKAA3WDI6cGh4ZFQfPAAZcABrAAUYDBd9Yj9zXmAkInY-YG1dA2BPbxFegH5gXvYc-JgOpEkQXCwYKDB2BC97WjpnN2AAMXxiABUKIIAKF2BWE4BKQIpcjBAsB2A91ABJCtxqsBqrSI50Ox1+-3qjWaQKUZEKO1oSAwYHCCQhgWAMP8-TWg2GvzaV3Y4NqDSaLTaDwCFAArnooHBwFA3gjPugoIwAEawAlEDG-cRSWTyUIAcjQHO5MFFakooCw7HS0WAnHhdAA9OqhoY8CAEPo6EiURB2VyeSIIewfhbaqRJebfMD+TaAQ5NdrgAgYI0AO7AOB0pASZGG5GoiVmmAkgHufS9S4uKH6YkuyHku2R3mwqkUaM1K5Jh5AA)).

packages/ui/src/working-groups/components/Activities/ApplicationWithdrawnContent.tsx
```diff
@@ -1,24 +1,17 @@
...
-      </Link>{' '}
-      has withdrawn application from "<Link to={`/working-groups/openings/${opening.id}`}>{opening.title}</Link>"
-      opening.
-    </>
-  )
-}
+export const ApplicationWithdrawnContent = ({ activity: { membership, opening } }: Props) => (
+  <>
+    <MemberModalLink call={{ modal: 'Member', data: { id: membership.id } }}>{membership.handle}</MemberModalLink> has
+    withdrawn application from "<Link to={`/working-groups/openings/${opening.id}`}>{opening.title}</Link>" opening.
```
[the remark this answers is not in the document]
p-sad: So far I went with C, I'm also partial to B despite it being quite messy indeed. I just feel like the type/component itself should enforce the types of arguments it accepts. It may be strictness for strictness' sake though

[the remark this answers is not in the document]
jodator: I think that this one might be temporary nevertheless #613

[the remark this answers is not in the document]
jodator: I think that this one might be temporary nevertheless #613

PR #663, opened 2021-05-25
💌 Encode metadata fields

### LINKED ISSUE #501: Check types of the string-like metadata in transactions
It looks like types used to provide text descriptions for some extrinsic changed to `Bytes` from previously defined.

For instance:

```
export interface BuyMembershipParameters extends Struct {
  readonly root_account: AccountId;
  readonly controller_account: AccountId;
  readonly name: Option<Text>;
  readonly handle: Option<Text>;
  readonly avatar_uri: Option<Text>;
  readonly about: Option<Text>;
  readonly referrer_id: Option<MemberId>;
}
```

changed into:

```
/** @name BuyMembershipParameters */
export interface BuyMembershipParameters extends Struct {
  readonly root_account: AccountId;
  readonly controller_account: AccountId;
  readonly handle: Option<Text>;
  readonly metadata: Bytes;
  readonly referrer_id: Option<MemberId>;
}
```

So the `name`, `about` and `avatar_uri` were migrated to `metadata` as `Bytes`.

### LINKED ISSUE #617: Typechecking errors with DetailedMember
DetailedMember is not a superset of Member so components that expect a Member in their props show up errors in IDEs.
Interestingly, it still compiles and runs, but can cause unexpected runtime errors in the future.

jodator:
Closes #501

Closes #617

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,4 +37,5 @@ After updating `packages/ui/src/api` any of `*.graphql` files run `yarn queries:
 ## Packages
 
 - [packages/ui](packages/ui) The Pioneer 2.0 package
-- [packages/type](packages/types) Copied [@joystream/types](https://github.com/Joystream/joystream/tree/master/types) at a compatible Olympia branch.
+- [packages/type](packages/types) Copied [@joystream/types](https://github.com/Joystream/joystream/tree/olympia/types) at a compatible Olympia branch.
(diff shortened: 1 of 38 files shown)


p-sad (commented):
My greatest contribution yet, a bunch of spellchecks.
Other than that, looks good

Issue #585, opened 2021-05-14
Add filters to the past proposals page


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2599%3A109614)

### Requirements

Implement the desing and the base components for past proposals filters.

Fields
- Stage: TBD  (can be empty for now)
- Type: Create using simple selects
- Search box
- Proposer: TDB (can be empty for now)
- Ended: Prepare calender select (this might be a bigger task to prepare). (requires: https://github.com/Joystream/pioneer/issues/584)

PR #667, opened 2021-05-25
🔬 Implement a Filter box component

thesan:
See #585


jodator (approved):
I've found one visual problem introduced by using a button. Changing it to `<a>` should fix that.

I've also created a follow-up for the members page: https://github.com/Joystream/pioneer/issues/690 as you worked on both members search and filter box.

Issue #305, opened 2021-03-25
Move funds to locked acount


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1405%3A83725)

### Requirements

- Allow user to move funds into a locked balances

jodator:
We have two modals/check:

1.  Move funds
2.  Move funds to a locked account

### 1\. Move funds

Let's say we have:

*   a required stake of `1000 JOY`
*   accounts:
    *   A with free balance `200 JOY`
    *   B with a free balance `500 JOY`
    *   C with a free balance `400 JOY`
    *   D with a free balance `0 JOY`

So none of the user accounts have a total `1000 JOY` of required stake but we can combine balance from accounts `A`, `B` and `C`.

The button should move you to the "accounts page" to manually manage accounts.

### 2\. Move funds to a locked account

Let's say we have:

*   a required stake of `1000 JOY`
*   accounts:
    *   A with free balance `0 JOY` and `200 JOY` locked (staked as a Validator, etc)
    *   B with a free balance `500 JOY`
    *   C with a free balance `0 JOY` and `200 JOY` locked
    *   D with a free balance `300 JOY`

In this case, none of the accounts has the required total balance (AFAIR it is locked + free) and you can't combine free balances to cover the stake. However, accounts `A` and `C` have enough tokens locked and moving tokens from `B` and `D` would create a total balance of `1000 JOY`.

## Technical details

*   To check account balance there's the `useBalance()` hook that should return proper balances (some are stubbed ATM, but being work on in #615.
*   There's also a combined hook (`useTotalBalances()`) that returns a total summary of the user funds (it just adds the balances from all of the user's accounts).
*   The checks should have tests.
*   The locks logics is more or less this: you can have multiple locks (`200`, `200`, `300`) but the total locked value is `300` - they just lock part of the funds so the funds are reusable. It should be safe to assume that the mentioned hooks return proper values.

jodator:
Closed with #668.

PR #668, opened 2021-05-25
💰Add move founds modal

wyrdrender:
See #305

*Display proper modal (transferable founds or locked founds) depends on your balance
*Redirect to profile page from modal


packages/ui/src/accounts/modals/MoveFoundsModal/MoveFoundsTransferableModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { spacing } from '@/common/utils/styles'
+
+import { ModalBody, MemberRow } from './styles'
+
+interface Props {
+  onClose: () => void
+  onManageAccountsClick: () => void
+  price: number
+  balances: AddressToBalanceMap
+  allAccounts: Account[]
```
jodator, line 20: I think that this modal should obtain as a prop set of the accounts to render.

This modal has only an informational character, the actual check might be invoked earlier in other steps of the application. That's because we need to inform the user that there are transferable funds that can be used even though none of the accounts have the required stake  ([the move funds case](https://github.com/Joystream/pioneer/issues/305#issuecomment-845803741)).

In other words, we might have actions:

*   [apply for a role](https://joystream.gitbook.io/joystream-handbook/governance/working-groups#staking)
*   [vote](https://joystream.gitbook.io/joystream-handbook/governance/council#vote) in elections
*   [create a proposal](https://joystream.gitbook.io/joystream-handbook/governance/proposals#staking)
*   apply as a candidate

and all of them will require some `stake` to be locked when issuing an action. The App will check the required stake and if it finds out that case 1 or 2 is met then it will display the `MoveFundsTransferableModal` or `MoveFundsLockedModal`.

packages/ui/src/accounts/modals/MoveFoundsModal/MoveFoundsTransferableModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { TextMedium, TokenValue } from '@/common/components/typography'
+import { Colors } from '@/common/constants'
+import { spacing } from '@/common/utils/styles'
+
+import { ModalBody, MemberRow } from './styles'
+
+interface Props {
+  onClose: () => void
+  onManageAccountsClick: () => void
+  price: number
```
jodator, line 18: This should be named: `requiredStake` as the name reflects the domain naming. See below :point\_down: . Some of the actions invoked by the users require [staking](https://joystream.gitbook.io/joystream-handbook/key-concepts/staking). If the user has no account with a required stake then the app should detect if there's is any combination of the accounts' balances that allows gathering required stake.

jodator (changes requested):
I think that we need to extract the logic from those modals. The rationale for this is that the check will be more frequent than displaying modals and it will actually precede the modals.

To finish this PR I would see that:
- the modals do not calculate anything
- they accept a list of accounts (move funds case) or accounts + relations (locked case)
- having this in mind it _should be_ possible to create a story for them

The following PR should have:
- a business logic + tests for detecting case 1
- a business logic + test for detecting case 2

After that, we can then integrate them together in the existing modals (3rd PR or a follow up issue) like in Apply for a Role.

--

The only problem with the storybook is that it is not connected to the blockchain IRC so the `useBalance()` will return no value thus the values will be empty ("-"). However, this might be OK to just see how the components are tied together. And they could display incorrect data because we're going to have the tests for the logic in different places.

I would also encourage you to create even smaller PRs in the future. Here we have two modals under one issue but those can be created separately on top of each other.

Issue #670, opened 2021-05-25
Implement the block range part of the Datepicker


thesan:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1372%3A64930)

PR #671, opened 2021-05-25
🏄 Range slider

thesan:
See #670


jodator:
Thank you! I will close the PR for now so we don't have opened PRs. Your branch will be there if we're going to merge it later.

PR #672, opened 2021-05-26
⛓ Simple link components

ClumsyVlad:


p-sad (commented):
There seems to be some overlap with `@/common/components/Link` (though that one opens the link in a new tab), is that by design/should it be integrated into the new ones?

jodator:
> There seems to be some overlap with `@/common/components/Link` (though that one opens the link in a new tab), is that by design/should it be integrated into the new ones?

@p-sad Do you mean that there's multiple link components?

ps.: @ClumsyVlad I see that this was inside `@common/components/buttons/Links` so the name suggests something different.

Issue #583, opened 2021-05-14
Past proposals - list


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2599%3A109614)

### Requirements

- display the past proposals list (past proposals have similar styles to past openings)
- Add stub hook for past proposals (eg. `usePastProposals()`)

PR #673, opened 2021-05-26
🕒 Past Proposals List

thesan:
See #583

### DIFF
--- a/packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
+++ b/packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
@@ -1,42 +1,29 @@
 import React from 'react'
-import styled from 'styled-components'
 
 import { List } from '@/common/components/List'
 import { ListHeader, ListHeaders } from '@/common/components/List/ListHeader'
-import { Colors } from '@/common/constants'
 import { ProposalColLayout } from '@/proposals/constants'
-import { PastProposal, Proposal } from '@/proposals/types'
+import { Proposal } from '@/proposals/types'
 
 import { ProposalListItem } from './ProposalListItem'
 
-export type ProposalListProps = CurrentProposalList | PastProposalList
-interface CurrentProposalList {
+export interface ProposalListProps {
   proposals: Proposal[]
-  past?: false
-}
-interface PastProposalList {
-  proposals: PastProposal[]
-  past: true
+  isPast?: boolean
 }
 
-export const ProposalList = (props: ProposalListProps) => (
-  <ListContainer past={props.past}>
+export const ProposalList = ({ proposals, isPast }: ProposalListProps) => (
+  <>
     <ListHeaders colLayout={ProposalColLayout}>
-      <ListHeader>{props.past ? 'Created' : 'Ended'}</ListHeader>
+      <ListHeader>{isPast ? 'Ended' : 'Created'}</ListHeader>
       <ListHeader>Stage</ListHeader>
       <ListHeader>Type</ListHeader>
       <ListHeader>Proposer</ListHeader>
     </ListHeaders>
-    <List>{props.past ? pastProposalItems(props.proposals) : proposalItems(props.proposals)}</List>
-  </ListContainer>
+    <List>
+      {proposals.map((proposal) => (
+        <ProposalListItem key={proposal.id} {...proposal} isPast={isPast} />
+      ))}
+    </List>
+  </>
 )
-
-const proposalItems = (proposals: Proposal[]) =>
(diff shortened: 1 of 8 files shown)


packages/ui/src/proposals/types/proposals.ts
```diff
@@ -19,3 +19,13 @@ export interface Proposal {
...
+
+export interface PastProposal {
+  id: string
+  endedAt: string
+  title: string
+  rationale: string
+  stage: ProposalStage
+  type: string
+  proposer: Member
+}
```
jodator, line 31: It looks like most of the fields are duplicated from a proposal. I think that we can safely assume that a proposals will have the same "type". THe past proposal will be a proposal that is in one of the stages that do not indicate that this is a current proposal (see [stages](https://github.com/Joystream/pioneer/pull/674/files)).

Let's have:

```suggestion
export interface PastProposal extends Proposal {
  endedAt: string
}
```

or just optional field on the `Proposal` interface.

packages/ui/src/proposals/constants/proposals.ts
```diff
@@ -0,0 +1,5 @@
+import { ProposalStage } from '../types'
+
+export const proposalStages: ProposalStage[] = ['DECIDING', 'DORMANT', 'GRACING', 'SUCCEEDED']
```
jodator, line 3: This looks like duplicating the `ProposalStage` type from \`[packages/ui/src/proposals/types/proposals.ts](https://github.com/Joystream/pioneer/pull/673/files#diff-a8741635086cb8ca5e34592af0f1627dbe7f55e8828ffe1bac1cd4113182cca8) \`.

packages/ui/src/proposals/constants/proposals.ts
```diff
@@ -0,0 +1,5 @@
...
+
+export const proposalTypes = ['FUNDING REQUEST']
```
jodator, line 5: This should be as `ProposalType` type I think. You can also use it for the `Proposal.type` instead of `string`.

jodator: But if you use those in the mocks only, please move them to the mocks so they would not interfere with the code. If so the constants file can be moved to `mocks` without the changes.

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
```diff
@@ -1,28 +1,42 @@
...
 
-export interface ProposalListProps {
+export type ProposalListProps = CurrentProposalList | PastProposalList
+interface CurrentProposalList {
   proposals: Proposal[]
+  past?: false
+}
+interface PastProposalList {
+  proposals: PastProposal[]
+  past: true
```
jodator, line 19: I think that you can use `Proposal` type with optional `endedAt`. This way the list will get a bit simpler.

Having only `isPast` type to decide which field to use to display a date:

```tsx
const pastProposals = usePastProposals()

<ProposalsList proposals={pastProposals} isPast />

// or

const proposal = useProposals()

<ProposalsList proposals={proposals} />
```

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
```diff
@@ -1,28 +1,42 @@
...
 )
+
+const proposalItems = (proposals: Proposal[]) =>
+  proposals.map(({ createdAt, ...props }) => <ProposalListItem key={props.id} date={createdAt} {...props} />)
+
+const pastProposalItems = (proposals: PastProposal[]) =>
+  proposals.map(({ endedAt, ...props }) => <ProposalListItem key={props.id} date={endedAt} {...props} />)
+
+const ListContainer = styled.div`
+  --items-background-color: ${({ past }: { past?: boolean }) => (past ? Colors.Black[50] : Colors.White)};
```
jodator, line 41: Please don't use CSS variables. We don't use them in the App and you should be able to achieve the same using styled-components:

```tsx
export const ListContainer = styled.div<{ isPast?: boolean }>`
  ${ ProposalItem } {
    background-color: ${({ isPast }) => (isPast ? Colors.Black[50] : Colors.White)};
  }
`
```

jodator (changes requested):
hi @thesan. Thanks for the PR :) I think that we can use single `Proposal` interface for now as the current / past proposals would look the same but will have only different stage and end date probably.

I like how the stories work 👍 I added some comments on simplifying the code when using single interface and some styled-components hints.

packages/ui/src/proposals/constants/proposals.ts
thesan: Thanks for the review.

You're right these constants are only used by the `mockProposals` code so I just moved them there.

jodator (approved):

PR #674, opened 2021-05-26
🔥 Hot fixes

jodator:
I've found various minor issues in the code. I don't like this PR, it looks like we broke various things when introducing the babel-loader and weren't aware of this. Hence, after reverting that change the multitude of errors emerged :sob:


packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
```diff
@@ -21,9 +24,16 @@ type Tab = 'GENERAL' | 'FORM'
...
-    modalData: { application },
+    modalData: { applicationId },
   } = useModal<ApplicationDetailsModalCall>()
 
+  const { loading, data } = useGetWorkingGroupApplicationQuery({
+    variables: { where: { id: applicationId } },
+  })
+  const application = data?.workingGroupApplicationByUniqueInput
+    ? asApplication(data.workingGroupApplicationByUniqueInput)
+    : null
```
p-sad, line 35: I guess this could be extracted and wrapped in a separate hook, further down the road

p-sad (approved):
Looks okay, though some component logic could be extracted into a hook

packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
jodator: Yeah, I'll write this down for the upcoming openings fix. Thanks!

PR #680, opened 2021-05-26
🔣 Additional Working Groups Activity cards

p-sad:
See #301


jodator:
@p-sad I've refactored icons a bit:

1.  All have duplicated code and there's base `<Icon />` component already
2.  I've changed their type because `React.FC` expects children while `Icon` should not. I was doing that map in a haste before. I think that now we have less code ;)
3.  Removed repeated interface

PR #681, opened 2021-05-27
🕰️ Open the Past Proposals tab

### LINKED ISSUE #583: Past proposals - list

thesan:
Closes #583

### DIFF
--- a/packages/ui/src/app/pages/Proposals/PastProposals.tsx
+++ b/packages/ui/src/app/pages/Proposals/PastProposals.tsx
@@ -1,0 +1,49 @@
+import React from 'react'
+
+import { AppPage } from '@/app/components/AppPage'
+import { ActivitiesBlock } from '@/common/components/Activities/ActivitiesBlock'
+import { Loading } from '@/common/components/Loading'
+import { ContentWithSidepanel, MainPanel, SidePanel } from '@/common/components/page/PageContent'
+import { PageHeader } from '@/common/components/page/PageHeader'
+import { PageTitle } from '@/common/components/page/PageTitle'
+import { useActivities } from '@/common/hooks/useActivities'
+import { AddProposalButton } from '@/proposals/components/AddProposalButton'
+import { NoProposals } from '@/proposals/components/NoProposals'
+import { ProposalList } from '@/proposals/components/ProposalList'
+import { usePastProposals } from '@/proposals/hooks/usePastProposals'
+
+import { ProposalsTabs } from './components/ProposalsTabs'
+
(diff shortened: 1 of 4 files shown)


packages/ui/src/app/pages/Proposals/PastProposals.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <ContentWithSidepanel>
+          <MainPanel>
+            <ProposalList proposals={proposals} isPast />
+          </MainPanel>
+          <SidePanel>
+            <ActivitiesBlock activities={activities} label="Proposals Activities" />
+          </SidePanel>
+        </ContentWithSidepanel>
+      ) : (
+        <MainPanel>
```
jodator, line 43: One thing – I think that we could have activities and proposals load independently. So maybe we should start working on this sooner? WDYT?

ATM activities are static (since we're waiting for the schema upgrade) but will be also loaded from GraphQL.

I would go with this layout:

```tsx
<ContentWithSidePanel>
  <MainPanel>{isLoading ? <Loadaing /> : ... }</MainPanel>
  <SidePanel><ActivitiesBlock ... /></SidePanel />
</ContentWithSidePanel />
```

jodator: Having this we will display something while the other thins will be loading.

Activities will remain static for now :man\_shrugging:

jodator (changes requested):
Hi, @thesan I really like to review such small PRs 👍 Only one thing to change as I figured out that we might load proposals and activities independently.

packages/ui/src/app/pages/Proposals/PastProposals.tsx
thesan: You're right I got confused because of the no current proposals case from the wiredframe. But I think even for the current proposal the sidebar could remain until we know there's no current proposals. WDYT ?

jodator (approved):

packages/ui/src/app/pages/Proposals/PastProposals.tsx
jodator: Yeah, exactly :+1:

jodator: Yeah, exactly :+1:

PR #689, opened 2021-05-27
😩 Fix side panel fail

ClumsyVlad:


ClumsyVlad:
Merged because we had failure build on main branch

PR #694, opened 2021-05-27
💸 Add hook that checks if the account has enough funds or if it can b…

wyrdrender:
…e transferred

### DIFF
--- a/packages/ui/src/accounts/hooks/useHasRequiredStake.ts
+++ b/packages/ui/src/accounts/hooks/useHasRequiredStake.ts
@@ -27,19 +27,26 @@ export const useHasRequiredStake = (stake: number) => {
   const accountsWithLockedFounds: { [key in Address]: Account[] } | null = useMemo(() => {
     if (
       allAccounts.length &&
-      hasRequiredStake !== true &&
+      hasRequiredStake === false &&
       totalTransferableToNumber <= stake &&
       totalLockedToNumber > 0 &&
       totalLockedToNumber + totalTransferableToNumber >= stake
     ) {
       return allAccounts.reduce((accounts, currentAccount) => {
+        const accountBalance = balances[currentAccount.address]
         const subAccounts = allAccounts.filter(
           (subAccount) =>
             balances[subAccount.address] &&
-            balances[subAccount.address].locked.toNumber() > 0 &&
+            balances[subAccount.address].transferable.toNumber() > 0 &&
             subAccount.address !== currentAccount.address
         )
-        if (!subAccounts.length) return { ...accounts }
+        const totalTransferable = subAccounts.reduce((a, b) => a + balances[b.address].transferable.toNumber(), 0)
+        if (
+          !subAccounts.length ||
+          accountBalance.locked.toNumber() <= 0 ||
+          totalTransferable + accountBalance.locked.toNumber() < stake
+        )
+          return { ...accounts }
         return {
           ...{
             [currentAccount.address]: subAccounts,
@@ -52,7 +59,7 @@ export const useHasRequiredStake = (stake: number) => {
   }, [hasRequiredStake, totalTransferableToNumber, totalLockedToNumber, allAccounts])
 
   const transferableAccounts: Account[] | null = useMemo(() => {
-    if (allAccounts.length && hasRequiredStake !== true && totalTransferableToNumber >= stake) {
+    if (allAccounts.length && hasRequiredStake === false && totalTransferableToNumber >= stake) {
       return allAccounts.filter(
         (account) => balances[account.address] && balances[account.address].transferable.toNumber() > 0
       )
--- a/packages/ui/src/accounts/hooks/useHasRequiredStake.ts
+++ b/packages/ui/src/accounts/hooks/useHasRequiredStake.ts
@@ -28,6 +28,7 @@ export const useHasRequiredStake = (address: Address, stake: number) => {
       allAccounts.length &&
       hasRequiredStake !== true &&
       totalTransferableToNumber <= stake &&
+      totalLockedToNumber > 0 &&
       totalLockedToNumber + totalTransferableToNumber >= stake
     ) {
       return allAccounts.reduce((accounts, currentAccount) => {
--- a/packages/ui/src/accounts/hooks/useHasRequiredStake.ts
+++ b/packages/ui/src/accounts/hooks/useHasRequiredStake.ts
@@ -1,0 +1,66 @@
+import { useEffect, useState, useMemo } from 'react'
+
+import { useAccounts } from '@/accounts/hooks/useAccounts'
+import { useBalance } from '@/accounts/hooks/useBalance'
+import { useBalances } from '@/accounts/hooks/useBalances'
+import { useTotalBalances } from '@/accounts/hooks/useTotalBalances'
+import { Account } from '@/accounts/types'
+import { Address } from '@/common/types'
+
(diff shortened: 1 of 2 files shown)


packages/ui/test/accounts/hooks/useHasRequiredStake.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+    const { result } = renderUseTotalBalances(1000)
+    expect(result.current).toStrictEqual({
+      hasRequiredStake: false,
+      transferableAccounts: null,
+      accountsWithLockedFounds: null,
+    })
+  })
+
+  it('Returns transferable accounts if exists and if account has not enough founds', () => {
```
jodator, line 125:
```suggestion
  it('Not enough funds with transferrable funds', () => {
```

packages/ui/test/accounts/hooks/useHasRequiredStake.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+    const { result } = renderUseTotalBalances(1000)
+    expect(result.current).toStrictEqual({
+      hasRequiredStake: false,
+      transferableAccounts: [aliceStash, bob],
+      accountsWithLockedFounds: null,
+    })
+  })
+
+  it('Returns accounts with locked founds if exists and if account has not enough founds', () => {
```
jodator, line 155:
```suggestion
  it('Not enough funds with transferrable to an account with locked founds', () => {
```

packages/ui/test/accounts/hooks/useHasRequiredStake.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+      <MockKeyringProvider>
+        <AccountsContextProvider>
+          <ApiContext.Provider value={useApi}>{children}</ApiContext.Provider>
+        </AccountsContextProvider>
+      </MockKeyringProvider>
+    )
+    return renderHook(() => useHasRequiredStake(alice.address, stake), { wrapper })
+  }
+
+  it('Returns true flag if account has enough founds', () => {
```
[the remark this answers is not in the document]
jodator: WDYT about that? Just a hint, so you can dismiss that. We fund such shorter descriptions easier to read and less tied to implementation but describing test case.

packages/ui/src/accounts/hooks/useHasRequiredStake.ts
```diff
@@ -0,0 +1,67 @@
+import { useEffect, useState, useMemo } from 'react'
+
+import { useAccounts } from '@/accounts/hooks/useAccounts'
+import { useBalance } from '@/accounts/hooks/useBalance'
+import { useBalances } from '@/accounts/hooks/useBalances'
+import { useTotalBalances } from '@/accounts/hooks/useTotalBalances'
+import { Account } from '@/accounts/types'
+import { Address } from '@/common/types'
+
+export const useHasRequiredStake = (address: Address, stake: number) => {
```
jodator, line 10: This check will be performed before the user selects an account for staking, so we can't assume one account only.

jodator (changes requested):
The `useHasRequiredStake()` should check every account (address) for the User, not only one provided.

The reason is that a member can choose any account to stake. Most of the checks should anticipate the transaction intention and check if most of the requirements are met before creating it.

packages/ui/test/accounts/hooks/useHasRequiredStake.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+  useBalance: () => balance,
+}
+
+jest.mock('../../../src/accounts/hooks/useBalance', () => useBalance)
+
+let balances: AddressToBalanceMap = {}
+
+const useBalances = {
+  useBalances: () => balances,
+}
```
jodator, line 38: This hook is not used in tests, and the `balances` can be inlined AFAICS.

packages/ui/test/accounts/hooks/useHasRequiredStake.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+        recoverable: new BN(0),
+        locks: [],
+      },
+      [bob.address]: {
+        total: new BN(20),
+        locked: new BN(20),
+        transferable: new BN(0),
+        recoverable: new BN(0),
+        locks: [],
+      },
```
jodator, line 121:
```suggestion
    balances = {
      [alice.address]: mockBalances(10, 0, 10, 0),
      [aliceStash.address]: mockBalances(50, 0, 50, 0),
      [bob.address]: mockBalances(20, 20, 0, 0),
```

WDYT about some helper? The balances object is big and hard for me to check the setup vs the expected part

jodator:
```suggestion
    balances = {
      [alice.address]: {
        ...zeroBalance,
        total: new BN(10),
        transferable: new BN(10),
      },
      [aliceStash.address]: {
        ...zeroBalance,
        total: new BN(50),
        transferable: new BN(50),
      },
      [bob.address]: {
        ...zeroBalance,
        total: new BN(20),
        locked: new BN(20),
      },
```

Or :point\_up: maybe even better? Dunno, :thinking:

packages/ui/test/accounts/hooks/useHasRequiredStake.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+    const { result } = renderUseTotalBalances(1000)
+    expect(result.current).toStrictEqual({
+      hasRequiredStake: false,
+      transferableAccounts: null,
+      accountsWithLockedFounds: {
+        [bob.address]: [aliceStash],
+        [aliceStash.address]: [bob],
+        [alice.address]: [aliceStash, bob],
+      },
+    })
```
jodator, line 198:
```typescript
    balances = {
      [alice.address]: {
      	...zeroBalance,
      	total: new BN(50),
      	transferable: new BN(50),
    	},
      [aliceStash.address]: {
        ..zeroBalance,
    	  total: new BN(250),
        locked: new BN(200),
        transferable: new BN(50),
      },
      [bob.address]: {
        total: new BN(700),
        locked: new BN(400),
        transferable: new BN(300),
      },
    }
    const { result } = renderUseTotalBalances(1000)
    expect(result.current).toStrictEqual({
      hasRequiredStake: false,
      transferableAccounts: null,
      accountsWithLockedFounds: {
        [bob.address]: [aliceStash],        // 700 + (50) < 1000
        [aliceStash.address]: [bob],        // 250 + (300) < 1000
        [alice.address]: [aliceStash, bob], // 50 + (50 + 300) < 1000
      },
    })
```

The numbers in the test setup are correct :+1:. However, the transfer logic is different -> you should only take the `transferrable` amount to check if, when transferred to another account, will give an expected total (`my total + other transferable`).

I think that I might provide you some input forgetting to explain that the "locked" amount can't be transferred.

Also, I've found [a bug](https://github.com/Joystream/pioneer/issues/720) in our calculation of total / transferable amount :(

jodator (changes requested):
Could you double-check my calculations? I was looking at them for a while.

packages/ui/test/accounts/hooks/useHasRequiredStake.test.tsx
wyrdrender: Wrong line, I'm guessing you meant `useBalance`

jodator (approved):

PR #695, opened 2021-05-28
⏱ Fix tooltip show/hide timeout

### LINKED ISSUE #664: Check tooltip delay times.
There's a bit of lag (intended) in the time user moves the mouse and when the tooltip disappears.

When there's only one tooltip it is OK. However, when there are multiple elements with many tooltips it tends to create multiple opened tooltips:

ClumsyVlad:
Closes #664


jodator (approved):
Yeah, now it works smoothly :)

PR #703, opened 2021-05-28
🧃 Add the Proposals filters component

thesan:
See #585

### DIFF
--- a/packages/ui/src/common/utils.ts
+++ b/packages/ui/src/common/utils.ts
@@ -4,9 +4,11 @@ export const isDefined = <T extends any>(something: T | undefined): something is
 
 export const isNumber = (something: unknown): something is number => typeof something === 'number'
 
-export const objectShallowContains = <T extends { [k: string]: any }>(contained: T): ((container: T) => boolean) => {
-  const containedKeys: Array<keyof T> = Object.keys(contained)
-  return (container) => containedKeys.every((key) => container[key] === contained[key])
+export const objectEquals = <T extends Record<string, any>>(reference: T): ((compared: T) => boolean) => {
+  const expectedKeys: Array<keyof T> = Object.keys(reference)
+  return (compared) =>
+    expectedKeys.length === Object.keys(compared).length &&
+    expectedKeys.every((key) => compared[key] === reference[key])
 }
 
 type Indexer<T> = (item: T, index: number) => string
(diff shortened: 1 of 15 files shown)


packages/ui/src/proposals/components/ProposalFilters/ProposalFilters.tsx
```diff
@@ -0,0 +1,143 @@
...
+        />
+
+        <SimpleSelect
+          title="Proposer"
+          options={proposersOptions}
+          value={proposer}
+          onChange={(value) => {
+            dispatch({ type: 'change', field: 'proposer', value })
+          }}
+        />
```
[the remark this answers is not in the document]
jodator: > Also would it be fine to reuse the `<MockApolloProvider />` from the testing code, to render `<SelectMember />` in storybook ?

I think that we could use that 👍

packages/ui/src/common/utils.ts
```diff
@@ -3,3 +3,12 @@ export const isFunction = (something: unknown): something is CallableFunction =>
 export const isDefined = <T extends any>(something: T | undefined): something is T => typeof something !== 'undefined'
 
 export const isNumber = (something: unknown): something is number => typeof something === 'number'
+
+export const objectShallowContains = <T extends { [k: string]: any }>(contained: T): ((container: T) => boolean) => {
```
jodator, line 7: I think that the name might be misleading. AFAICS it compares an object to be shallowly equal to others, so maybe just `objectEquals()`? 

Below some variables names hint, but you can use any other you see fit.

```suggestion
export const objectEquals = <T extends Record<string, any>}>(reference: T): ((compared: T) => boolean) => {
```

jodator (changes requested):
Hi, I think that the clear filters change might enforce other changes in the code so I'll stop here. But, apart other things that I've commented the PR looks good.

packages/ui/src/common/utils.ts
thesan: Hi @jodator, thanks for the review.

Actually it was only doing something similar to Jest  [`expect.objectContaining`](https://jestjs.io/docs/expect#expectobjectcontainingobject). But I guess it makes more sense to check the keys length of the `compared` object to have something easier and safer to reuse somewhere else. So I changed the logic a bit.

jodator (approved):

PR #704, opened 2021-05-30
🏗️ Display "Reward per x blocks"

### LINKED ISSUE #678: Display reward per x blocks fields
The reward per block is calculated from the joystream blockchain constant.

*   [ ] Check the joystream code for the value (might be 3600 blocks :shrug: )
*   [ ] Create configuration/const value (link to joystrem code would be awesome)
*   [ ] Change the occurrences in the dashboard and lists:
    *   [ ] Opening list (Working Groups -> openings)
    *   [x] My applications
    *   [x] My roles

p-sad:
Closes #678 

A couple of considerations:
* Right now group name is cast from `string` to `GroupName` type inside the `getReward` function; I think it could be beneficial to use `GroupName` in every instance of group name in objects from the Working Groups domain. I added a simple test to illustrate the case - also in the end the names will be known and finite. 
* Values in `GroupRewardPeriods` are taken from Joystream node code cited in the issue. I'm not 100% sure if they're correct (maybe I'm not imagining the scale of numbers right though)

### DIFF
--- a/packages/ui/src/working-groups/model/getReward.ts
+++ b/packages/ui/src/working-groups/model/getReward.ts
@@ -1,11 +1,16 @@
 import BN from 'bn.js'
 
-import { GroupName, GroupRewardPeriods } from '../types'
+import { GroupName, GroupRewardPeriods, KnownWorkingGroups } from '../types'
 import { Reward } from '../types/Reward'
 
 export function getReward(rewardPerBlock: number, groupName: string): Reward {
-  return {
-    payout: GroupRewardPeriods[groupName as GroupName].mul(new BN(rewardPerBlock)),
-    blockInterval: GroupRewardPeriods[groupName as GroupName].toNumber(),
-  }
+  return KnownWorkingGroups.includes(groupName as GroupName)
+    ? {
+        payout: GroupRewardPeriods[groupName as GroupName].mul(new BN(rewardPerBlock)),
+        blockInterval: GroupRewardPeriods[groupName as GroupName].toNumber(),
+      }
+    : {
+        payout: new BN(0),
+        blockInterval: 14400,
+      }
 }
--- a/packages/ui/src/working-groups/model/getReward.ts
+++ b/packages/ui/src/working-groups/model/getReward.ts
@@ -1,0 +1,10 @@
+import BN from 'bn.js'
+
+import { GroupName, GroupRewardPeriods } from '../types'
+
+export function getReward(rewardPerBlock: number, groupName: string) {
+  return {
+    value: GroupRewardPeriods[groupName as GroupName].mul(new BN(rewardPerBlock)),
+    interval: GroupRewardPeriods[groupName as GroupName].toNumber(),
+  }
+}
--- a/packages/ui/src/working-groups/model/getReward.ts
+++ b/packages/ui/src/working-groups/model/getReward.ts
@@ -1,16 +1,18 @@
 import BN from 'bn.js'
 
-import { GroupName, GroupRewardPeriods, KnownWorkingGroups } from '../types'
+import { GroupRewardPeriods, isKnownGroupName } from '../types'
 import { Reward } from '../types/Reward'
 
 export function getReward(rewardPerBlock: number, groupName: string): Reward {
-  return KnownWorkingGroups.includes(groupName as GroupName)
-    ? {
-        payout: GroupRewardPeriods[groupName as GroupName].mul(new BN(rewardPerBlock)),
-        blockInterval: GroupRewardPeriods[groupName as GroupName].toNumber(),
-      }
-    : {
-        payout: new BN(7),
-        blockInterval: 1,
-      }
+  if (isKnownGroupName(groupName)) {
+    return {
+      payout: GroupRewardPeriods[groupName].mul(new BN(rewardPerBlock)),
+      blockInterval: GroupRewardPeriods[groupName].toNumber(),
+    }
+  }
+
+  return {
+    payout: new BN(rewardPerBlock),
+    blockInterval: 1,
+  }
 }
--- a/packages/ui/src/working-groups/model/getReward.ts
+++ b/packages/ui/src/working-groups/model/getReward.ts
@@ -10,7 +10,7 @@ export function getReward(rewardPerBlock: number, groupName: string): Reward {
         blockInterval: GroupRewardPeriods[groupName as GroupName].toNumber(),
       }
     : {
-        payout: new BN(0),
-        blockInterval: 14400,
+        payout: new BN(7),
+        blockInterval: 1,
       }
 }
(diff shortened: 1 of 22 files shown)


jodator:
> *   Values in `GroupRewardPeriods` are taken from Joystream node code cited in the issue. I'm not 100% sure if they're correct (maybe I'm not imagining the scale of numbers right though)

They might be correct ([Pioneer 1 tokenomics](https://testnet.joystream.org/#/tokenomics)):

Curators lead earns 1,780,940 per week -> 254,420 per day -> 10,600.83 (?) per hour -> 176.68 per minute -> 2.94 per second.

This means that if block time is ~6s then values should be:

*   18 per block
*   259,560 per 14,420 blocks

Also, 14400 blocks are 24 hr period for 6s block time

packages/ui/src/working-groups/types/Reward.ts
```diff
@@ -0,0 +1,6 @@
+import BN from 'bn.js'
+
+export interface Reward {
+  value: BN
+  interval: number
```
jodator, line 5: We might forget what was what ;)

```suggestion
  payout: BN
  blockInterval: number
```

jodator: Payout - because this would be a payout done to the worker (this is also used in schema / node code).

packages/ui/test/working-groups/model/getRewardAmount.test.ts
```diff
@@ -0,0 +1,13 @@
+import { getReward } from '@/working-groups/model/getReward'
+
+describe('getRewardAmount', () => {
+  it('Example use', () => {
+    const reward = getReward(1, 'forum')
+    expect(reward.value.toNumber()).toEqual(14410)
```
jodator, line 6:
```suggestion
    const reward = getReward(2, 'forum')
    expect(reward.value.toNumber()).toEqual(28820)
```

Just to have it clear :)

packages/ui/test/working-groups/model/getRewardAmount.test.ts
```diff
@@ -0,0 +1,13 @@
...
+    expect(reward.interval).toEqual(14410)
+  })
+
+  it('Not a group name', () => {
+    expect(() => getReward(1, 'incorrect')).toThrow(TypeError)
+  })
```
jodator, line 12: I wonder if throwing will be a good idea here. Maybe a better would be `UnknownGroupReward` :thinking:

packages/ui/test/working-groups/model/getRewardAmount.test.ts
```diff
@@ -0,0 +1,13 @@
+import { getReward } from '@/working-groups/model/getReward'
+
+describe('getRewardAmount', () => {
```
jodator, line 3:
```suggestion
describe('getReward', () => {
```

packages/ui/test/working-groups/model/getRewardAmount.test.ts
jodator: After the second thought, we might need to have an unknown reward because it will not break Pioneer when a new group is added (but not configured). In such a scenario, we can just go with a value per 1 block instead of throwing.

packages/ui/src/working-groups/types/WorkingGroup.ts
```diff
@@ -40,3 +40,12 @@ export const asWorkingGroup = (group: WorkingGroupFieldsFragment): WorkingGroup
     budget: new BN(group.budget),
   }
 }
+
+export type GroupName = 'forum' | 'storage' | 'content' | 'membership'
```
jodator, line 44: We might introduce configuration for this so it might come from a JSON in the future :thinking: 

Right now I think that we can have 

```typescript
const KnowWorkingGroups = ['Membership', 'Storage', 'Other'] as const

type GroupName = typeof KnowWorkingGroups[number]

const isGroupname = ( name: string ): name is GroupName => {
  return KnowWorkingGroups.includes(name as any)
}
```

jodator (changes requested):
I'd go with not throwing here as when a new working group might be introduced the App should crash.

p-sad:
@jodator I went with your idea, now it returns a zero payout for an unknown group, to make it immediately noticeable that something's amiss

packages/ui/test/working-groups/model/getRewardAmount.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    const reward = getReward(2, 'forum')
+    expect(reward.payout.toNumber()).toEqual(28820)
+    expect(reward.blockInterval).toEqual(14410)
+  })
+
+  it('Not a known group name', () => {
+    expect(getReward(1, 'incorrect')).toEqual({
+      payout: new BN(0),
+      blockInterval: 14400,
+    })
```
jodator, line 16: The interval is not known, so the fallback should return:

```suggestion
    expect(getReward(7, 'incorrect')).toEqual({
      payout: new BN(7),
      blockInterval: 1,
    })
```

as we know, that we pay 7 per block (still not correct, but IMO better than `0` per something.

jodator (changes requested):
Quick R- ;)

- I'd return something known,
- the PR has conflicts

packages/ui/src/working-groups/model/getReward.ts
```diff
@@ -0,0 +1,16 @@
+import BN from 'bn.js'
+
+import { GroupName, GroupRewardPeriods, KnownWorkingGroups } from '../types'
+import { Reward } from '../types/Reward'
+
+export function getReward(rewardPerBlock: number, groupName: string): Reward {
+  return KnownWorkingGroups.includes(groupName as GroupName)
+    ? {
+        payout: GroupRewardPeriods[groupName as GroupName].mul(new BN(rewardPerBlock)),
+        blockInterval: GroupRewardPeriods[groupName as GroupName].toNumber(),
```
jodator: Using type guard function allows you to embrace the type system:

```typescript
export const isKnownGroupName = (name: string): name is GroupName => {
  return KnownWorkingGroups.includes(name as GroupName)
}
```

As:

```typescript
export function getReward(rewardPerBlock: number, groupName: string): Reward {
  if (isKnownGroupName(groupName)) {
    return {
      payout: GroupRewardPeriods[groupName].mul(new BN(rewardPerBlock)),
      blockInterval: GroupRewardPeriods[groupName].toNumber(),
    }
  }

  return {
    payout: new BN(rewardPerBlock),
    blockInterval: 1,
  }
}
```

packages/ui/src/working-groups/model/getReward.ts
```diff
@@ -0,0 +1,16 @@
...
+      }
+    : {
+        payout: new BN(7),
```
jodator:
```suggestion
        payout: new BN(rewardPerBlock),
```

Typo!

jodator (commented):
Thanks for the updates. Unfortunately, I've found one hard coded value there ;)

I've fixed that and used a type guard to improve type safety there

PR #721, opened 2021-05-31
🤑 Display User Earnings

### LINKED ISSUE #404: Display user earnings
A follow-up to the #287:

*   fetch & calculate past user earnings
*   display them in the worker & openings dashboard

DevNikita16:
Closes #404


packages/ui/src/working-groups/hooks/useMyEarnings.ts
```diff
@@ -0,0 +1,40 @@
...
+    createdAt_gte: startOfMonth(Date.now()).toISOString(),
+  }
+  const { loading, data } = useGetRewardsQuery({ variables: { where } })
+
+  if (loading) {
+    return {
+      last24hours: null,
+      month: null,
+    }
+  }
```
p-sad, line 28: Other query hooks return the query's `loading` as `isLoading`; I think it would be good to return it here as well, both to stay consistent and to explicitly state what's going on in the component (`{isLoading  ? <Loading /> : <...>}`)

p-sad (commented):
One suggestion

packages/ui/src/working-groups/hooks/useMyEarnings.ts
DevNikita16: For value in stats we have "-" when null or undefined passed as value, that's why I returned null.

p-sad: My problem with `null` is that it doesn't communicate much in terms of intent; it can mean that value is still loading or that it's missing altogether and one cannot really tell at a glance which one it is. I don't think it's a dealbreaker here but it does affect readability somewhat (hence the comment instead of changes request).

packages/ui/test/working-groups/components/MyEarningsStat.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { MyEarningsStat } from '@/working-groups/components/MyEarningsStat'
+import { UseMyEarnings } from '@/working-groups/hooks/useMyEarnings'
+
+let mockEarnings: UseMyEarnings
+
+jest.mock('../../../src/working-groups/hooks/useMyEarnings', () => {
+  return {
+    useMyEarnings: () => mockEarnings,
+  }
+})
```
jodator, line 16: Mocking the hook gives us little value because it will test only the part of reading the hook and displaying returned values.

I would use a mock server here as in other tests.

packages/ui/src/working-groups/hooks/useMyEarnings.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface UseMyEarnings {
+  last24hours: BN | null
+  month: BN | null
+}
+
+export function useMyEarnings(): UseMyEarnings {
+  const { allAccounts } = useAccounts()
+
+  const where: RewardPaidEventWhereInput = {
+    rewardAccount_in: allAccounts.map((account) => account.address),
```
jodator, line 18: A reward account can be changed during the last 30 day period. We need to check all payouts related to the user.

packages/ui/src/working-groups/hooks/useMyEarnings.ts
```diff
@@ -0,0 +1,40 @@
...
+    createdAt_gte: startOfMonth(Date.now()).toISOString(),
```
jodator, line 19: This will return [Jun 1st](https://date-fns.org/v2.22.1/docs/startOfMonth) for today ;)

jodator, line 19: This will return [Jun 1st](https://date-fns.org/v2.22.1/docs/startOfMonth) for today ;)

packages/ui/src/working-groups/hooks/useMyEarnings.ts
jodator: I'm torn here. For other things having `isLoading` is need as we return collections in most cases. This allows checking for a loading state that will differ from an empty collection.

OTOH this returns either number or null so it is a bit consistent with `useBalance()` hooks IIRC.

I think that returning `null` in this case may be OK for now.

jodator (changes requested):
The tests should mock the API response, not the inner hook return value.

This allows covering a broader part of the app as an integration test.

packages/ui/src/working-groups/components/MyEarningsStat.tsx
```diff
@@ -2,14 +2,17 @@ import BN from 'bn.js'
...
+  const earnings = useMyEarnings()
+
   return (
     <MultiTokenValueStat
       title="Earned in past"
       values={[
-        { label: '24 hours', value: new BN(200) },
-        { label: 'Month', value: new BN(102_000) },
+        { label: '24 hours', value: earnings.last24hours as BN },
+        { label: 'Month', value: earnings.month as BN },
```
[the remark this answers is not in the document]
DevNikita16: Because of `null` in hook when loading

packages/ui/src/mocks/resolvers/baseResolvers.ts
```diff
@@ -31,6 +31,14 @@ const getFilter = (where: Record<string, any>) => {
     if (type === 'in') {
       filters.push((model: Record<string, any>) => checkValue.includes(model[field]))
     }
+
+    if (type === 'gte') {
+      if (field === 'createdAt') {
+        filters.push((model: Record<string, any>) => new Date(model[field]) >= new Date(checkValue))
```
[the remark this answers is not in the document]
DevNikita16: Yes, seems like`.getTime()` is better here

packages/ui/src/working-groups/hooks/useMyEarnings.ts
DevNikita16: Ooops, forgot about this one

packages/ui/src/working-groups/hooks/useMyEarnings.ts
```diff
@@ -0,0 +1,40 @@
...
+      last24hours: null,
+      month: null,
+    }
+  }
+
+  const monthEarnings = data ? data.rewardPaidEvents.reduce((a, b) => a + Number(b.amount), 0) : 0
+
+  const last24hoursEvents = data
+    ? data.rewardPaidEvents.filter(
+        (event: RewardPaidEventFieldsFragment) => new Date(event.createdAt) >= subHours(Date.now(), 24)
```
jodator: Let's use `.getTime()` everywhere to compare dates.

packages/ui/src/working-groups/components/MyEarningsStat.tsx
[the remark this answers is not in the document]
jodator: I understand now. Using `as` disables type checking so indicates that you'd need to change the component's interface. See my commit.

packages/ui/test/working-groups/components/MyEarningsStat.test.tsx
```diff
@@ -0,0 +1,110 @@
+import { cryptoWaitReady } from '@polkadot/util-crypto'
+import { screen, render, waitForElementToBeRemoved } from '@testing-library/react'
+import { startOfToday, subDays } from 'date-fns'
+import React from 'react'
+import { BrowserRouter } from 'react-router-dom'
```
jodator: Dunno why I was using BrowserRouter in tests as there's MemoryRouter as well. I'll fix that

jodator (commented):
Thanks for the changes :)

I've made some adjustments though:
- removed `as BN` cast
- used `MemoryRouters` in all tests (this should be a separate PR, so sorry for that)
- refactored `useMyEarnings()`

For the refactor – I've made small adjustments because I saw minor duplication for calculating earings - this might be as well one function.

The other thing I did was creating a HOF for the filter which created compare date once and returned a filter function that uses that. Previously it was creating a new date + calculations for every event checked. Also I've used .getTime() there as well.

jodator (approved):

PR #723, opened 2021-06-01
🧮 Update balances calculation

### LINKED ISSUE #720: The transferable balance is wrong
The current useBalance calculations are wrong.

It should be `total = transferable + locked` and right now transferable is the same as locked.

jodator:
Closes #720

### DIFF
--- a/packages/ui/src/accounts/model/toBalances.ts
+++ b/packages/ui/src/accounts/model/toBalances.ts
@@ -1,0 +1,32 @@
+import { DeriveBalancesAll } from '@polkadot/api-derive/types'
+import { LockIdentifier } from '@polkadot/types/interfaces'
+import BN from 'bn.js'
+
+import { lockTypes } from '@/accounts/model/lockTypes'
+import { Balances } from '@/accounts/types'
+import { capitalizeFirstLetter } from '@/common/helpers'
+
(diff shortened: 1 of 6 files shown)


packages/ui/src/accounts/model/toBalances.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { lockedBalance, availableBalance, lockedBreakdown, freeBalance } = balances
+
+  const locks = lockedBreakdown.map((lock) => ({
+    amount: lock.amount,
+    type: lockLookup(lock.id),
+  }))
+
+  const transferable = availableBalance.toBn()
+  const locked = lockedBalance.toBn()
+  const total = freeBalance.toBn() // Should be .add(reservedBalance) ?
```
DevNikita16, line 23: I think `const total = freeBalance.add(lockedBalance)` according to issue description

jodator: I've just found this comprehensive answer: [https://stackoverflow.com/a/56418101/2273046.](https://stackoverflow.com/a/56418101/2273046.) 

I'm still not 100% if Joystream will allow reserving a balance.

So,

*   total = free + reserved
*   transferable = free - locked (looks like duplicates `available`)

PR #724, opened 2021-06-01
☄️ Activities icons replacement

pawelEth:


packages/ui/src/common/components/icons/activities/index.ts
```diff
@@ -0,0 +1,25 @@
...
+export * from './ClosedIcon'
+export * from './CreatedIcon'
+export * from './DecreasedIcon'
+export * from './ExiteRoleIcon'
+export * from './HiredIcon'
+export * from './IncreasedIcon'
+export * from './JoystreamIcon'
+export * from './OpeningIcon'
+export * from './OpeningStarIcon'
+export * from './OtherIcon'
```
jodator, line 13: Let's remove `OtherIcon` as it is the same as Joystream icon

packages/ui/src/common/components/icons/activities/index.ts
```diff
@@ -0,0 +1,25 @@
+export * from './AplicationIcon'
+export * from './AppliedIcon'
```
jodator, line 2: This uses the "linear" variant while other are "solid"

packages/ui/src/common/components/icons/activities/index.ts
```diff
@@ -0,0 +1,25 @@
...
+export * from './ProposalIcon'
+export * from './ProposalUpdatedIcon'
+export * from './RewardIcon'
+export * from './SlashedIcon'
+export * from './StakeReducedIcon'
+export * from './StatusUpdatedIcon'
+export * from './TerminatedIcon'
+export * from './UpcomingIcon'
+export * from './VoteIcon'
+export * from './WarnedIcon'
```
jodator, line 25: This uses the "linear" variant while other are "solid"

packages/ui/src/common/components/icons/activities/index.ts
```diff
@@ -0,0 +1,25 @@
+export * from './AplicationIcon'
+export * from './AppliedIcon'
+export * from './BudgetIcon'
+export * from './ClosedIcon'
```
jodator, line 4: This one misses "x".

jodator (changes requested):
Nice :) Some icons require updates as it seems that something went wrong during the export.

PR #728, opened 2021-06-01
🧻 Add unstaking period to Leave Role modal

### LINKED ISSUE #678: Display reward per x blocks fields

p-sad:
Closes #678

### DIFF
--- a/packages/ui/src/working-groups/queries/workingGroups.graphql
+++ b/packages/ui/src/working-groups/queries/workingGroups.graphql
@@ -28,7 +28,10 @@ fragment WorkerFields on Worker {
   }
   application {
     id
-    openingId
+    opening {
+      id
+      unstakingPeriod
+    }
   }
 }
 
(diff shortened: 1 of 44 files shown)


packages/ui/src/working-groups/queries/workingGroups.graphql
```diff
@@ -28,7 +28,10 @@ fragment WorkerFields on Worker {
   }
   application {
     id
-    openingId
+    opening {
+      id
+      unstakingPeriod
+    }
```
jodator, line 34: Maybe fetching an opening in the modal would be better - we don't always need unstacking period.

jodator, line 34: Maybe fetching an opening in the modal would be better - we don't always need unstacking period.

jodator (changes requested):
As talked F2F use openings

PR #729, opened 2021-06-01
🕴️Update members filters

### LINKED ISSUE #690: Update members filters
The design for the members page has changed due to the Proposal page design (both were made similar).

Update the filter box using #667 changes:

*   The "members" title + total count
*   The layout of the components
*   Please don't work on selects (they changed a bit) as those are WIP (ie Role select)
*   The filter apply should be invoked automatically when filters have changed
*   Show "clear all Filters" only when filters are set

thesan:
Closes #690

### DIFF
--- a/packages/ui/src/common/components/selects/select.tsx
+++ b/packages/ui/src/common/components/selects/select.tsx
@@ -43,7 +43,6 @@ export const Select = <T extends any>({
   useEffect(() => {
     if (isDefined(selected)) {
       setSelectedOption(selected)
-      onChange(selected)
     }
   }, [selected])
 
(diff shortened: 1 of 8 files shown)


packages/ui/src/common/components/selects/select.tsx
```diff
@@ -43,7 +43,6 @@ export const Select = <T extends any>({
   useEffect(() => {
     if (isDefined(selected)) {
       setSelectedOption(selected)
-      onChange(selected)
```
thesan, line 46: I removed this so the `onChange` handler applies the new filter only once per change and not on the first render. I checked with a few select through the app, and it doesn't appear to impact their behavior.

jodator:
I must ask @wyrdrender again for help with the reviews :)

wyrdrender:
> I must ask @wyrdrender again for help with the reviews :)

Sure thing 👍

wyrdrender (approved):

PR #730, opened 2021-06-01
🐽 Display information that no openings found

### LINKED ISSUE #707: Display information that there are no openings
In the Working Groups -> Openings:

*   display information when no openings were fetched similar to other "empty" states.

Follow-up for: #665

DevNikita16:
Closes #707

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpenings.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpenings.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react'
+import React, { JSXElementConstructor, useRef, useState } from 'react'
 
 import { WorkingGroupOpening } from '@/common/api/queries'
 import { ActivitiesBlock } from '@/common/components/Activities/ActivitiesBlock'
@@ -18,6 +18,7 @@ import { OpeningsList, UpcomingOpeningsList } from '@/working-groups/components/
 import { RolesList } from '@/working-groups/components/Roles/RolesList'
 import { useOpenings } from '@/working-groups/hooks/useOpenings'
 import { useUpcomingOpenings } from '@/working-groups/hooks/useUpcomingOpenings'
+import { BaseOpening } from '@/working-groups/types'
 
 import { AppPage } from '../../components/AppPage'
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpenings.tsx
```diff
@@ -43,7 +48,23 @@ export const WorkingGroupsOpenings = () => {
...
+    if (!(type === 'CURRENT' ? currentOpenings : upcomingOpenings).length) {
+      return <TextBig>No {type.toLowerCase()} openings found</TextBig>
+    }
+
+    return type === 'CURRENT' ? (
+      <OpeningsList openings={currentOpenings} />
+    ) : (
+      <UpcomingOpeningsList openings={upcomingOpenings} />
+    )
+  }
```
jodator, line 66: I'm not sure if this is super easy to read. Maybe something like  

```tsx
  const displayOpenings = <T extends BaseOpening>(
    loading: boolean,
    openings: T[],
    Component: JSXElementConstructor<{ openings: T[] }>
  ) => {
    if (loading) {
      return <Loading />
    }

    if (openings.length) {
      return <TextBig>No openings found</TextBig>
    }

    return <Component openings={openings} />
  }
```

then:

```tsx
{activeTab === 'OPENINGS' && displayOpenings(upcomingLoading, upcomingOpenings, UpcomingOpeningsList)}
{activeTab === 'UPCOMING' && displayOpenings(currentLoading, currentOpenings, OpeningsList)}
```

Then you can also move this function out of the main render function scope, so fewer function creations as well.

jodator: And it looks like component :upside\_down\_face: 

`<OpeningsTab isLodaing={upcomingLoading} openings={upcomingOpenings} component={UpcomingOpenings} />`

jodator (commented):
I think that fewer ifs the better, WDYT?

packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpenings.tsx
DevNikita16: Agree, looks just awesome!

PR #732, opened 2021-06-01
🦖 Add filters to the past proposals page

### LINKED ISSUE #585: Add filters to the past proposals page

thesan:
Closes #585


packages/ui/src/app/pages/Proposals/PastProposals.tsx
```diff
@@ -1,34 +1,44 @@
...
 
   return (
     <AppPage>
-      <PageHeader>
-        <PageTitle>Proposals</PageTitle>
+      <FilterPageHeader ref={searchSlot} title="Proposals">
         <AddProposalButton />
         <ProposalsTabs />
-      </PageHeader>
+      </FilterPageHeader>
```
jodator, line 31: I've just noticed that the search box is bigger than the "add new proposal" button.

jodator (changes requested):
filtersI've found only some minor CSS issues:

- the search box is bigger than the button next to it
- I think that clear all filters text is not bold on the design

It might be a good point to make the filters box match the design more [closely](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2599%3A109614). If you have any doubts, please let me know as I might be missing something.

jodator (approved):

PR #736, opened 2021-06-02
💳 Implement move founds modal on apply for opening

wyrdrender:
* Add new modal in case of insufficient balance
* Refactor move founds modal to meet useHasRequiredStake value
* WorkingGroupsOpenings eslint errors

### DIFF
--- a/packages/ui/src/accounts/modals/MoveFoundsModal/MoveFoundsAccountItem.tsx
+++ b/packages/ui/src/accounts/modals/MoveFoundsModal/MoveFoundsAccountItem.tsx
@@ -2,7 +2,6 @@ import React, { memo } from 'react'
 import styled from 'styled-components'
 
 import { AccountInfo } from '@/accounts/components/AccountInfo'
-import { useBalance } from '@/accounts/hooks/useBalance'
 import { useTotalBalances } from '@/accounts/hooks/useTotalBalances'
 import { MemberRow } from '@/accounts/modals/MoveFoundsModal/styles'
 import { Account, AddressToBalanceMap } from '@/accounts/types'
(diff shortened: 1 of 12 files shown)


packages/ui/src/working-groups/components/OpeningsList/OpeningDetails.tsx
```diff
@@ -20,6 +23,36 @@ import { isOpeningOpen } from '@/working-groups/model/isOpeningOpen'
...
+          },
+        })
+      } else {
+        showModal<MoveFundsModalCall>({
+          modal: 'MoveFundsModal',
+          data: { type: MoveFoundsModalType.NO_FOUNDS, requiredStake },
+        })
+      }
+    }
+  }, [hasRequiredStake, transferableAccounts])
```
jodator, line 55: I would move that to the `ApplyForRoleModal` this way we don't have to copy that check when invoking some modal. I don't recall if we have only one place with this button but there's a chance that we have at least two.

jodator: The other thing is that the main modal, like ApplyForRole modal, could check only for `hasRequiredStake` and display the `<MoveFundsModal />`. The inner checks for which variant to display might be moved to the MoveFundsModall as well.

jodator (changes requested):
Looks good :) But I think that the check should be done inside the ApplyForRole modal.

packages/ui/src/working-groups/components/OpeningsList/OpeningDetails.tsx
wyrdrender: Sorry, there was a slight misunderstanding. I understood that you wanted to rid the modal of all logic. I'll fix it 👍

jodator (approved):

Issue #685, opened 2021-05-27
Display next payout in...


jodator:
To display next payout we need to:

*   get current block id from polkadot api
*   calculate when the reward block is happening (https://github.com/Joystream/pioneer/issues/678#issuecomment-849523092)

We should display them in My roles dashboards (see #665)

PR #738, opened 2021-06-02
🎺 Display blocks until next payout

p-sad:
See: #685 
Right now it says (blocks) JOY, I feel like setting up styles (currently hard-coded for JOY token values) is worthy of a separate PR

### DIFF
--- a/packages/ui/test/working-groups/model/getNextPayout.test.ts
+++ b/packages/ui/test/working-groups/model/getNextPayout.test.ts
@@ -1,0 +1,45 @@
+import BN from 'bn.js'
+
+import { getNextPayout } from '../../../src/working-groups/model/getNextPayout'
+
+describe('getNextPayout', () => {
+  it('Single role', () => {
+    const blockNumber = new BN(14400)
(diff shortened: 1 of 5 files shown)


packages/ui/test/working-groups/model/getNextPayout.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+  })
+})
+
+/*  Payout intervals in blocks:
+    forum: 14400 + 10
+    storage: 14400 + 20
+    content: 14400 + 30
+    membership: 14400 + 40
+*/
```
jodator, line 45:
```suggestion
})

```

jodator (changes requested):
PR has tests, so it is a good PR 👍 I've added some stylistic remarks.

> Right now it says (blocks) JOY, I feel like setting up styles (currently hard-coded for JOY token values) is worthy of a separate PR

Nah, let's have a proper component already - a new component that displays "text" instead of JOY. And pleas add it to the [storybook](http://localhost:6006/?path=/story/common-statistics--statistic) (local link).

PR #749, opened 2021-06-07
🤵 Implement Proposer Select

### LINKED ISSUE #722: Implement Proposer Select
> Currently the proposers select uses a static array of proposers, but it's not a sustainable long term solution. I wanted to use `<SelectMember />` but it's currently not a controlled input so it wouldn't get reset when clearing the filters, and it would need a several style changes to fit this use. Should I start a PR to modify this component or create a new one, or we keep it as it is for now ?

Let's keep it for now, but please create an issue for this. i think that we need:

- allow typing a handle in the field (same as in member select)
- figure out if all members or only proposer to display here
- single or multi-select (single select is OK now)

_Originally posted by @jodator in https://github.com/Joystream/pioneer/pull/703#discussion_r642245097_

Follow-up for #585

thesan:
Closes #722

### DIFF
--- a/packages/ui/src/app/pages/Members/Members.tsx
+++ b/packages/ui/src/app/pages/Members/Members.tsx
@@ -6,8 +6,8 @@ import { PageTitle } from '@/common/components/page/PageTitle'
 import { Pagination } from '@/common/components/Pagination'
 import { MemberList } from '@/memberships/components/MemberList'
 import { MemberListEmptyFilter, MemberListFilters } from '@/memberships/components/MemberListFilters'
-import { memberRoleAbbreviation, memberRoleTitle } from '@/memberships/helpers'
 import { DefaultMemberListOrder, MemberListOrder, MemberListSortKey, useMembers } from '@/memberships/hooks/useMembers'
+import { MemberRole } from '@/memberships/types'
 
 import { AppPage } from '../../components/AppPage'
 
@@ -27,10 +27,10 @@ export const Members = () => {
 
   const { members, isLoading, totalCount, pageCount } = useMembers({ order, filter, page })
 
-  const roles: { [key: string]: string } = {}
+  const roles = new Set<MemberRole>()
   for (const member of members) {
     for (const role of member.roles) {
-      roles[memberRoleTitle(role)] = memberRoleAbbreviation(role)
+      roles.add(role)
     }
   }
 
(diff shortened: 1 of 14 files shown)


packages/ui/src/app/pages/Members/Members.tsx
```diff
@@ -27,10 +27,10 @@ export const Members = () => {
...
   const { members, isLoading, totalCount, pageCount } = useMembers({ order, filter, page })
 
-  const roles: { [key: string]: string } = {}
+  const roles = new Set<MemberRole>()
   for (const member of members) {
     for (const role of member.roles) {
-      roles[memberRoleTitle(role)] = memberRoleAbbreviation(role)
+      roles.add(role)
     }
   }
```
jodator, line 35: I think that we need some sort of map/configuration of possible roles. For now, we have Group + Worker/Leader. So you can assume that the list is static.

However, it might be too soon to implement a query for it, so don't worry about that. We're waiting for the schema upgrade anyway so we can add that later.

packages/ui/src/proposals/components/ProposalFilters/SelectProposer.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  },
+} as Meta
+
+export const Default: Story = ({ onChange }) => {
+  const [value, setValue] = useState<Member | null>(null)
+
+  return (
+    <MockApolloProvider>
+      <MembershipContextProvider>
+        <SelectMember onChange={onChange} /> {/* TODO remove */}
```
jodator, line 25: TODO found ;) What we should do with that?

ps.: AFAIR the `<SelectMember />` looks OK when wrapped in the Input component (like in the forms).

jodator (changes requested):
Looks good 👍 

See my comments for details we should have a static list of roles. How they will be implemented is another story as we're preparing for the schema upgrade.

packages/ui/src/proposals/components/ProposalFilters/SelectProposer.stories.tsx
thesan: Sorry about that :facepalm:
I actually meant to remove it from the story before committing (I only added it to make sure I wasn't breaking `<SelectMember />` while editing the select component).

I will create an issue to add  `<SelectMember />` in a separate story though.

packages/ui/src/app/pages/Members/Members.tsx
jodator: Let's have this done in https://github.com/Joystream/pioneer/issues/757#issuecomment-856535966.

PR #753, opened 2021-06-07
🧅 Add Next Payout component

### LINKED ISSUE #685: Display next payout in...

p-sad:
Closes #685

### DIFF
--- a/packages/ui/src/common/model/blocksToTime.ts
+++ b/packages/ui/src/common/model/blocksToTime.ts
@@ -1,16 +1,18 @@
 import BN from 'bn.js'
 
+import { SECONDS_PER_BLOCK } from '../constants'
+
 import { significantTimeUnit } from './significantTimeUnit'
 
 export function blocksToTime(blocks: BN) {
-  const seconds = blocks.toNumber() * 7
+  const seconds = blocks.toNumber() * SECONDS_PER_BLOCK
   if (seconds < 0) {
     return 'never'
   }
   const { count, unit } = significantTimeUnit(seconds)
   if (unit) {
-    const format = Intl.NumberFormat('en', { style: 'unit', unit, unitDisplay: 'long' })
+    const format = Intl.NumberFormat('en', { style: 'unit', unit, unitDisplay: 'short' })
     return format.format(count)
   }
-  return 'less than a minute'
+  return '<1 min'
 }
--- a/packages/ui/src/common/model/blocksToTime.ts
+++ b/packages/ui/src/common/model/blocksToTime.ts
@@ -1,0 +1,16 @@
+import BN from 'bn.js'
+
+import { significantTimeUnit } from './significantTimeUnit'
+
+export function blocksToTime(blocks: BN) {
+  const seconds = blocks.toNumber() * 7
+  if (seconds < 0) {
+    return 'never'
+  }
+  const { count, unit } = significantTimeUnit(seconds)
+  if (unit) {
+    const format = Intl.NumberFormat('en', { style: 'unit', unit, unitDisplay: 'long' })
+    return format.format(count)
+  }
+  return 'less than a minute'
+}
--- a/packages/ui/test/common/model/blocksToTime.test.ts
+++ b/packages/ui/test/common/model/blocksToTime.test.ts
@@ -22,4 +22,11 @@ describe('blocksToTime', () => {
     const blocks = new BN(20)
     expect(blocksToTime(blocks)).toEqual('2 min')
   })
+
+  it('bigger numbers of blocks', () => {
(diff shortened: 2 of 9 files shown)


packages/ui/src/common/model/blocksToTime.ts
```diff
@@ -0,0 +1,16 @@
+import BN from 'bn.js'
+
+import { significantTimeUnit } from './significantTimeUnit'
+
+export function blocksToTime(blocks: BN) {
+  const seconds = blocks.toNumber() * 7
```
jodator, line 6: There's a magic number and I think that it might be wrong :) (shouldn't it be 6 seconds?).

Please use a const loaded from one location (maybe the same as `1441--` payout periods?

packages/ui/test/common/model/blocksToTime.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  it('a minute', () => {
+    const blocks = new BN(9)
+    expect(blocksToTime(blocks)).toEqual('1 minute')
+  })
+
+  it('two minutes', () => {
+    const blocks = new BN(18)
+    expect(blocksToTime(blocks)).toEqual('2 minutes')
+  })
```
jodator, line 24: Could you add tests for longer periods? Like 48 minutes, 2+hours, maximum period of payout, etc?

packages/ui/src/common/model/blocksToTime.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (seconds < 0) {
+    return 'never'
+  }
+  const { count, unit } = significantTimeUnit(seconds)
+  if (unit) {
+    const format = Intl.NumberFormat('en', { style: 'unit', unit, unitDisplay: 'long' })
+    return format.format(count)
+  }
+  return 'less than a minute'
```
jodator, line 15: I was thinking a bit about this. I think that we can have this for now, but maybe use 'short' as in the design?

So here it would be "\< 1 min" or we can just go to seconds :shrug: – it's a coin toss for me which will be better:

*   \< 1 min
*   12 sec
*   now (for 0 s)

But the `"short"` format for sure :)

jodator (changes requested):
I think that we could add more tests here and use short format for dates.

packages/ui/src/common/model/blocksToTime.ts
p-sad: Payout periods are in Working Groups domain, while I believe this constant is more general - moved it to common/constants for now

packages/ui/src/common/model/blocksToTime.ts
p-sad: I think "<1 min" is good enough precision, especially that not every block takes exactly the same time to mine

p-sad: Also I'd forego "now" for the exact same reason, especially that it'll roll over to ~14400b/1 day right next block

packages/ui/test/common/model/blocksToTime.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    expect(blocksToTime(blocks)).toEqual('2 min')
+  })
+
+  it('bigger numbers of blocks', () => {
+    expect(blocksToTime(new BN(481))).toEqual('48 min')
+    expect(blocksToTime(new BN(800))).toEqual('1 hr')
+    expect(blocksToTime(new BN(2800))).toEqual('4 hr')
+    expect(blocksToTime(new BN(14430))).toEqual('1 day')
+  })
+})
```
jodator, line 32:
```suggestion
describe('blocksToTime', () => {
  it('Less than one minute', () => {
    const blocks = new BN(0)
    expect(blocksToTime(blocks)).toEqual('<1 min')
  })

  it('Negative number of blocks', () => {
    const blocks = new BN(-1)
    expect(blocksToTime(blocks)).toEqual('never')
  })

  it('A minute', () => {
    const blocks = new BN(10)
    expect(blocksToTime(blocks)).toEqual('1 min')
  })

  it('Two minutes', () => {
    const blocks = new BN(20)
    expect(blocksToTime(blocks)).toEqual('2 min')
  })

  it('Bigger numbers of blocks', () => {
    expect(blocksToTime(new BN(481))).toEqual('48 min')
    expect(blocksToTime(new BN(800))).toEqual('1 hr')
    expect(blocksToTime(new BN(2800))).toEqual('4 hr')
    expect(blocksToTime(new BN(14430))).toEqual('1 day')
  })
})
```

Issue #709, opened 2021-05-31
Working Groups list updates


jodator:
In the Working Groups -> Working Groups list:

*   Display only active workers (this might require schema upgrade to filter related fields or a separate inner-query for this value). See #708
*   Display the number of open openings
*   Annotate "new" opening (optional, might require clarification)

Follow-up for: #665

PR #762, opened 2021-06-07
🔢 Count working group workers and openings

DevNikita16:
See #709 , maybe closes

I was using already existing fiilters, but i tried to filter on step of querying GraphQL. And I have not found any good info how to query on nested fields

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/WorkingGroups.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/WorkingGroups.tsx
@@ -1,10 +1,11 @@
 import React from 'react'
 
-import { Loading } from '../../../common/components/Loading'
-import { PageHeader } from '../../../common/components/page/PageHeader'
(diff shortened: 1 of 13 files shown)


jodator:
I've asked about that on rocket chat. I think that querying for everything looks like a bit overkill as for every related field you'd need an additional sub-query in the underlining DB engine. For now, let's see what the options are there.

From my POV I see that we could have simple "count" queries on `workingGroupOpeningsConnection` type. Those would return `totalCount` so it might be better. The downside of this is that those will be very small components that would need to handle "loading" state inside a list item.

jodator:
@DevNikita16 Let's go with counting using `*Connection` query in every component. Getting all the fields of Opening is not needed when querying for the Working Groups list.

DevNikita16:
@jodator, you can have a look on my last commit. I introduced there 2 resolvers: onlyActive for workers, onlyOpened for openings. I was not able to manage status_json({typeOf_eq}) resolver, as soon as there should be some nested filtering

PR #766, opened 2021-06-08
🐴 Access Member Profile through a link

### LINKED ISSUE #613: Allow linking to member profile

p-sad:
Closes #613

### DIFF
--- a/packages/ui/src/memberships/components/MemberProfile/MemberProfile.tsx
+++ b/packages/ui/src/memberships/components/MemberProfile/MemberProfile.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components'
 
 import { ButtonGhost, CloseButton } from '@/common/components/buttons'
 import { LinkIcon } from '@/common/components/icons/LinkIcon'
+import { useCopyToClipboard } from '@/common/hooks/useCopyToClipboard'
 
 import { MemberInfoWrap } from '..'
 import { EditSymbol } from '../../../common/components/icons/symbols'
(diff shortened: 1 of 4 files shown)


packages/ui/src/memberships/components/MemberProfile/MemberProfile.tsx
```diff
@@ -67,7 +69,7 @@ export const MemberProfile = React.memo(() => {
                   <EditSymbol />
                 </EditMembershipButton>
               )}
-              <ButtonGhost size="small">
+              <ButtonGhost size="small" onClick={() => copyValue(`${window.location.host}/#/members/${member.id}`)}>
```
p-sad, line 72: I wanted to use either this or `window.location.origin`

jodator: I think that in this case it is OK. I don't recall what's best, but host AFAIR is good enough.

jodator: However, it looks like we should have a way to handle links globally. (ATM requires changing links in a couple of places). But that's for later.

jodator (approved):

packages/ui/src/memberships/components/MemberProfile/MemberProfile.tsx
jodator: Moved to: #768.

Issue #743, opened 2021-06-04
Connect to Olympia testnet


jodator:
### Requirements

*   [x] Upgrade schema & types #767
*   [x] Allow switching networks (local/testnet) #771
*   [x] Connect to the Olympia testnet #771
*   [x] Align code to the real-world data #784 

#### Tech details:

*   WS endpoint: wss:_//olympia-dev.joystream.app/rpc_
*   Query-node graphql server endpoint: [https:_//olympia-dev.joystream.app/query/server/graphql_](https://olympia-dev.joystream.app/query/server/graphql)
*   Indexer graphql endpoint: [https:_//olympia-dev.joystream.app/query/gateway/graphql_](https://olympia-dev.joystream.app/query/gateway/graphql)

PR #767, opened 2021-06-08
🏺Upgrade Olympia types

jodator:
See #743


packages/ui/src/common/api/schemas/schema.graphql
```diff
@@ -46,8 +38,8 @@ type ApplicationFormQuestionAnswerConnection {
 }
 
 input ApplicationFormQuestionAnswerCreateInput {
-  applicationId: ID!
-  questionId: ID!
+  application: ID!
```
DevNikita16, line 41: Is there any reason not to use `applicationId` or `questionId`? For me `...Id` represents ID, but without ID is object

packages/ui/src/mocks/data/seedWorkers.ts
```diff
@@ -37,11 +37,11 @@ export const seedWorker = (worker: RawWorker, server: any) => {
 const seedWorkerStatus = (status: WorkerStatus, id: string, server: any) => {
   switch (status) {
     case 'active':
-      return server.schema.create('WorkerStatusActive', {})
+      return server.schema.create('WorkerStatusActive', { phantom: 0 })
```
DevNikita16, line 39: I think I can use this in my new commit instead of my resolvers, I should just filter by phantom:0

DevNikita16 (commented):
Like this changes 👍 My only question why you've renamed `applicationId` to `application` etc.

packages/ui/src/common/api/schemas/schema.graphql
jodator: The schema is defined externally. Unfortunately, we have to live with that as we're the consumers here :)

packages/ui/src/mocks/data/seedWorkers.ts
jodator: I think that filtering by `__typename: 'WorkerStatusActive'` will be more future-proof. The `phantom` was used to overcome Hydra schema shortcomings AFAIR.

DevNikita16: https://dzhelezov.gitbook.io/hydra/docs/schema-spec/variant-types
I was not able to filter by __typename, but maybe will be able to intoduce isTypeOf after you changes for nested filtering

p-sad (approved):

PR #770, opened 2021-06-09
Count working group openings

DevNikita16:
See #709

### DIFF
--- a/packages/ui/src/working-groups/hooks/useWorkingGroup.ts
+++ b/packages/ui/src/working-groups/hooks/useWorkingGroup.ts
@@ -14,7 +14,7 @@ export function useCountOpenings(groupId: string) {
     variables: { groupId_eq: groupId, status_json: { isTypeOf_eq: 'OpeningStatusOpen' } },
   })
 
-  return loading ? 0 : data?.workingGroupOpeningsConnection.totalCount
+  return { isLoading: loading, openings: data?.workingGroupOpeningsConnection.totalCount }
 }
 
 export function useWorkingGroup(where: WhereInput) {
--- a/packages/ui/src/working-groups/hooks/useWorkingGroup.ts
+++ b/packages/ui/src/working-groups/hooks/useWorkingGroup.ts
@@ -1,8 +1,24 @@
-import { useGetWorkingGroupQuery } from '../queries'
+import { useGetWorkingGroupOpeningsConnectionQuery, useGetWorkingGroupQuery } from '../queries'
 import { asWorkingGroup } from '../types'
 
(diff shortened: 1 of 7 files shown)


packages/ui/src/working-groups/hooks/useWorkingGroup.ts
```diff
@@ -1,10 +1,26 @@
...
   name?: string | undefined
 }
 
+// export function useCountWorkers(groupId: string) {
+//   const { data, loading } = useGetWorkersConnectionQuery({
+//     variables: { groupId, onlyActive: true },
+//   })
+//
+//   return loading ? 0 : data?.workersConnection.totalCount
+// }
```
jodator, line 14: :police\_car:
```suggestion
```

jodator (changes requested):
I think that we need at least some simple test for this and this will be done.

packages/ui/src/working-groups/hooks/useWorkingGroup.ts
DevNikita16: Will the test for WorkingGroupListItem be OK?

packages/ui/src/working-groups/queries/workingGroups.graphql
```diff
@@ -157,8 +157,8 @@ fragment WorkingGroupOpeningFieldsConnection on WorkingGroupOpeningConnection {
   }
 }
 
-query getWorkingGroupOpeningsConnection($groupId_eq: ID, $first: Int, $last: Int) {
-  workingGroupOpeningsConnection(where: { group_eq: $groupId_eq }, first: $first, last: $last) {
+query getWorkingGroupOpeningsConnection($groupId_eq: ID, $status_json: JSONObject, $first: Int, $last: Int) {
+  workingGroupOpeningsConnection(where: { group_eq: $groupId_eq, status_json: $status_json }, first: $first, last: $last) {
     ... WorkingGroupOpeningFieldsConnection
   }
```
jodator, line 163: Oh, one last thing – I've forgot to check this. I think that it might be better to have a dedicated query:

```
query CountWorkingGroupOpenings($groupId_eq: ID, $status_json: JSONObject, $first: Int, $last: Int) {
  workingGroupOpeningsConnection(where: { group_eq: $groupId_eq, status_json: $status_json }, first: $first, last: $last) {
    id
  }
```

This way the GraphQL server will do a simple select to return the count. The `...WorkingGroupOpeningFields` fetches too much data and it might be slow (joins, etc).

jodator (approved):
👍

PR #771, opened 2021-06-09
⚙️ Network switch

jodator:
See #743:

*   added Settings page (simplest scenario)
*   added network switch in the Settings page

### DIFF
--- a/packages/ui/src/common/hooks/useNetwork.ts
+++ b/packages/ui/src/common/hooks/useNetwork.ts
@@ -1,0 +1,10 @@
+import { useLocalStorage } from './useLocalStorage'
+
+export type NetworkType = 'local' | 'olympia-testnet'
+
+export const useNetwork = () => {
+  const [network, setNetwork] = useLocalStorage<NetworkType>('network')
+  const resolvedNetwork = network || 'local'
+
+  return [resolvedNetwork, setNetwork] as const
+}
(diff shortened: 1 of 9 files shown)


jodator:
:roll\_eyes: I forgot that the MirageJS server needs to be created before apollo's one otherwise we overriding fetch do not work.

packages/ui/src/common/hooks/useNetwork.ts
```diff
@@ -0,0 +1,10 @@
+import { useLocalStorage } from './useLocalStorage'
+
+export type NetworkType = 'local' | 'olympia-testnet'
+
+export const useNetwork = () => {
+  const [network, setNetwork] = useLocalStorage<NetworkType>('network')
+  const resolvedNetwork = network || 'local'
+
+  return [resolvedNetwork, setNetwork] as const
```
DevNikita16, line 9: Why `as const` needed here?

DevNikita16 (approved):

packages/ui/src/common/hooks/useNetwork.ts
jodator: See this: [https://github.com/typescript-cheatsheets/react#custom-hooks](https://github.com/typescript-cheatsheets/react#custom-hooks)

Issue #769, opened 2021-06-09
Check queries performance


jodator:
I think that some..

p-sad:
### Before vs after:
getWorkingGroups query
![image](https://user-images.githubusercontent.com/21345040/121503894-c8e1d500-c9e1-11eb-8bd6-bc93c075a301.png)
![image](https://user-images.githubusercontent.com/21345040/121503966-dd25d200-c9e1-11eb-9198-2e52b2666d64.png)

getWorkingGroupsConnectionQuery
![image](https://user-images.githubusercontent.com/21345040/121503770-ab147000-c9e1-11eb-9596-612ad05d1d42.png)
![image](https://user-images.githubusercontent.com/21345040/121504123-03e40880-c9e2-11eb-886c-c7306f664b33.png)

getWorkersQuery
![image](https://user-images.githubusercontent.com/21345040/121504313-355cd400-c9e2-11eb-86e5-964c2df21129.png)
![image](https://user-images.githubusercontent.com/21345040/121504399-460d4a00-c9e2-11eb-8bc6-7679d3110462.png)

The workers query still takes a bit long; it might be worth it to include only some member fields per worker instead of all of them.

jodator:
Thanks! Nice summary :)

PR #772, opened 2021-06-09
🪗 Optimize queries

p-sad:
See #769 
Updated two kinds of queries so far, workers and working groups.
I think Openings have potential to be slimmed down as well, seeing that the multiple openings query gets an entire list of applicants. They're already a bit complicated though, being inherited from BaseOpening interface

### DIFF
--- a/packages/ui/src/memberships/hooks/useMemberRoles.ts
+++ b/packages/ui/src/memberships/hooks/useMemberRoles.ts
@@ -1,12 +1,12 @@
 import { useMemo } from 'react'
 
-import { useGetWorkersQuery } from '@/working-groups/queries'
+import { useGetDetailedWorkersQuery } from '@/working-groups/queries'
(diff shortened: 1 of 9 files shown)


jodator:
@p-sad Could you post a loading time comparison as before / after?

p-sad:
* Working Groups page: 1773ms before vs 126ms after
* Working Groups Workers sidebar: 1ms vs 1ms

So collections of objects are, predictably, a higher performance hit.
I'm removing the applicants list from the (non-unique) openings query, I'll see how that plays out.

jodator:
> Working Groups page: 1773ms before vs 126ms after

That's huge :sunglasses:

p-sad:
Openings without applicants shows improvement from 705ms to 334ms with current mocks.
This will obviously scale with the number of applicants, but I'd say it's still a marked improvement.

jodator (approved):
Let's have that merged 🚀

PR #773, opened 2021-06-09
⛑ Add link to application in workers/applicants list

### LINKED ISSUE #715: Handle worker button in workers list
In the workes list (ie in the Working Group summary) there's a button that does nothing.

*   [x] verify what the button should link to either:
    *   ~the worker's role page~
    *   application side modal
*   [x] implement the action

DevNikita16:
Closes #715 

Actually this task had a bit bigger scope. Our Worker template was used in applicants list, I handle this as well

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
@@ -127,8 +127,8 @@ export const WorkingGroupOpening = () => {
           <SidePanel neighbor={sideNeighborRef}>
             <ApplicantsList
               allApplicants={opening.applications}
-              myApplication={hiredMember?.member.id === activeMembership?.id ? activeMembership : undefined}
-              hired={hiredMember?.member}
+              myApplication={hiredMember && hiredMember?.member.id === activeMembership?.id ? hiredMember : undefined}
+              hired={hiredMember || undefined}
               hiringComplete={opening.status !== OpeningStatuses.OPEN}
               leaderId={opening.leaderId}
             />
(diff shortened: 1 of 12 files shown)


jodator (commented):
The change looks OK.

Unfortunately, we've introduced some changes in other PRs so sorry for the merge conflicts.

packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
```diff
@@ -127,8 +127,8 @@ export const WorkingGroupOpening = () => {
           <SidePanel neighbor={sideNeighborRef}>
             <ApplicantsList
               allApplicants={opening.applications}
-              myApplication={hiredMember?.member.id === activeMembership?.id ? activeMembership : undefined}
-              hired={hiredMember?.member}
+              myApplication={hiredMember && hiredMember?.member.id === activeMembership?.id ? hiredMember : undefined}
+              hired={hiredMember || undefined}
```
jodator: Maybe it's OK to allow `undefined | null` in the ApplicantsList component? WDYT? Having to deal with undefined | null is cumbersome and I don't see a big need for this.

We could change the hook above to just `return` so there's no need to write the check.

jodator: I had to double-check what `hiredMember` is. I think that we could also rename that to something that means "application" not "member".

jodator (approved):
The PR is OK.

I've noticed however that the logic behind my application is not right. Hence my update.

packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
jodator: NVM, the logic behind `myMembership` was odd -> it should get your application from all the applications not only from the "hired" one. This was not related to your PR though, so I've changed that.

Issue #692, opened 2021-05-27
Implement Roles select


jodator:
Implement the [Roles select](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2985%3A111662):

*   It should allow selecting multiple roles
*   The selection should be confirmed by Apply Filter button
*   The selection can be cleared by Clear filter button
*   The selected roles should be displayed [as role icons](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2993%3A2794)
    *   If there are more roles selected than we can display show (+NN)
    *   It looks like member roles component could be potentially re-used :thinking:

PR #774, opened 2021-06-09
👉 Implement a `<MultiSelect />` component

thesan:
See #692 (the PR for this issue started to get too long)


jodator:
@thesan Hey, could you check the errors? You can also see the latest changes on master as there were a few updates.

thesan:
@jodator yes I fixed it already, but I realized some of the changes affect `<SelectAccount />` so I'm fixing this now.

packages/ui/src/accounts/components/SelectAccount/SelectAccount.tsx
```diff
@@ -33,6 +33,7 @@ export const SelectAccount = React.memo(({ onChange, filter, selected }: Props)
   const filteredOptions = useMemo(() => filterByText(options, search), [search, options])
   const keyring = useKeyring()
 
+  useEffect(() => setSelectedOption(selected), [selected])
```
thesan, line 36: The changes broke the Account selection, but I don't understand how it use to work without this line. Am I missing something ?

jodator: Previously the idea was to allow passing a pre-defined account to the component. It is used in line 30 as a default selected option in `useState()`.

jodator: So there are two options
- passing account as `selected` -> the component is rendered with an already selected item
- passing `undefined` as `selected` -> the component is rendered as empty

thesan: I get this, but I'm confused because it seems that after the component has been mounted: `setSelectedOption()` was only called when `filteredOptions.length === 0`. So I don't understand how the value of `selectedOption` updated, unless the component got unmounted

jodator: The hook below (line 37) should be rewritten AFAICS, as it is hard to parse. _I think_ that it is used to set "unknown" account when user pastes an address to the search box which is not a known account.

PR #780, opened 2021-06-10
☯️ Use query filtering instead of frontend filtering for status

### LINKED ISSUE #472: Use a `where` clause for retrieving historical openings and workers
Filtering for current/historical items is currently done using `array.filter()` in the frontend; it should be done by constructing a more accurate GQL query instead.
Hooks:
`useOpenings`
`useWorkers`

DevNikita16:
Closes #472


packages/ui/src/working-groups/hooks/useWorkers.ts
```diff
@@ -1,38 +1,39 @@
...
-}
-
-const getWorkersFilter = (statusIn?: Status[]) => {
-  const statusTypeIn = statusIn?.map((status) => StatusTypename[status]) ?? ['WorkerStatusActive']
-  return (worker: WorkerFieldsFragment) => statusTypeIn.includes(worker.status.__typename)
-}
+// const getWorkersFilter = (statusIn?: WorkerStatus[]) => {
+//   const statusTypeIn = statusIn?.map((status) => StatusTypename[status]) ?? ['WorkerStatusActive']
+//   return (worker: WorkerFieldsFragment) => statusTypeIn.includes(worker.status.__typename)
+// }
```
jodator, line 39: :police\_officer:

packages/ui/src/working-groups/hooks/useWorkers.ts
```diff
@@ -1,38 +1,39 @@
...
 
 export const useWorkers = ({ groupId, statusIn }: UseWorkersProps) => {
-  const options = { variables: { where: { group_eq: groupId } } }
+  const options = {
+    variables: {
+      where: {
+        group_eq: groupId,
+        status_json: statusIn ? { isTypeOf_in: statusIn.map((status) => WorkerStatusTypename[status]) } : undefined,
+      },
+    },
```
jodator, line 20:
```suggestion
    variables: {
      where: {
        group_eq: groupId,
        status_json: getStatus(statusIn),
      },
    },
```

and something like this:

```typescript
const getSatus = (statusIn?: WorkerStatus[]) => {
  if (!statusIn) {
    return
  }
  
  return { isTypeOf_in: statusIn.map((status) => WorkerStatusTypename[status]) }
}
```

WDYT? Just to hide the implementation from the `variables` definition. There's so much happening on the `status_json` line and that's not necessary to have that there.

jodator: The same applies for the other hooks that uses `statusIn`

jodator (changes requested):
R- mostly because of the 🚓 -> commented out code left. Overall the PR is good.

I've added one optional comment about code readability (lower cognitive load when reading).

jodator (approved):

PR #799, opened 2021-06-10
🎰 Add pagination for WG past workers

### LINKED ISSUE #795: Add pagination for WG past workers

DevNikita16:
Closes #795 
Also fixed Eslint warning in `PageContent.tsx`

### DIFF
--- a/packages/ui/src/working-groups/hooks/useWorkersPagination.ts
+++ b/packages/ui/src/working-groups/hooks/useWorkersPagination.ts
@@ -1,0 +1,43 @@
+import { useEffect, useMemo, useState } from 'react'
+
+import { getStatusWhere, UseWorkersProps } from '@/working-groups/hooks/useWorkers'
+import { useGetWorkersConnectionQuery } from '@/working-groups/queries'
+import { asWorkerBaseInfo } from '@/working-groups/types'
+
+export const WORKERS_PER_PAGE = 5
+
+interface UseWorkersPaginationProps extends UseWorkersProps {
+  page?: number
+}
+
+export const useWorkersPagination = ({ groupId: group_eq, statusIn, page }: UseWorkersPaginationProps) => {
+  const variables = {
+    where: {
+      group_eq,
+      status_json: getStatusWhere(statusIn),
+    },
+    first: page !== undefined ? page * WORKERS_PER_PAGE : undefined,
+    last: page !== undefined && page > 1 ? WORKERS_PER_PAGE : undefined,
+  }
+
+  const { loading, data } = useGetWorkersConnectionQuery({ variables })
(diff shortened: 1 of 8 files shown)


packages/ui/src/working-groups/hooks/useWorkersPagination.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const useWorkersPagination = ({ groupId: group_eq, statusIn, page }: UseWorkersPaginationProps) => {
+  const variables = {
+    where: {
+      group_eq,
+      status_json: getStatusWhere(statusIn),
+    },
+    first: page !== undefined ? page * WORKERS_PER_PAGE : undefined,
+    last: page !== undefined && page > 1 ? WORKERS_PER_PAGE : undefined,
```
jodator, line 20: I think that we might have a similar code from the past. However, it is not the best solution for "paginated" results.

We have 2 kinds of queries from Hydra:

1.  Normal `ModelWhere` type query – it uses `offeset` and `limit` params and it can be used to query any "page" from all of the "pages"
2.  The `ModelConnection` type query - that uses `cursor`s – those are best suited for "infinite scroll"

The (2) returns `totalCount` – which is needed to properly count the number of pages for (1) :shrug: 

Compare that with `useMembers()` query: [https://github.com/Joystream/pioneer/blob/6eb67bfc61b0db5fa97c3d041709e700f5edd49a/packages/ui/src/memberships/hooks/useMembers.ts#L32-L55](https://github.com/Joystream/pioneer/blob/6eb67bfc61b0db5fa97c3d041709e700f5edd49a/packages/ui/src/memberships/hooks/useMembers.ts#L32-L55)

It uses both `Where` and `Conneciton` query.

packages/ui/src/working-groups/hooks/useWorkersPagination.ts
```diff
@@ -0,0 +1,43 @@
...
+import { useGetWorkersConnectionQuery } from '@/working-groups/queries'
+import { asWorkerBaseInfo } from '@/working-groups/types'
+
+export const WORKERS_PER_PAGE = 5
+
+interface UseWorkersPaginationProps extends UseWorkersProps {
+  page?: number
+}
+
+export const useWorkersPagination = ({ groupId: group_eq, statusIn, page }: UseWorkersPaginationProps) => {
```
jodator, line 13: :+1: for `*Pagination` suffix

jodator (changes requested):
I think that we have technical debt in other hooks.

DevNikita16:
I've resolved all your comments, but can't have a look because WG is not loading

jodator:
@DevNikita16 see #812

jodator (approved):

PR #815, opened 2021-06-11
📻 Close context menu on action select

### LINKED ISSUE #788: Close context menu on action select
The context menu popup in my roles page should be closed when invoking action.

p-sad:
Closes #788

### DIFF
--- a/packages/ui/src/common/components/ContextMenu/ContextMenu.tsx
+++ b/packages/ui/src/common/components/ContextMenu/ContextMenu.tsx
@@ -8,8 +8,13 @@ import { Animations, BorderRad, Colors, Shadows, Transitions } from '../../const
 import { ButtonGhost, ButtonLink } from '../buttons'
 import { KebabMenuIcon } from '../icons'
 
+export interface ContextMenuItem {
+  text: string
+  onClick: () => void
+}
+
 export interface ContextMenuProps {
-  items: { text: string; onClick: () => void }[]
+  items: ContextMenuItem[]
 }
 
 export interface ContextMenuAlignmentProps {
--- a/packages/ui/src/common/components/ContextMenu/ContextMenu.tsx
+++ b/packages/ui/src/common/components/ContextMenu/ContextMenu.tsx
@@ -1,6 +1,8 @@
 import item from '@polkadot/ui-keyring/options/item'
-import React, { useEffect, useRef, useState } from 'react'
+import React, { useRef, useState } from 'react'
 import styled, { css } from 'styled-components'
+
+import { useOutsideClick } from '@/common/hooks/useOutsideClick'
 
 import { Animations, BorderRad, Colors, Shadows, Transitions } from '../../constants'
 import { ButtonGhost, ButtonLink } from '../buttons'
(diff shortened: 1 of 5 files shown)


packages/ui/src/common/components/ContextMenu/ContextMenu.tsx
```diff
@@ -23,12 +24,43 @@ export const ContextMenu = ({ children, align }: ContextMenuProps & ContextMenuA
...
+    const outsideClickListener = (event: MouseEvent) => {
+      if (!event.target) return
+      const target = event.target as Node
+      const clickedOutside = !container.current?.contains(target)
+      if (!clickedOutside) return
+      setMenuVisible(false)
+      window.removeEventListener('mousedown', outsideClickListener)
+    }
+    addEventListener('mousedown', outsideClickListener)
+    return () => removeEventListener('mousedown', outsideClickListener)
```
jodator, line 39: Let's have full if's – that's a few lines more, but gives better overview.

packages/ui/src/common/components/ContextMenu/ContextMenu.tsx
```diff
@@ -1,19 +1,20 @@
...
 import styled, { css } from 'styled-components'
 
 import { Animations, BorderRad, Colors, Shadows, Transitions } from '../../constants'
-import { ButtonGhost } from '../buttons'
+import { ButtonGhost, ButtonLink } from '../buttons'
 import { KebabMenuIcon } from '../icons'
 
 export interface ContextMenuProps {
-  children: React.ReactNode
+  items: { text: string; onClick: () => void }[]
```
jodator, line 10:
```suggestion
  items: ContextMenuItem[]
```

I think that `{}[]` is a bit harder to read, because the information about an array type is "hidden" after the object.

packages/ui/src/common/components/ContextMenu/ContextMenu.tsx
jodator: This should be moved to a hook/helper as we have a similar function somewhere.

jodator (changes requested):
Some code style hints + I think that we can extract the click outside logic.

packages/ui/src/common/hooks/useOutsideClick.ts
```diff
@@ -0,0 +1,26 @@
+import React, { Dispatch, SetStateAction, useEffect } from 'react'
+
+export function useOutsideClick(
+  container: React.RefObject<any>,
+  isOpen: boolean,
+  setOpen: Dispatch<SetStateAction<boolean>>
+) {
+  useEffect(() => {
+    if (!isOpen) return
```
jodator, line 9:
```suggestion
    if (!isOpen) {
      return
    }
```

packages/ui/src/common/hooks/useOutsideClick.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    const closePopup = (event: MouseEvent) => {
+      event.stopPropagation()
+      if (!event.target) return
```
jodator, line 13:
```suggestion
      if (!event.target) {
        return
      }
```

jodator (approved):

PR #818, opened 2021-06-11
🥩 Implement WG average stake

### LINKED ISSUE #415: Display working group dashboard
Follow-up of [#289](https://github.com/Joystream/pioneer/issues/289).

Implement:

*   [x] current budget
*   [x] debt
*   [x] avg stake See: #394

As for now only, the "current budget" is described in the API. Waiting for other fields to be defined or implemented.

DevNikita16:
Closes #415


packages/ui/src/working-groups/model/getAverageStake.ts
```diff
@@ -0,0 +1,9 @@
+import BN from 'bn.js'
+
+import { WorkerFieldsFragment } from '@/working-groups/queries'
+
+export const getAverageStake = (workers: Pick<WorkerFieldsFragment, 'stake'>[]): BN => {
+  const totalStake = workers.reduce((a, b) => a + b.stake, 0)
```
jodator, line 6: Similarly to the other PR - you can use `BN` here.

packages/ui/src/working-groups/queries/workingGroups.graphql
```diff
@@ -49,7 +49,7 @@ fragment WorkingGroupFields on WorkingGroup {
     ...WorkingGroupMetdataFields
   }
   workers {
-    id
+    stake
```
jodator, line 52: So we don't use workers ids anywhwere? IIRC we changed how the list of workers is fetched, so this makes sense. Just to double check.

jodator (changes requested):
Just the BN thing to update and we can merge it.

packages/ui/src/working-groups/queries/workingGroups.graphql
DevNikita16: Yep, we catch workers separately from working group now.
That's why having worker ids is useless now.

packages/ui/src/working-groups/model/getAverageStake.ts
DevNikita16: ✅

jodator:
@DevNikita16 could you check merge conflicts as well?

jodator (approved):

PR #825, opened 2021-06-11
Global layout component

### LINKED ISSUE #755: Prepare page layout templates
I think that this is a good point to introduce page layout templates. The reason is that we're going to have multiple pages that have similar layouts but we always construct them by hand. We should define them once and just re-use them when appropriate.

*   React [template](https://medium.com/engineering-zemoso/atomic-design-in-react-react-native-using-a-theming-library-part-2-6aa6d3b9de13) from an atomic design example.

Requirements:

*   prepare a list of used layouts
*   add template components
*   prepare storybook for page layouts using empty components

ClumsyVlad:
Closes #755


packages/ui/src/app/pages/Profile/MyAccounts.tsx
```diff
@@ -13,18 +14,22 @@ export function MyAccounts() {
...
-      </PageHeader>
-      <Statistics>
-        <TokenValueStat title="Total balance" tooltipText="Lorem ipsum..." value={total} />
-        <TokenValueStat title="Total transferable balance" tooltipText="Lorem ipsum..." value={transferable} />
-        <TokenValueStat title="Total locked balance" tooltipText="Lorem ipsum..." value={locked} />
-        <TokenValueStat title="Total recoverable" tooltipText="Lorem ipsum..." value={recoverable} />
-      </Statistics>
-      <Accounts />
-    </AppPage>
+    <PageLayout
```
jodator, line 17: :+1:

packages/ui/src/app/components/PageLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+      main={content}
+      lowSidebar={lowSidebar}
+      highSidebar={highSidebar}
+      footer={footer}
+    >
+      {header}
+      {content}
+      {lowSidebar}
+      {highSidebar}
+      {footer}
```
jodator, line 38: Maybe a distinction Layouts components would be better instead of one that does everything.

`PageLayout` => `PageWithHighSidebarLayout`, `PageWithLowSidebarLayout`, etc

packages/ui/src/app/components/PageLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+      return PageLayoutWithFooter
+    } else if (props.main && props.lowSidebar && !props.highSidebar && !props.footer) {
+      return PageLayoutWithLowSidebar
+    } else if (props.main && props.lowSidebar && props.footer && !props.highSidebar) {
+      return PageLayoutWithLowSidebarAndFooter
+    } else if (props.main && props.highSidebar && !props.lowSidebar && !props.footer) {
+      return PageLayoutWithHighSidebar
+    } else if (props.main && props.highSidebar && props.footer && !props.lowSidebar) {
+      return PageLayoutWithHighSidebarAndFooter
+    }
```
jodator, line 115: That's hard to parse, so maybe split it as in :point\_up: comment.

ClumsyVlad: Check HistoryTab.tsx (inner content separation needed)

PR #826, opened 2021-06-11
🫁 Implement WG debt

### LINKED ISSUE #415: Display working group dashboard

DevNikita16:
This PR together with #818 close #415

I've done counting debt only for active workers, because later we can just add leaving status to our WHERE

### DIFF
--- a/packages/ui/src/working-groups/hooks/useGroupDebt.ts
+++ b/packages/ui/src/working-groups/hooks/useGroupDebt.ts
@@ -1,0 +1,21 @@
+import BN from 'bn.js'
+
+import { WorkerStatusTypename } from '@/working-groups/types'
+
+import { useGetGroupDebtQuery } from '../queries'
+
+export const useGroupDebt = (groupId: string) => {
+  const { data, loading } = useGetGroupDebtQuery({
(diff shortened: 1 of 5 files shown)


packages/ui/src/working-groups/hooks/useGroupDebt.ts
```diff
@@ -0,0 +1,21 @@
...
+  })
+
+  if (loading || !data) {
+    return {
+      debt: null,
+    }
+  }
+
+  return {
+    debt: new BN(data.workers.reduce((a, b) => a + b.missingRewardAmount, 0)),
```
p-sad, line 19: Depending on how much debt can be racked up, `BN` might be a better option for an accumulator than `number` - not too important yet but might be worth keeping in mind

p-sad (approved):

packages/ui/src/working-groups/hooks/useGroupDebt.ts
jodator: It might be a good option. A bit future proof as well :+1:

jodator (approved):

PR #827, opened 2021-06-11
Select member roles

### LINKED ISSUE #692: Implement Roles select

### LINKED ISSUE #828: Add a test suite for `common/utils` functions

thesan:
Closes  #692
Closes #828

### DIFF
--- a/packages/ui/src/common/components/selects/SimpleSelect.tsx
+++ b/packages/ui/src/common/components/selects/SimpleSelect.tsx
@@ -188,7 +188,7 @@ const Options = styled.div`
   top: 100%;
   user-select: none;
   min-width: 100%;
-  z-index: 10;
+  z-index: 51;
 `
 
 const OptionsFooter = styled.div`
--- a/packages/ui/src/common/components/selects/SimpleSelect.tsx
+++ b/packages/ui/src/common/components/selects/SimpleSelect.tsx
@@ -193,4 +193,7 @@ const Options = styled.div`
 
 const OptionsFooter = styled.div`
   display: flex;
+  & > * {
+    flex-grow: 1;
+  }
 `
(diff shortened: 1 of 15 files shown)


thesan:
It looks like #817 breaks the member filters box layout

packages/ui/src/common/components/selects/SimpleSelect.tsx
```diff
@@ -188,9 +188,12 @@ const Options = styled.div`
   top: 100%;
   user-select: none;
   min-width: 100%;
-  z-index: 10;
+  z-index: 51;
```
jodator, line 191: Let's put that to constants/styles.ts -> ZIndex variable

packages/ui/src/common/utils.ts
```diff
@@ -9,19 +9,24 @@ export const isString = (something: unknown): something is string => typeof some
 export const isRecord = (something: unknown): something is Record<string, any> =>
   typeof something === 'object' && something !== null
 
+interface EqualsOption {
+  checkExtraKeys?: boolean
+  deep?: boolean
+}
+
 export const objectEquals = <T extends Record<string, any>>(
```
jodator, line 17: This one would require a test. There's a lot going on and a test could clear up how it should be used.

jodator (changes requested):
The component works very well :) I'd like to ask you only to add a test and move the z-index declaration to constant so we could have better z-index management.

PR #830, opened 2021-06-14
🐣 Setup base Proposals mocks and queries

### LINKED ISSUE #800: Setup base Proposals mocks and queries
The Proposal schema is almost ready – we should start working with the mocks close to the target API as soon as possible.

### Requirements:

- setup base Proposal types
- stub Proposals mocks
- intorduce queries to fetch:
  - current proposals
  - past proposals
  - a proposal

DevNikita16:
Closes #800 

I've left `usePastProposals.ts` but actually somebody will need to remove it, when doing filters.
So to fetch active or past proposals just `useProposals.ts` is used. 
Also I defined such things as statusToTypename and typenameToStatus, I guess these are good helpers


DevNikita16:
Ok, i see that i need to resolve conflicts

DevNikita16:
Also I introduced new script `queries:parse` 
it's working as expected, replacing `implements Some, Any` to `implements Some & Any`
But I can't run it together with `queries:generate`

packages/ui/dev/scripts/parseSchema.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const parse = async () => {
+  const schemaPath = join(__dirname, '..', '..', 'src', 'common', 'api', 'schemas', 'schema.graphql')
+  let schemaContent = await readFileSync(schemaPath, 'utf-8')
+
+  schemaContent = schemaContent.replace(/\b(implements)\b(.*?),((\s*).*?,)?/g, '$1$2 &$3')
+  await writeFileSync(schemaPath, schemaContent, { encoding: 'utf-8' })
+}
+
+parse()
```
jodator, line 12: Should be something like patch or update. Parse does not tell that you alter the file

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
```diff
@@ -22,7 +22,7 @@ export const ProposalList = ({ proposals, isPast }: ProposalListProps) => (
     </ListHeaders>
     <List>
       {proposals.map((proposal) => (
-        <ProposalListItem key={proposal.id} {...proposal} isPast={isPast} />
+        <ProposalListItem key={proposal.id} proposal={proposal} />
```
jodator, line 25: The `isPast` should stay AFAICS

packages/ui/src/working-groups/components/OpeningsList/OpeningsList.stories.tsx
```diff
@@ -2,9 +2,10 @@ import { Meta, Story } from '@storybook/react'
 import React from 'react'
 import { MemoryRouter } from 'react-router'
 
-import { getMockAsOpening } from '@/mocks/data/mockOpenings'
 import { getMockAsUpcomingOpening } from '@/mocks/data/seedUpcomingOpening'
 
+import { getMockAsOpening } from '../../../mocks/data/seedOpenings'
```
jodator, line 7: Let's keep the `@/mocks/` prefix.

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
jodator: OK I see that you've changed the ListItem logic there. ATM I'm torn between:

1.  Having it consistent with other lists (the prop is set from the outside)
2.  Having it automatically check

WDYT?

packages/ui/package.json
```diff
@@ -4,6 +4,7 @@
   "license": "GPL-3.0-only",
   "scripts": {
     "build": "node --max_old_space_size=4096 ./node_modules/.bin/webpack --mode production --progress",
+    "queries-parse": "ts-node dev/scripts/parseSchema.ts",
     "queries:generate": "graphql-codegen --config codegen.config.yml && yarn lint:fix",
```
jodator, line 8:
```suggestion
    "queries:patch": "ts-node dev/scripts/parseSchema.ts",
    "queries:generate": "graphql-codegen --config codegen.config.yml && yarn lint:fix",
    "queries": "yarn queries:patch && yarn queries:generate"
```

you can do something like the above

jodator, line 8:
```suggestion
    "queries:patch": "ts-node dev/scripts/parseSchema.ts",
    "queries:generate": "graphql-codegen --config codegen.config.yml && yarn lint:fix",
    "queries": "yarn queries:patch && yarn queries:generate"
```

you can do something like the above

jodator (changes requested):
That's a lot of changes, but I'm glad that it works very well :)

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
DevNikita16: From my point of view and as I would prefer to do: just check if it's past or not inside component. Because we can invoke isProposalActive in any place, so I would prefer to leave it as it is now.

jodator (approved):

PR #831, opened 2021-06-14
📦 Extract Markdown editor package

### LINKED ISSUE #819: Extract markdown editor package
I think that extracting markdown editor building from the UI package will improve build times.

jodator:
Closes #819


jodator:
Extracting component ATM might be premature. It looks like we might need to add some components to display some date inside the editor so it would be better to have them together.

Issue #821, opened 2021-06-11
The useOpenings query don't query proper values


jodator:
The openings should be loaded using `Where` query.

We should also add `useOpeningsPagination()` query.

p-sad:
Closed by #840

PR #840, opened 2021-06-15
🐢 Add Openings hook version with pagination

p-sad:
See #821


packages/ui/src/working-groups/components/OpeningsList/OpeningsList.tsx
```diff
@@ -6,14 +6,14 @@ import { UpcomingWorkingGroupOpening, WorkingGroupOpening } from '@/working-grou
 import { OpeningsListRow } from './OpeningsListRow'
 
 export interface OpeningsListProps {
-  openings: WorkingGroupOpening[] | UpcomingWorkingGroupOpening[]
+  openings?: WorkingGroupOpening[] | UpcomingWorkingGroupOpening[]
```
jodator, line 9: Could you double check if this is needed? If used together with `isLoading` check you can add guard for this and always require openings list (like in the LoadingOpenings component)

packages/ui/src/working-groups/components/OpeningsList/OpeningsPagination.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useOpeningsPagination } from '@/working-groups/hooks/useOpeningsPagination'
+
+type Props = UseOpeningsParams
+
+export const OpeningsPagination = (props: Props) => {
+  const [page, setPage] = useState(1)
+  const { isLoading, openings, pageCount } = useOpeningsPagination({ ...props, page })
+
+  if (isLoading) return <Loading />
+  if (openings.length === 0) return <TextBig>No openings found</TextBig>
```
jodator, line 17: I'm torn here :D Usually, I'd prefer:

```suggestion
  if (isLoading) {
    return <Loading />
  }
  if (openings.length === 0) {
    return <TextBig>No openings found</TextBig>
  }
```

Just because single line ifs may lead to errors (if you forgot to add bracket when refactoring. Also, using braces always is more consistent with other code.

packages/ui/src/working-groups/hooks/useCountOpenings.ts
```diff
@@ -1,9 +1,9 @@
 import { useCountWorkingGroupOpeningsQuery } from '@/working-groups/queries'
 
-export function useCountOpenings(groupId: string) {
+export function useCountOpenings(groupId?: string) {
   const { data, loading } = useCountWorkingGroupOpeningsQuery({
-    variables: { groupId_eq: groupId, status_json: { isTypeOf_eq: 'OpeningStatusOpen' } },
+    variables: { where: { group_eq: groupId, status_json: { isTypeOf_eq: 'OpeningStatusOpen' } } },
```
jodator, line 5: :+1:

jodator, line 5: :+1:

jodator (approved):
Just two things to check and we can merge

packages/ui/src/working-groups/components/OpeningsList/OpeningsList.tsx
p-sad: Turns out I took care of that in the hook some commits later. Not needed indeed

PR #844, opened 2021-06-16
♟ Use runtime IDs in Groups extrinsics

### LINKED ISSUE #781: Use runtime id as extrinsics params
We've assumed that the `id` in the query node will be the same as and `id` on the blockchain.

However, the objects related to Working Groups return `runtimeId` alongside `id`. That's because the items are indexed per working group, not globally. In other words a `Worker` with `id=forumWorkingGroup-1` belongs to forum working group and has runtime id `1`.

p-sad:
Closes #781


DevNikita16 (commented):
Code looks fine!
But I thought this PR will fix Apply For Role Modal, but it's still not working

So should there be another ticket?

jodator:
> So should there be another ticket?

Either new PR for #781 or let's fix all "runtimeId" in extrinsics where: `Worker`, `Opening` are used. I don't think that we have anything more to handle ATM.

p-sad:
I completely missed Openings having runtimeIds as well. It shouldn't crash now.

DevNikita16 (approved):
👍✌️

PR #854, opened 2021-06-17
🪐 Add proper popup for role tags

### LINKED ISSUE #850: Proper description for role tooltip
https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=3420%3A2

DevNikita16:
Closes #850


packages/ui/src/proposals/model/proposalDetails.ts
```diff
@@ -1,5 +1,5 @@
-import { lowerFirstLetter } from '@/common/helpers'
-import { ProposalDetails } from '@/proposals/types'
+import { lowerFirstLetter } from '../../common/helpers'
+import { ProposalDetails } from '../types'
```
jodator, line 2: Could you check you settings? This change looks like unnecessary

packages/ui/src/proposals/model/proposalStatus.ts
```diff
@@ -1,5 +1,5 @@
-import { capitalizeFirstLetter, lowerFirstLetter } from '@/common/helpers'
-import { ProposalStatus } from '@/proposals/types'
+import { lowerFirstLetter, capitalizeFirstLetter } from '../../common/helpers'
+import { ProposalStatus } from '../types'
```
jodator, line 2: Same here ;)

packages/ui/src/memberships/components/MemberRoles.tsx
```diff
@@ -85,6 +96,32 @@ export const MemberRoles = ({ size, max, wrapable, roles }: MemberRolesProps) =>
   )
 }
 
+export interface MemberRolePopupContentProps {
+  roles: MemberRole[]
+}
+
+export const MemberRolePopupContent = ({ roles }: MemberRolePopupContentProps) => {
+  const groupAddress = `#/working-groups/${groupNameToURLParam(roles[0].groupName)}`
```
jodator, line 104: It might be worth checking if this can't be constructed using react-router utils. Because we might change the routing creation mechanism and then `#` will break that.

packages/ui/src/memberships/components/MemberRoles.tsx
```diff
@@ -85,6 +96,32 @@ export const MemberRoles = ({ size, max, wrapable, roles }: MemberRolesProps) =>
...
+  return (
+    <>
+      {roles.map((role, index) => (
+        <PopupRoleItem key={index}>
+          <PopupRoleTitle>
+            {memberRoleTitle(role)} {roles.length > 1 ? index + 1 : ''}
+          </PopupRoleTitle>
+          {role.createdAt && (
```
jodator, line 112: How big the effort it would be to make `createdAt` obligatory?

jodator (changes requested):
Works great :)

I'd change the PR a bit:
- The URL should be generated using react-router utils if possible
- It might be worth making `createdAt` obligatory
- Also, this component doesn't look that good in storybook (we should have story updated for `createdAt`).

Also, I was thinking about a test, but maybe a storybook is enough for that.

packages/ui/src/proposals/model/proposalDetails.ts
DevNikita16: Well, we use `proposalDetails.ts` and `proposalStatus.ts` in mock generators, and it couldn't find file when `@/...` is used. Maybe You have solution for this?

packages/ui/src/memberships/components/MemberRoles.tsx
DevNikita16: We have so called Member Role(default one) and it doesn't have time

packages/ui/src/memberships/components/MemberRoles.tsx
```diff
@@ -101,7 +102,8 @@ export interface MemberRolePopupContentProps {
 }
 
 export const MemberRolePopupContent = ({ roles }: MemberRolePopupContentProps) => {
-  const groupAddress = `#/working-groups/${groupNameToURLParam(roles[0].groupName)}`
+  const groupAddress = generatePath(`/working-groups/${groupNameToURLParam(roles[0].groupName)}`)
```
jodator, line 105: I think that we can inline `/working-groups/${groupNameToURLParam(roles[0].groupName)}` because `generatePath()` should be used when you'd like to properly use params object:

```js
 generatePath("/user/:id/:entity(posts|comments)", {
  id: 1,
  entity: "posts"
});
```

With one parameter I think that with one parameter it is enough to use a template string.

jodator (changes requested):
Just one more thing ;)

packages/ui/src/memberships/components/MemberRoles.tsx
DevNikita16: Yep, forgot to undo

jodator (approved):

PR #862, opened 2021-06-18
🔫 Select components - assorted fixes

### LINKED ISSUE #855: Selected account doesn't display in account selector
Selected account used to show up in the selector but doesn't

p-sad:
Turns out account selectors didn't display selected accounts because the selected account was given to them by the form they're in via props. It makes sense since the form state should be the "master index" of all values filled in ­— so I've made sure to avoid confusion by completely removing the selected item from select components' state - it can only be supplied by props now. The select components' main job is to pass the correct data back to form upon selection.
I have also noticed that "unsaved accounts" weren't being passed back to the form (while their internal state made them display the account, meaning that displayed values were inconsistent with the form state) so I fixed that as well. This use case is a bit odd/specific to the Account Selector, so I went ahead and added a test case for it, as well as a click-selection sanity check.
Closes #855


jodator (approved):
Great summary & thanks for the PR :)

Issue #479, opened 2021-04-29
Improve modal with stepper abstraction


jodator:
ATM I've used a local reducer and magic numbers for steps. We can anticipate a modal with 10+ steps IRC. Having numbers in the code for each step isn't the best practice.

In this terms, I would try to either:

*   introduce simple reducer with typed steps
*   use existing state-machine library from React ecosystem

PR #863, opened 2021-06-18
🪜 Base stepper abstraction

jodator:
Adds base stepper abstraction for modals – see #479.

I've used [xstate](https://xstate.js.org/docs/about/concepts.html#finite-state-machines) to abstract various stages of modals. The idea behind it is to introduce base machines for common tasks:

*   forms validation
*   transactions
*   stake gathering
*   etc

and be able to construct complex flows programmatically by composing them.

Additionally, by using the `meta` property you can define that some steps are used to display a stepper in the modals. This is automatically parsed from the state machine in the stepper UI.

What's left to do:

*   [better typings](https://xstate.js.org/docs/guides/typescript.html)
*   transaction state machine
*   form validation machine together with saving form state


DevNikita16 (approved):
The base machine code was a bit hard to understand, but it's because of complicated logic.

Then when I just read code in modals, it looks fine and readable)

👍

PR #864, opened 2021-06-18
Update stepper

jodator:
I think that this should be merge as #863

See #479


jodator:
Closing in favour of #863.

PR #881, opened 2021-06-22
🗺️ Map events to activities

p-sad:
A possible way we could map events to avtivities/notifications. Having separate hooks/queries for user-centric/group-centric activities might be a necessity - I don't think a generic query can be written, and appropriate events need to be hand-picked.


packages/ui/src/working-groups/hooks/useGroupActivities.ts
```diff
@@ -0,0 +1,25 @@
...
+            ...data.appliedOnOpeningEvents.map(asAppliedOnOpeningActivity),
+            ...data.applicationWithdrawnEvents.map(asApplicationWithdrawnActivity),
+            ...data.budgetSpendingEvents.map(asBudgetSpendingActivity),
+          ].sort((a, b) => b.createdAt.localeCompare(a.createdAt))
+        : [],
+    [data, loading]
+  )
+
+  return { isLoading: loading, activities }
+}
```
jodator, line 25: That's an improvement as it allows to query for the actual events for the working group.

I'd rather see that we could have this returned from the API. However, If we'd get that from the API in the future it looks like the change would be limited to this hook and all the dependant code should stay the same.

packages/ui/src/common/components/Activities/ActivityContent.tsx
```diff
@@ -21,30 +21,30 @@ interface Props {
...
-    case 'StakeDecreasedEvent':
-    case 'StakeIncreasedEvent':
+    case 'StakeDecreased':
+    case 'StakeIncreased':
       return <StakeChangedContent activity={activity} />
-    case 'WorkerExitedEvent':
+    case 'WorkerExited':
       return <WorkerExitedContent activity={activity} />
-    case 'WorkerStartedLeavingEvent':
+    case 'WorkerStartedLeaving':
```
jodator, line 47: Why this change? It looks like you only dropped `Event` suffix. I did not follow the whole code so I'm asking here.

jodator (commented):
Two quick questions for start. Otherwise the rest of changes makes sense AFAICS :)

packages/ui/src/common/components/Activities/ActivityContent.tsx
p-sad: I originally used `Event` because I thought I could come up with a generic query and then sort and map by `__typename`, and in that case it would be reasonable to just copy `__typename` into the object.
At the present state, I don't think this matters much and we might as well cut the `Event` for brevity

jodator:
OK. I think that this is a good start, so we should go in this direction. Thoughs for another PRs:

*   I'm thinking about testing the `use...Activites()` hook. However, I'm worried that we'd be testing our resolver implementation.
*   Separate hooks for related queries are the way to go IMO
*   We need to order the results to get the latest events first (as those are most important) -> this makes a good candidate for tests IMO

Issue #866, opened 2021-06-18
Base Proposal Statistics


thesan:
Implement the proposal page statistics

(Part of #801)

PR #884, opened 2021-06-23
🍫 Implement the StatisticBar component

thesan:
See #866
(Reuse and refactor some code from #671)

### DIFF
--- a/packages/ui/src/common/components/statistics/FractionValue.tsx
+++ b/packages/ui/src/common/components/statistics/FractionValue.tsx
@@ -1,0 +1,22 @@
+import React, { ReactNode } from 'react'
+import styled from 'styled-components'
+
+import { Colors } from '@/common/constants'
+
+import { NumericValue } from './NumericValueStat'
+
+export interface FractionValueProps {
+  numerator: ReactNode | number
+  denominator: ReactNode | number
+}
+
+export const FractionValue = ({ numerator, denominator }: FractionValueProps) => (
+  <NumericValue>
+    {numerator}
+    <DenominatorValue>/ {denominator}</DenominatorValue>
+  </NumericValue>
+)
+
+const DenominatorValue = styled.span`
+  color: ${Colors.Black[400]};
+`
--- a/packages/ui/src/common/components/statistics/FractionValue.tsx
+++ b/packages/ui/src/common/components/statistics/FractionValue.tsx
@@ -11,12 +11,17 @@ export interface FractionValueProps {
 }
 
 export const FractionValue = ({ numerator, denominator }: FractionValueProps) => (
-  <NumericValue>
+  <Fraction>
     {numerator}
     <DenominatorValue>/ {denominator}</DenominatorValue>
-  </NumericValue>
+  </Fraction>
 )
+
+const Fraction = styled(NumericValue)`
+  font-size: 16px;
+`
 
 const DenominatorValue = styled.span`
   color: ${Colors.Black[400]};
+  font-weight: 400;
 `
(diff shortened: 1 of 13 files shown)


packages/ui/src/common/components/statistics/FractionValue.tsx
```diff
@@ -0,0 +1,22 @@
...
+export interface FractionValueProps {
+  numerator: ReactNode | number
+  denominator: ReactNode | number
+}
+
+export const FractionValue = ({ numerator, denominator }: FractionValueProps) => (
+  <NumericValue>
+    {numerator}
+    <DenominatorValue>/ {denominator}</DenominatorValue>
+  </NumericValue>
```
jodator, line 17: I think that we have a similar component. I'm not sure if it can be re-used here, so if not then this one is OK.

packages/ui/src/common/components/statistics/StatisticBar.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <Statistics>
+      <TwoRowStatistic>
+        <StatisticBar {...args} numerator={floor(10 * value)} denominator={`${ceil(10 * threshold)} votes`} />
+        <StatisticBar {...args} numerator={`${ceil(100 * ratio)}%`} denominator={`${floor(100 * threshold)}%`} />
+      </TwoRowStatistic>
+      <TwoRowStatistic>
+        <StatisticBar {...args} value={0} threshold={0.1} numerator="0" denominator="max 2 votes" />
+        <StatisticBar {...args} value={0} threshold={0.1} numerator="0%" denominator="max 10%" />
```
jodator, line 28: Could you pass the calculated values here as well?

packages/ui/src/common/components/statistics/StatisticBar.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { StatsContent, StatisticHeader, StatisticHeaderProps } from '.'
+import { FractionValue, FractionValueProps } from './FractionValue'
+
+export interface StatisticBarProps extends StatisticHeaderProps, FractionValueProps {
+  value: number
+  threshold: number
+}
+
+export const StatisticBar = ({ value, threshold, numerator, denominator, ...headerProps }: StatisticBarProps) => (
+  <div>
```
jodator, line 16: I needed to dig into why there's a `<div>` and not the `<StaticsticBlock>` so maybe adding a  component that will tell more would be good?

I see this is meant to be used as:

```
<TwoRowStatistic>
  <div>...</div>
  <div>...</div>
</TwoRowSatistic>
```

So maybe an exported component Like `<TwoRowStatisticRow>` would be enough?

YOu could also add a generic example to: [https://deploy-preview-884--pioneer-2-storybook.netlify.app/?path=/story/common-statistics--statistic](https://deploy-preview-884--pioneer-2-storybook.netlify.app/?path=/story/common-statistics--statistic.) page on how to use them together.

jodator (changes requested):
The component works great. I had only small asks about the storybook. The extraction of `StatisticHeader` and other components makes sense 👍 .

There's also some mismatch of the item from the design (despite the width):

Design | App
---|---
Screenshot 2021-06-23 at 12 18 04 | Screenshot 2021-06-23 at 12 18 31

The `2 / 6 votes` and percentage should not be bolded

packages/ui/src/common/components/statistics/FractionValue.tsx
thesan: I looked for an existing component like in `WorkingGroupsOpening.tsx`: WorkingGroupsOpening, but it doesn't seem to be implemented yet. Did I miss an already existing component?

jodator: No, that's OK than. No need to change this. I wanted to be sure if we can't re-use.

packages/ui/src/common/components/statistics/StatisticBar.tsx
thesan: Since this kind of statistic is also used in the usual statistic layout like in [6.3.1](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2014%3A595) I wanted to export a single statistic. But after adding it to the `Statistics.stories.tsx` I realize I should export a `React.Fragment` to make it reusable

jodator (approved):

PR #891, opened 2021-06-24
📺 Fix wide button and 'no members' block layout

ClumsyVlad:


jodator (approved):
👍 It's OK.

But let's remember that 3rd component like this must be extracted as a common one.

PR #892, opened 2021-06-24
🐝 Handle required stake and staking account in add proposal modal

### LINKED ISSUE #835: Proposal required stake check
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=6%3A13)

Add staking prerequest check for selected proposal.

See Apply for role modal for details – a similar check for given proposal must be implemented, where we display tha user either
- has no account with a required stake
- has accounts with compatible balance
- has accounts with compatible locked balance

If user has required stake do not display anything (proceed to the next step).

### LINKED ISSUE #836: Stake step
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=6%3A565)

Add staking step to the create proposal modal.

DevNikita16:
Closes #836 
Closes #835

### DIFF
--- a/packages/ui/src/proposals/modals/AddNewProposal/AddNewProposalModal.tsx
+++ b/packages/ui/src/proposals/modals/AddNewProposal/AddNewProposalModal.tsx
@@ -44,7 +44,9 @@ export const AddNewProposalModal = () => {
   const { hideModal, showModal } = useModal<AddNewProposalModalCall>()
   const [state, send, service] = useMachine(addNewProposalMachine)
   const constants = useConstants(state.context.proposalType)
-  const { hasRequiredStake, transferableAccounts, accountsWithLockedFounds } = useHasRequiredStake(10000000000)
+  const { hasRequiredStake, transferableAccounts, accountsWithLockedFounds } = useHasRequiredStake(
+    constants?.requiredStake.toNumber() || 0
+  )
   const [isValid, setValid] = useState<boolean>(false)
 
   const [txParams] = useState<NewProposalParams>({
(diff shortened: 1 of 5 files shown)


packages/ui/src/proposals/modals/AddNewProposal/AddNewProposalModal.tsx
```diff
@@ -56,20 +64,29 @@ export const AddNewProposalModal = () => {
...
+      }
+
+      if (feeInfo && !feeInfo.canAfford) {
+        return send('FAIL')
+      }
     }
 
-    if (feeInfo && !feeInfo.canAfford) {
-      return send('FAIL')
+    if (state.matches('generalParameters.stakingAccount') && !hasRequiredStake) {
```
jodator, line 81: So, the main benefit of using a state machine is to have multiple `boolean` values turned into distinct states. I think that we can model this state into the machine.

There should be another step between `selectType` and `generalDetails` in which we check the required stake, it can transition to either:

*   `generalDetails` or
*   `requiredStakeFail` (final state) -> in which we display the "required stake" modals

jodator (changes requested):
@DevNikita16 Hey. Let's focus on one task at a time. So closing 2 tickets in one PR is better to be avoided. Even though, the steps are small and might be done together it doesn't mean that they have to be done together.

I'll comment only on the #835 part as this PR should be split into two. You can always create a new branch on top of the previous PR and rebase on the main after it got merged.

DevNikita16:
@jodator OK, i have introduced requiredStakeVerification and requiredStakeFailed steps

But I think I need to write test for required stake, unfortunately was not able to do this now

Will try tomorrow also!

PR #898, opened 2021-06-28
🎒 Map Member notifications

p-sad:
A small batch of member notifications. There are a couple of things to consider still:
* `AppliedOnOpening` and `ApplicationWithdrawn` events don't contain Worker ID; if none is planned to be added, we might need to run another query to retrieve worker's applications before these two events can be fetched
* Group activities reference members by name, while Role activities and Member notifications refer to the selected member/worker as "you". There are a couple approaches one could take:
a) Create another set of notification texts, this is pretty bad because it involves duplicating components instead of reusing them
b) Parametrize name vs pronoun by fetching current member and their roles either when displaying the activities list, or whilst mapping events to activity objects. I think the latter would be most efficient.


jodator:
>     * `AppliedOnOpening` and `ApplicationWithdrawn` events don't contain Worker ID; if none is planned to be added, we might need to run another query to retrieve worker's applications before these two events can be fetched

If I understand this correctly, those can't contain `Worker` since there will be none. Only `Member` will be available. So this event should reference `Member` and `Opening`. If we have both, then it is OK to have another query for details. If there's no `Member` nor `Openign` we should ask to add them in Hydra.





>     * Group activities reference members by name, while Role activities and Member notifications refer to the selected member/worker as "you".

I'd go with one component with a prop that would select the proper form of the message. Two strings to translate would be OK. Two components rather not.

packages/ui/src/memberships/hooks/useMyRoleIds.ts
```diff
@@ -0,0 +1,16 @@
+import { useMemo } from 'react'
+
+import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
+import { useGetWorkerIdsQuery } from '@/working-groups/queries'
```
jodator, line 4: It should go to `@/working-groups` since it's a working group related. A broader search is `useMyWorkers()`.

I wonder if there's a need for this hook, or couldn't we re-use `useMyWorkers()` – I think that this is not the case here probably. If those hooks are used on different pages it makes sense to have two of them.

jodator (changes requested):
I'd move the `useMyRolesIds()` to working-groups namespace.

packages/ui/src/memberships/hooks/useMyRoleIds.ts
p-sad: I just wanted to minimize the data payload - not sure if it's all that needed, but Worker fields seem to be the among the heftiest query results at the moment

jodator (approved):

Issue #801, opened 2021-06-10
Base proposal page


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=2022%3A107529)

### Requirements

- Setup Proposal wireframe page
- Title with a copy link
- Proposal stage badge with time left
- Voting dashboard
- Proposal-specific dashboard (placeholder -> should allow displaying different stats depending on proposal type)
- Rationale (long markdown field)
- Discussion (separate ticket: #868)
- Side bar:
  - Votes component (placeholder)
  - Proposal stages history (placeholder)
  - Proposer

jodator:
Closing this issue, as we have the base proposal page already.

PR #899, opened 2021-06-28
🧮 Add the ProposalStatistics section

### LINKED ISSUE #866: Base Proposal Statistics

### LINKED ISSUE #901: Add a test for the proposal page
https://github.com/Joystream/pioneer/pull/899#pullrequestreview-693925785
> The other thing is to add a test for the proposal page. I think that it'll be safer for us to start adding those for more complex pages.

thesan:
Closes #866
Closes #901 
See #801


thesan:
There's no proposal constants for the `fillWorkingGroupLeadOpening` proposals. Thus the Proposal with the id 1 (in the newly generated proposals) doesn't show any statistics.

packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
```diff
@@ -82,7 +85,7 @@ export const ProposalPreview = () => {
         <MainPanel ref={sideNeighborRef}>
           <RowGapBlock gap={24}>
             {/* Statistics */}
-            <Statistics></Statistics>
+            {voteCount && constants && <ProposalStatistics constants={constants} voteCount={voteCount} />}
```
jodator, line 88: I think that we can remove those empty comments now. They just duplicate the code below.

The other thing is that maybe we could show "empty" state as well. I think it will be better to show something while the data is loading rather than not showing anything.

I'd go with a simple `-` (similarly to what we show when loading accounts data for instance).

packages/ui/src/proposals/types/proposalsDetailed.ts
```diff
@@ -0,0 +1,18 @@
+import { asBlock, Block } from '@/common/types'
+import { ProposalDetailedFieldsFragment, VoteFieldsFragment } from '@/proposals/queries'
+
+import { asProposal, Proposal } from './proposals'
+
+export interface ProposalDetailed extends Proposal {
```
jodator, line 6: Maybe we should go with `ProposalWithDetails` just because we already have `MemberWithDetails` and `WorkerWithDetails`.

The benefit of the `FooWithDetails` is that you can write `Foo...` and see what other forms are there with autocomplete (usually that'd be `Foo` and `FooWithDetails`. Also the cast functions names have the same order as the base object, e.g.: `asFooWithDetails()` vs `asDetailedFoo` (for `FooDetailed`).

jodator (changes requested):
I had only one minor comment about naming types.

The other thing is to add a test for the proposal page. I think that it'll be safer for us to start adding those for more complex pages.

packages/ui/src/proposals/hooks/useConstants.ts
```diff
@@ -14,9 +14,41 @@ export const useConstants = (proposalType?: ProposalDetails): ProposalConstants
...
+  ['setMembershipLeadInvitationQuota', 'setMembershipLeadInvitationQuotaProposalParameters'],
+  ['setMembershipPrice', 'setMembershipPriceProposalParameters'],
+  ['setReferralCut', 'setReferralCutProposalParameters'],
+  ['setWorkingGroupLeadReward', 'setWorkingGroupLeadRewardProposalParameters'],
+  ['signal', 'signalProposalParameters'],
+  ['slashWorkingGroupLead', 'slashWorkingGroupLeadProposalParameters'],
+  ['terminateWorkingGroupLead', 'terminateWorkingGroupLeadProposalParameters'],
+  ['unlockBlogPost', 'unlockBlogPostProposalParameters'],
+  ['updateWorkingGroupBudget', 'updateWorkingGroupBudgetProposalParameters'],
+  ['veto', 'vetoProposalProposalParameters'],
```
jodator, line 53: I had to check that – looks weird, but that's what API gives us 😉

PR #902, opened 2021-06-29
🧬 Adapt activity text to context

p-sad:
Changed member notifications to refer to the selected member in 2nd person. It's done through a prop in `ActivitiesBlock` and introduces some unfortunate prop drilling, but this way seems to be most efficient overall.

I've also refactored `ActivityContent` to use a `Record` as a map, instead of an elaborate switch. The final version of the refactor is pretty much on par with the previous one, gaining some strengths in one way while getting weaker in another.
The Good:
* `Record` enforces mapping of all activity categories to components, making sure no category is omitted by mistake
* `ActivityContent` is much easier on the eyes

The Bad:
* Type checking in respective content components is completely gone. However, this shouldn't pose a problem as long as activity contents are only used in the `ActivityContent` component.

### DIFF
--- a/packages/ui/src/app/pages/Proposals/PastProposals.tsx
+++ b/packages/ui/src/app/pages/Proposals/PastProposals.tsx
@@ -17,9 +17,9 @@ import { ProposalsTabs } from './components/ProposalsTabs'
 
 export const PastProposals = () => {
   const searchSlot = useRef<HTMLDivElement>(null)
-  const [filters, setFilters] = useState(ProposalEmptyFilter)
+  const [, setFilters] = useState(ProposalEmptyFilter)
 
-  const { types, stages } = usePastProposals({ filters })
+  const { types, stages } = usePastProposals()
   const { isLoading, proposals } = useProposals({ status: 'past' })
 
   const activities = useActivities()
--- a/packages/ui/src/common/components/Activities/Activities.stories.tsx
+++ b/packages/ui/src/common/components/Activities/Activities.stories.tsx
@@ -2,6 +2,10 @@ import { Meta, Story } from '@storybook/react'
 import BN from 'bn.js'
 import React from 'react'
 import { MemoryRouter } from 'react-router'
+import styled from 'styled-components'
+
+import { ActivitiesBlock } from '@/common/components/Activities/ActivitiesBlock'
+import { Activity } from '@/common/types'
 
 import { ModalContext } from '../../providers/modal/context'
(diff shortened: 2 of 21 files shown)


jodator (changes requested):
Good job 👍

I've removed the use of React.FC as it was too broad. It was used before as it was quicker to implement.

The only thing to change is to how we construct messages. I think that we should keep in mind the future Internationalization of the project (or start it sooner than later).

The changes I've done in 4db8be3 are mostly in mind to ease i18n later. We should have in mind that those messages will need to be mapped to some strings (ids or full messages) so we should avoid conditionals in the middle of the message. The separate messages or sets of components will be easier to translate:

```ts
export const WorkerStartedLeavingContent: ActivityContentComponent = ({ activity, isOwn }) => {
  const { member } = activity as WorkerStartedLeavingActivity

  if (isOwn) {
    return <>You left role and you are still in the unstaking period.</>
  }

  return (
    <>
      <MemberModalLink call={{ modal: 'Member', data: { id: member.id } }}>{member.handle} </MemberModalLink>
      left a role and is still in the unstaking period.
    </>
  )
}
```

can be easily translated using [react-i18n-next](https://react.i18next.com/latest/using-with-hooks#translate-your-content) using either [hook](https://react.i18next.com/latest/usetranslation-hook#what-it-does) or [Trans](https://react.i18next.com/latest/trans-component#samples) component.

```typescript
export const WorkerStartedLeavingContent: ActivityContentComponent = ({ activity, isOwn }) => {
  const { t } = useTranslation()
  const { member } = activity as WorkerStartedLeavingActivity

  if (isOwn) {
    return <>{t('You left role and you are still in the unstaking period.')}</>
  }

  return (
    <Trans i18nKey="activities.otherWorkerLeft">
      <MemberModalLink call={{ modal: 'Member', data: { id: member.id } }}>{member.handle} </MemberModalLink>
      left a role and is still in the unstaking period.
    </Trans>
  )
}
```

I've used this lib as an example but the rules should stay the same for other libs.

jodator:
> \* Type checking in respective content components is completely gone

I've added this: [`3528d56`](https://github.com/Joystream/pioneer/pull/902/commits/3528d565186bb2b3c112ca11d345a90d1e150254) WDYT?

p-sad:
I like the changes, especially the re-added type checking on activities.
I've added separation of own/not own text in two components that you seem to have missed, I think the PR is good to go now

jodator (approved):

PR #904, opened 2021-06-29
🥋 Handle proposal details step in new proposal

### LINKED ISSUE #837: General parameters step – title & rationale
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=6%3A863)

Allow user to fill:
- title
- rationale (using mardkown editor)

The currently selected member is the proposer to display.

Add side-modal with help for "how to write a good rationale?".

DevNikita16:
Closes #837

### DIFF
--- a/packages/ui/src/proposals/modals/AddNewProposal/components/ProposalDetailsStep.tsx
+++ b/packages/ui/src/proposals/modals/AddNewProposal/components/ProposalDetailsStep.tsx
@@ -1,12 +1,17 @@
-import React from 'react'
+import React, { useCallback } from 'react'
+import styled from 'styled-components'
 
 import { CKEditor } from '@/common/components/CKEditor'
 import { InputComponent, InputText } from '@/common/components/forms'
+import { Link } from '@/common/components/Link'
 import { Row } from '@/common/components/Modal'
 import { RowGapBlock } from '@/common/components/page/PageContent'
 import { TextMedium } from '@/common/components/typography'
+import { Colors } from '@/common/constants'
+import { useModal } from '@/common/hooks/useModal'
 import { SelectMember } from '@/memberships/components/SelectMember'
 import { Member } from '@/memberships/types'
+import { RationaleModalCall } from '@/proposals/modals/AddNewProposal/components/RationaleModal'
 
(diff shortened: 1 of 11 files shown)


packages/ui/src/proposals/modals/AddNewProposal/components/ProposalDetailsStep.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <InputComponent label="Proposal title" required inputSize="m" id="field-title">
+            <InputText id="field-title" onChange={(event) => setTitle(event.target.value)} />
+          </InputComponent>
+          <InputComponent label="Rationale" required inputSize="auto" id="field-rationale">
+            <CKEditor
+              id="field-rationale"
+              onChange={(event, editor) =>
+                setRationale(
+                  editor ? editor.getData() : ((event as unknown) as ChangeEvent<HTMLInputElement>).target.value
+                )
```
jodator, line 52: This should be as:

```suggestion
                setRationale(editor.getData())
```

Because we expect this to be "CKEditor". How we do this in the tests is another story :wink: (see below)

packages/ui/test/_mocks/components/CKEditor.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { CKEditorProps } from '@/common/components/CKEditor'
+import { Input } from '@/common/components/forms'
+
+const CKEditor = (props: CKEditorProps) => (
+  <Input
+    id={props.id || 'ck-test'}
+    name={props.id}
+    type="text"
+    autoComplete="off"
+    onChange={(props.onChange as unknown) as ChangeEventHandler<Element>}
```
jodator, line 12:
```suggestion
    onChange={(event) => onChange(event, { getData: () => event.target.value } ) }
```

I'd go with a mock that has the same interface – this way you don't need to change the production code if the tests implementation change.

I think that the test _should not_ (as long as possible) influence the production code.

jodator (changes requested):
Looks good 👍 I've added some comments on types and mocking the editor.

ps.: A storybook for the rationale side modal is also needed.

PR #910, opened 2021-06-30
🤑 Filter accounts by compatible balance for the stake step

### LINKED ISSUE #907: Filter accounts by compatible balance for the stake step
In the stake, step user can select an account with 0 balance – which should not happen.

Since we already check for valid stake we know that there's at least one account so we're safe displaying the select.

However, I think that the accounts list should be filtered to display only the accounts which have proper balance.

The filter method should use the same algorithm we use for making the required stake check.

DevNikita16:
Closes #907

### DIFF
--- a/packages/ui/src/accounts/components/SelectAccount/SelectAccount.tsx
+++ b/packages/ui/src/accounts/components/SelectAccount/SelectAccount.tsx
@@ -1,5 +1,7 @@
+import BN from 'bn.js'
 import React, { useEffect, useMemo, useState } from 'react'
 
+import { useMyBalances } from '@/accounts/hooks/useMyBalances'
 import { useKeyring } from '@/common/hooks/useKeyring'
 import { Address } from '@/common/types'
 
@@ -9,7 +11,7 @@ import { accountOrNamed } from '../../model/accountOrNamed'
 import { isValidAddress } from '../../model/isValidAddress'
 import { Account } from '../../types'
 
-import { filterByText } from './helpers'
+import { filterByMinBalance, filterByText } from './helpers'
 import { OptionAccount } from './OptionAccount'
 import { OptionListAccount } from './OptionListAccount'
 
@@ -23,11 +25,16 @@ interface Props {
   filter?: (option: Account) => boolean
   selected?: Account
   disabled?: boolean
+  minBalance?: BN
 }
 
-export const SelectAccount = React.memo(({ onChange, filter, selected }: Props) => {
+export const SelectAccount = React.memo(({ onChange, filter, selected, minBalance }: Props) => {
   const { allAccounts } = useMyAccounts()
-  const options = allAccounts.filter(filter || (() => true))
+  let options = allAccounts.filter(filter || (() => true))
+  if (minBalance) {
+    const balances = useMyBalances()
+    options = filterByMinBalance(options, balances, minBalance)
+  }
   const [search, setSearch] = useState('')
 
   const filteredOptions = useMemo(() => filterByText(options, search), [search, options])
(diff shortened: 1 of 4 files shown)


packages/ui/src/proposals/modals/AddNewProposal/components/StakingAccountStep.tsx
```diff
@@ -33,7 +34,11 @@ export const StakingAccountStep = ({ requiredStake, account: chosenAccount, setA
             </TextMedium>
           </RowGapBlock>
           <InputComponent label="Select account for Staking" required inputSize="l">
-            <SelectAccount onChange={(account) => setAccount(account)} selected={chosenAccount} />
+            <SelectAccount
+              onChange={(account) => setAccount(account)}
+              selected={chosenAccount}
+              filter={(account) => useBalance(account.address)?.transferable.gte(requiredStake) ?? true}
```
jodator, line 40: Two things here:

1.  Hooks must be placed at the top of the component since [rules of the hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) (there's a slight possibility that hooks will render in a different order).
2.  Please extract this check – because it might be not that simple. For instance, we should check (later) if there is a lock that would allow reusing locked balance for this purpose.

jodator (changes requested):
Unfortunately, you can't use hooks in loops – there should be always the same number of hooks calls on re-renders. I think that `useMyBalances()` can be used here.

packages/ui/src/proposals/modals/AddNewProposal/components/StakingAccountStep.tsx
DevNikita16: I've decided to introduce new prop `minBalance` for selected account component, I think it should be more clear than writing filter each time

packages/ui/src/accounts/components/SelectAccount/SelectAccount.tsx
```diff
@@ -23,11 +25,16 @@ interface Props {
...
 }
 
-export const SelectAccount = React.memo(({ onChange, filter, selected }: Props) => {
+export const SelectAccount = React.memo(({ onChange, filter, selected, minBalance }: Props) => {
   const { allAccounts } = useMyAccounts()
-  const options = allAccounts.filter(filter || (() => true))
+  let options = allAccounts.filter(filter || (() => true))
+  if (minBalance) {
+    const balances = useMyBalances()
+    options = filterByMinBalance(options, balances, minBalance)
```
jodator, line 36: A hook can't be inside `if` as it _might_ break the rules of hooks.

packages/ui/src/accounts/components/SelectAccount/SelectAccount.tsx
```diff
@@ -23,11 +25,16 @@ interface Props {
   filter?: (option: Account) => boolean
   selected?: Account
   disabled?: boolean
+  minBalance?: BN
 }
 
-export const SelectAccount = React.memo(({ onChange, filter, selected }: Props) => {
+export const SelectAccount = React.memo(({ onChange, filter, selected, minBalance }: Props) => {
```
jodator, line 31: The problem with `minBalance` is that it mixes this logic with filters and might be very limiting. At this stage, I can't be sure that we will filter only by minimal required balance or do we allow a "compatible" lock to be present as well.

Remember, that locked value can be reused for different purposes. So if you have 2 locks and would like to stake (create 3rd lock) that can be set with those 2 we'd need to have a more complex check (filter) for this.

Right now, the simple totalBalance >= minStake is used as it is good enough. We can add a better check later, but we should be prepared to update this check without the need to modify the component itself (something along the lines of open-closed principle).

jodator (changes requested):
Hi @DevNikita16 – let's make the filter passed from the outside as I'm not sure if the check will not be different for different staking purposes.

DevNikita16:
Fine, I understood everything!

PR #914, opened 2021-07-01
🥄 Add Opening Filled activity

p-sad:
Also comes with a little refactor of mock events seeds


jodator (changes requested):
Nice, that's a lot of changes but it looks like it will be easier to maintain it.

One change, though ;)

PR #921, opened 2021-07-02
💘 Fix tooltip arrow

ClumsyVlad:


packages/ui/src/common/components/Tooltip/Tooltip.tsx
```diff
@@ -72,7 +73,7 @@ export const Tooltip = ({
     onFocus: mouseIsOver,
     onBlur: mouseLeft,
     onMouseEnter: mouseIsOver,
-    onMouseLeave: mouseLeft,
+    // onMouseLeave: mouseLeft,
```
jodator, line 76: This should be reverted or removed?

packages/ui/src/common/components/Tooltip/Tooltip.tsx
```diff
@@ -94,6 +95,7 @@ export const Tooltip = ({
                 {...attributes.popper}
                 {...popUpHandlers}
                 isTooltipActive={isTooltipActive}
+                forBig={forBig}
```
jodator, line 98: Maybe we could use the [arrow](https://popper.js.org/docs/v2/modifiers/arrow/) from popper?

jodator, line 98: Maybe we could use the [arrow](https://popper.js.org/docs/v2/modifiers/arrow/) from popper?

jodator (commented):
Maybe you could check the [arrow modifier](https://popper.js.org/docs/v2/modifiers/arrow/)?

packages/ui/src/common/components/Tooltip/Tooltip.tsx
ClumsyVlad: I've tried, it doesn't work for me :(

Issue #913, opened 2021-07-01
Base Proposal History


thesan:
Implement the history sidebar
(Part of #801)

PR #928, opened 2021-07-02
🦶 Extract the Stepper to use it for the proposal history

thesan:
See #913


packages/ui/src/common/components/Stepper/Stepper.tsx
```diff
@@ -34,53 +34,43 @@ export const Stepper = ({ steps, theme = StepperTheme.light }: StepperProps) =>
   const stepsToRender = asStepsToRender(steps)
 
   return (
-    <ThemeProvider theme={theme}>
-      <StepperWrap>
-        {stepsToRender.map((step, index) => (
-          <StepWrap key={index} {...step}>
```
thesan: My understanding is that the tests were failing because the theme context is not available during the first render, so using [`waitForElement`](https://react-testing-examples.com/jest-rtl/styled-components/) on the tests which were failing would have solved the problem. But since the theme is only used within this component I figure it would be easier to just pass the theme as a prop and rewrite the styles a little.

packages/ui/src/common/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { TextInlineSmall } from '@/common/components/typography'
+import { BorderRad, Colors, Transitions } from '@/common/constants'
+import { spacing } from '@/common/utils/styles'
+
+import { StepperTheme } from './themes'
+import { asStepsToRender, StepperStep, StepToRender } from './types'
+
+export interface StepperProps {
+  steps: StepperStep[]
+  theme?: typeof StepperTheme.light
```
jodator, line 14: I'd go with:

```suggestion
  theme?: keyof typeof StepperTheme
```

And have a default `dark`. The usage will be:

```typescript
<Stepper steps={} theme="light" />
```

instead of importing the object, the component itself would choose the proper theme style object or pass it to the styled-components, WDYT?

jodator (changes requested):
I'd change one thing, other than that it looks good 👍

packages/ui/src/common/components/Stepper/Stepper.tsx
thesan: Right! that looks better :+1:

jodator (approved):

PR #930, opened 2021-07-05
🪨 Add a script for generating event mocks separately

### LINKED ISSUE #924: Add a script that will only generate specific mocks
Recently I've been adding mappings for events and because we only have a script that generates all the data, nearly every PR that adds new schema objects has massive diffs that only make it more difficult to read. Maybe it's worth it to add a separate script that will generate only additional data based on already existing files.

p-sad:
Closes #924 
Only introduced for events for now, since they don't have as much interdependency as other schema objects.


packages/ui/dev/scripts/generateEventMocks.ts
```diff
@@ -0,0 +1,57 @@
...
+  const eventTypes = distinctParams.map(type => type.toString()).filter(type => type in eventGenerators)
+
+  let newMocks: { [key: string]: any[] } = {}
+
+  if (eventTypes === undefined) {
+    newMocks = generateAllEvents(mocks)
+  } else {
+    eventTypes.forEach((type) => {
+      const t = type as keyof typeof eventGenerators
+      newMocks[type] = eventGenerators[t](mocks)
```
p-sad, line 50: TS is unfortunately incapable of inferring the type from the `filter` in line 41; if anyone has an idea how to get it working please let me know

PR #931, opened 2021-07-06
🦀 Add worker leaving + status changed activities

p-sad:
* Added worker exited + started leaving activities. Might want to add a more descriptive text for the worker exited activity, to at least hint at what they were doing and in which group
* Added activities for `StatusTextChangedEvent`. There might be more activities tied to it that haven't been designed yet, e.g. cancelled upcoming openings. The event itself is quite confusing overall, it has fields for changed metadata and added openings but then it also has a "result" type field which potentially doubles both above mentioned fields and at the same time seems to hint at them being mutually exclusive... I think it could use a redesign.


jodator (approved):
LGTM 👍

Issue #933, opened 2021-07-06
Notifications- Remaining activities


p-sad:
||Working group|Worker|Member|Proposal|Forum
| - | - | - | - | - | -
|ApplicationWithdrawnEvent|:heavy_check_mark:||:heavy_check_mark:!||
|AppliedOnOpeningEvent|:heavy_check_mark:|:heavy_check_mark:!|:heavy_check_mark:!||
|BudgetSetEvent|:heavy_check_mark:||||
|BudgetSpendingEvent|:heavy_check_mark:||||
|InitialInvitationBalanceUpdatedEvent|||||
|InitialInvitationCountUpdatedEvent|||||
|InvitesTransferredEvent|||||
|LeaderInvitationQuotaUpdatedEvent|||||
|LeaderSetEvent|||||
|LeaderUnsetEvent|||||
|MemberAccountsUpdatedEvent|||||
|MemberInvitedEvent|||||
|MemberProfileUpdatedEvent|||||
|MemberVerificationStatusUpdatedEvent|||||
|MembershipBoughtEvent|||||
|MembershipPriceUpdatedEvent|||||
|NewMissedRewardLevelReachedEvent|||||
|OpeningAddedEvent|:heavy_check_mark:||||
|OpeningCanceledEvent|:heavy_check_mark:||||
|OpeningFilledEvent|:heavy_check_mark:|x!|x!||
|ProposalCancelledEvent||||:heavy_check_mark:|
|ProposalCreatedEvent||||:heavy_check_mark:|
|ProposalDecisionMadeEvent||||:heavy_check_mark:|
|ProposalDiscussionPostCreatedEvent||||:heavy_check_mark:|
|ProposalDiscussionPostDeletedEvent||||:heavy_check_mark:|
|ProposalDiscussionPostUpdatedEvent||||:heavy_check_mark:|
|ProposalDiscussionThreadModeChangedEvent||||:heavy_check_mark:|
|ProposalExecutedEvent||||:heavy_check_mark:|
|ProposalStatusUpdatedEvent||||:heavy_check_mark:|
|ProposalVotedEvent|||as voter!|:heavy_check_mark:|
|ReferralCutUpdatedEvent|||||
|RewardPaidEvent|||||
|StakeDecreasedEvent|:heavy_check_mark:|:heavy_check_mark:|w:heavy_check_mark:||
|StakeIncreasedEvent|:heavy_check_mark:|:heavy_check_mark:|w:heavy_check_mark:||
[...]

PR #936, opened 2021-07-06
🦑 Add queries for existing activities

p-sad:
See #933 
A bit of housekeeping, added queries for the last few events that had components already made for them.
Also went back on my decision to divert the `eventType` parameter from the event's `__typename`, because turns out it's useful and efficient to have it this way.


p-sad:
Also, forgot to mention that `BudgetSetEvent` is still displayed inaccurately because on the design it's displayed as "Budget increased/decreased". Still not sure how to do it elegantly and efficiently (you need to know the previous `BudgetSetEvent` to tell if it increased or decreased

jodator (approved):
LGTM 👌

PR #940, opened 2021-07-07
🔧 Fix the `useProposals` status filter

### LINKED ISSUE #939: The `proposals/current` tab is now empty

thesan:
Closes #939


jodator (approved):
Thanks! IIRC the `_in` operator currently does not work with Hydra, but it should be fixed so the fix is good.

PR #942, opened 2021-07-07
🫖 Remove `LeaderSetActivity`

p-sad:
The component's design disappeared from [figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Design?node-id=1405%3A69471) so it looks like it's not needed anymore.


jodator (approved):
:ok:

PR #944, opened 2021-07-07
🦗 Add Terminated Worker/Leader activities

p-sad:
See #933


packages/ui/src/working-groups/components/Activities/WorkerTerminatedContent.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const WorkerTerminatedContent: ActivityContentComponent<WorkerTerminatedActivity> = ({ activity, isOwn }) => {
+  const { member, groupName, eventType } = activity
+  const isLeader = eventType === 'TerminatedLeaderEvent'
+  return isOwn ? (
+    <>You have been terminated by {getGoverningBody(isLeader, groupName)}.</>
+  ) : (
+    <>
+      <MemberModalLink call={{ modal: 'Member', data: { id: member.id } }}>{member.handle}</MemberModalLink> has been
+      terminated by {getGoverningBody(isLeader, groupName)}.
```
jodator, line 18: I'd split those into two components or three messages (i18n) as before.

jodator (changes requested):
Minor change to the activity text content. Other than that looks nice 👍

packages/ui/src/working-groups/components/Activities/WorkerTerminatedContent.tsx
p-sad: I was under the impression that extracting it into a separate function would help with internationalization. I'll try to experiment with the two other options

jodator: Well, it might but it might be hard to translate this.

_Usually_ the translator would see something like this:

```
"%s has been terminated by the council"
"%s has been terminated by %s Working Group Lead"
```

vs separated strings

```
"%s has been terminated by %s"
"council"
"%s Working Group Lead"
```

It might be hard seeing those properly translate in the context.

jodator (approved):
👌

Issue #790, opened 2021-06-10
Handle recoverable balance


jodator:
Follow-up for the #303:

*   Display recoverable balances
*   Allow recovering balance

See [solution 1](https://github.com/Joystream/pioneer/issues/790#issuecomment-937706768) for the implementation details.

jodator:
AFAICS the recoverable balance is the "Staking for voting" kind. It can be removed when other locks are present.

jodator:
OK, I've found out one bug in the code. The locks were using wrong ids for some actions. ATM I was able to create locks for:

*   staking candidate (11)
*   membership worker (09)
*   member invitation (10)

jodator:
So AFAICS the only recoverable lock is Staking/Voting. The logic to recover balance is updated to handle this.

> \- recoverable balance: the amount by which the free balance can increase if one removes an idle lock. Idle locks are locks that are just present as a result of some staking in the past. The only such kind of lock is a lock for voting. Importantly, you may have an idle lock on an account but a recoverable balance of 0 on that account, this can happen if there is another lock on the same account which is at least as large, and so it still freezes all the funds, an nothing is recoverable. A user should probably still remove any idle locks, as they may cause inconvenience in the future anyway.

I've also found in the API releasing council candidacy stake. I've found two calls to release stake:  
1\. Referendum: (**`VotingLockId`**) API to release: **`api.tx.referendum.releaseVoteStake()`**  
2\. Candidacy: (**`CandidacyLockId`**) API to release: **`api.tx.council.releaseCandidacyStake()`**

jodator:
So AFAICS the only recoverable lock is Staking/Voting. The logic to recover balance is updated to handle this.

> \- recoverable balance: the amount by which the free balance can increase if one removes an idle lock. Idle locks are locks that are just present as a result of some staking in the past. The only such kind of lock is a lock for voting. Importantly, you may have an idle lock on an account but a recoverable balance of 0 on that account, this can happen if there is another lock on the same account which is at least as large, and so it still freezes all the funds, an nothing is recoverable. A user should probably still remove any idle locks, as they may cause inconvenience in the future anyway.

I've also found in the API releasing council candidacy stake. I've found two calls to release stake:  
1\. Referendum: (**`VotingLockId`**) API to release: **`api.tx.referendum.releaseVoteStake()`**  
2\. Candidacy: (**`CandidacyLockId`**) API to release: **`api.tx.council.releaseCandidacyStake()`**

[...]

PR #950, opened 2021-07-08
🔒 Update locks icons

jodator:
See #790. I was missing the Lock icon, so I've added it and refactored some code.


packages/ui/src/common/components/icons/locks/LockIcon.tsx
```diff
@@ -0,0 +1,27 @@
...
+  className?: any
+}
+
+export const LockIcon = React.memo(({ className }: Props) => (
+  <Icon
+    preserveAspectRatio="xMidYMid meet"
+    color="currentColor"
+    className={className}
+    size="16"
+    viewBox="0 0 24 24"
```
jodator, line 15: Could you check why this was needed? If I set `viewBox="0 0 16 16"` the lock was cropped.

PR #966, opened 2021-07-09
🚀 Improve tests performance

jodator:
Comparison on my local machine:

Before:

After:

Few findings:

*   `findByRole()` is very slow, most of my changes relate to that by adding `getButton()` helper. See the "AddNewProposalModal.test.tsx" above :scream:
*   The `--runInBand` has no effect on CI
*   Not so great improvement on CI :disappointed:
*   Some server seeds can be created once per file (only reads, no writes) or reducing number of seeds (especially with Members)

### DIFF
--- a/packages/ui/test/_helpers/getButton.ts
+++ b/packages/ui/test/_helpers/getButton.ts
@@ -1,0 +1,5 @@
+import { screen } from '@testing-library/react'
+
+export async function getButton(text: string | RegExp) {
+  return (await screen.findByText(text, { selector: 'span' })).parentElement!
+}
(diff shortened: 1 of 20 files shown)


packages/ui/test/_helpers/getButton.ts
```diff
@@ -0,0 +1,5 @@
+import { screen } from '@testing-library/react'
+
+export async function getButton(text: string | RegExp) {
```
DevNikita16, line 3: 👍👍👍

DevNikita16 (approved):

p-sad (approved):

Issue #868, opened 2021-06-21
Discussion preview


jodator:
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=116%3A9113)

The discussion component of the proposal page requires:
- Threaded discussion display
- Allowing to mention someone
- Post replay to selected message
- Coply a link to message
- Handle discussion mode: Open / Closed

jodator:
Most of the tasks are handled besides post reply and edit which will be handled in the #1067

PR #978, opened 2021-07-09
🗣️  Implement the Forum/Comment component 

thesan:
See #868
See #801


jodator (approved):
Overall I don't see anything big to change, so 👍 from me.

I've asked alse @ClumsyVlad to check that but there shouldn't be bigger changes.

jodator:
OK, So @ClumsyVlad also aproved. Merging

PR #980, opened 2021-07-12
🤌 Add rules of hooks ESLint plugin

### LINKED ISSUE #972: Add ESLint rule for proper hooks usage
It's easy to forget about [rules of hooks](https://reactjs.org/docs/hooks-rules.html) so let's make the machine [police us](https://reactjs.org/docs/hooks-rules.html).

The setup should catch the errors like the one fixed in #971

p-sad:
Closes #972 
The plugin also includes a ruleset for ensuring that all variables present in Effect/Memo hooks are included in the dependencies list, but I'd opt for discussing it before including it, for interesting reasons.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,7 +13,8 @@
     "plugin:import/typescript"
   ],
   "plugins": [
-    "import"
+    "import",
+    "react-hooks"
   ],
   "ignorePatterns": [".eslintrc.js"],
   "parser": "@typescript-eslint/parser",
@@ -68,7 +69,8 @@
       {
         "noUselessIndex": true
       }
-    ]
+    ],
+    "react-hooks/rules-of-hooks": "error"
   },
   "overrides": [
     {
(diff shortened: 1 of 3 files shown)


jodator:
> The plugin also includes a ruleset for ensuring that all variables present in Effect/Memo hooks are included in the dependencies list, but I'd opt for discussing it before including it, for interesting reasons.

Could you write what's that?

p-sad:
@jodator There are two reasons. The minor one is that as far as I understand, it is supposed to make sure all used variables are watched, which may not always be needed/correct in a given business case. By default it's supposed to be set to warning, but depending on what we want in case of each hook, we may end up with some warnings staying forever.

The more important issue I have with it is tied to a bug (feature?) in this plugin. When presented with deps such as these:
```ts
useEffect( () => {...}, [member?.id])
```
it can crash with the unchecked exception `TypeError: Cannot read property 'type' of undefined` because it cannot handle `member` being possibly undefined very well. For it to work, we would need to change all instances of the above and similar code to explicitly state undefined:
```ts
useEffect( () => {...}, [member?.id ?? undefined])
```
all the while it'll still be possible that someone misses this and gets greeted with an unchecked exception that they'll be confused about the meaning of.

jodator (commented):
🚀

PR #982, opened 2021-07-12
🏧 Add proposal funding request

### LINKED ISSUE #871: Add Proposal: Funding request
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=9%3A9368)

Handle proposal's specific parameters:

- [ ] Amount
- [ ] Account

DevNikita16:
Closes #871


DevNikita16:
@ClumsyVlad I was not adding highlighted fields, because I think we don't have proper styles for them 
I found something similar in ApplyForRole modal, but probably their also weird, at least not like in figma
image

jodator (approved):
I think that we can continue with this.

Issue #803, opened 2021-06-11
Proposal: Create Working Group Lead opening


jodator:
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=267%3A3920)

Handle proposals specific properties:

- [x] Working group
- [x] Required stake
- [x] Unstaking period
- [x] Reward per X blocks (or per block)
- [x] Opening Description with preview

PR #987, opened 2021-07-14
🦨 Add Working Group Lead Opening proposal preview details

p-sad:
See #803 
I used the statistic components for proposal details, but these components only vaguely resemble what's present in the design. I think we need a proper grid layout that looks more like the designs.
The modal with opening description isn't done yet, I'm not sure whether it should be global or a local child node.


DevNikita16 (approved):
Good job 👍

PR #1026, opened 2021-07-20
📉 Add proposal decrease group lead stake

### LINKED ISSUE #874: Add Proposal: Decrease Working Group Lead stake
@[figma](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=232%3A4185)

Handle proposal's specific parameters:

- [x] Group Select
- [x] Current Leader
- [x] Decrease stake value + shortcut buttons

DevNikita16:
Closes #874 

All groups, except Membership has leaders


packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
```diff
@@ -0,0 +1,123 @@
...
+          )}
+          {groupId && group && !group.leaderId && (
+            <Info
+              title="Warning"
+              content={
+                <TextMedium>
+                  {capitalizeFirstLetter(group.name)} Working Group has no any Leader yet. Please choose other Group.
+                </TextMedium>
+              }
+            ></Info>
```
jodator:
```suggestion
            />
```

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
```diff
@@ -0,0 +1,123 @@
...
+              </InputComponent>
+              {group && (
+                <Info
+                  content={
+                    <TextMedium>
+                      The actual stake for {capitalizeFirstLetter(group.name)} Working Group Lead is{' '}
+                      <TextInlineMedium bold>{formatTokenValue(group.leaderWorker?.stake)} JOY</TextInlineMedium>.
+                    </TextMedium>
+                  }
+                ></Info>
```
jodator:
```suggestion
            />
```

This might be a collapsed block as it has no children.

jodator (approved):
I think that we should refactor this after we add all steps.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
jodator: I've refactored this `<Info/>` component

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
```diff
@@ -0,0 +1,118 @@
...
+            required
+            inputSize="l"
+            tooltipText="Please select an identifier for Working Group"
+          >
+            <SelectWorkingGroup selectedGroupId={groupId} onChange={(selected) => setGroupId(selected.id)} />
+          </InputComponent>
+          {(!groupId || isGroupLoading || (group && leader)) && (
+            <>
+              <InputComponent label="Working Group Leader" inputSize="l" disabled>
+                <SelectMember onChange={() => true} disabled selected={leader} />
```
jodator, line 69: This should be selected option or empty option component. Not the whole select.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <RowGapBlock gap={20}>
+          <InputComponent
+            label="Working Group"
+            required
+            inputSize="l"
+            tooltipText="Please select an identifier for Working Group"
+          >
+            <SelectWorkingGroup selectedGroupId={groupId} onChange={(selected) => setGroupId(selected.id)} />
+          </InputComponent>
+          {(!groupId || isGroupLoading || (group && leader)) && (
```
jodator, line 66: This check makes the form below to flash. I think that it would be better to display empty data if the group is not selected. This way the UI will not flash/jump.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
```diff
@@ -0,0 +1,118 @@
...
+  stakingAmount,
+  groupId,
+  setStakingAmount,
+  setGroupId,
+  setWorkerId,
+}: DecreaseWorkingGroupLeadStakeProps) => {
+  const [amount, setAmount] = useNumberInput(0, stakingAmount)
+  const setHalf = () => setAmount(stakingAmount ? stakingAmount.divn(2).toString() : '')
+  const setThird = () => setAmount(stakingAmount ? stakingAmount.divn(3).toString() : '')
+
```
jodator, line 41: This should use current working group leader stake. So if current WG Lead stake is 6000 JOY:
- set half will always set 3000 JOY
- set third will always set 2000 JOY
The logic (callback) will change when the group is changed.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
```diff
@@ -0,0 +1,118 @@
...
+interface DecreaseWorkingGroupLeadStakeProps extends DecreaseWorkingGroupLeadStakeParameters {
+  setStakingAmount: (amount: BN) => void
+
+  setGroupId(groupId: string): void
+
+  setWorkerId(workerId?: number): void
+}
+
+export const DecreaseWorkingGroupLeadStake = ({
```
jodator, line 31: Could add a stories as well? You can see how other steps stories were created.

jodator (changes requested):
I like the cleanup you've made with the mock commands :+1:

I was worried that we stub too much in the tests, but I don't see a better option ATM.

The only thing to change is the step logic and how we display things as I don't think that the form should be hidden when no group is selected.

Also, when the group has no leader everything should be disabled (inputs, buttons, etc).

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
DevNikita16: I am not sure that I can add stories for this step. Because it uses worker mocks

jodator: You can always add them ;)

[https://github.com/Joystream/pioneer/blob/c4f14898a3b7f40af3136920a177a37596817118/packages/ui/src/mocks/components/storybook/MockApolloProvider.tsx#L35-L38](https://github.com/Joystream/pioneer/blob/c4f14898a3b7f40af3136920a177a37596817118/packages/ui/src/mocks/components/storybook/MockApolloProvider.tsx#L35-L38)

jodator (changes requested):
It works OK now, but I think that the stories will be helpful anyway.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
DevNikita16: ✅

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
DevNikita16: ✅

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/DecreaseWorkingGroupLeadStake.tsx
DevNikita16: ✅

jodator (approved):

PR #1034, opened 2021-07-21
🧹 Fix proposal pages browser console errors

### LINKED ISSUE #1033: Fix the dom warnings on proposals lists and and preview pages

thesan:
Closes #1033

### DIFF
--- a/packages/ui/src/common/types/helpers.ts
+++ b/packages/ui/src/common/types/helpers.ts
@@ -1,3 +1,1 @@
 export type Reducer<Accumulator, Value = Accumulator> = (acc: Accumulator, value: Value, index: number) => Accumulator
-
-export type MakeTransient<T, K extends keyof T & string> = Omit<T, K> & { [P in `$${K}`]: T[K] }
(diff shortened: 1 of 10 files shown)


packages/ui/src/common/types/helpers.ts
```diff
@@ -1,3 +1 @@
 export type Reducer<Accumulator, Value = Accumulator> = (acc: Accumulator, value: Value, index: number) => Accumulator
-
-export type MakeTransient<T, K extends keyof T & string> = Omit<T, K> & { [P in `$${K}`]: T[K] }
```
thesan: @jodator I can't figure out why [prettier fails on the template literal types](https://github.com/Joystream/pioneer/runs/3125801868) since [the current version is supposed to have typescript 4.3 support](https://github.com/prettier/prettier/blob/main/CHANGELOG.md#231), any idea ?

jodator: Hmm, hard to tell I didn't have that problem.

PR #1052, opened 2021-07-22
🥑 Apply filters to past proposals

### LINKED ISSUE #586: Apply filters to the past proposal list

p-sad:
Closes #586 
Filters work well with past proposals at the moment. Their behaviour for active proposals is currently undefined, but by design they aren't meant to used with Active ones. The only filter that would cause problems with active proposals would be the lifetime filter since active proposals don't have an end date.
Lifetime filter has been tested for start date only/end date only, as we don't have an OR clause resolver in our mocks. The intervals used are correct in theory however.

### DIFF
--- a/packages/ui/src/app/pages/Proposals/PastProposals.tsx
+++ b/packages/ui/src/app/pages/Proposals/PastProposals.tsx
@@ -17,10 +17,10 @@ import { ProposalsTabs } from './components/ProposalsTabs'
 
 export const PastProposals = () => {
   const searchSlot = useRef<HTMLDivElement>(null)
-  const [, setFilters] = useState(ProposalEmptyFilter)
+  const [filters, setFilters] = useState(ProposalEmptyFilter)
 
   const { types, stages } = usePastProposals()
-  const { isLoading, proposals } = useProposals({ status: 'past' })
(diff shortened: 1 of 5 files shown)


jodator:
> Filters work well with past proposals at the moment.

And it should be only for the past ones :+1:

DevNikita16 (changes requested):
Everything works well 👍 But you should provide proper text for select in types/stages

p-sad:
@DevNikita16 As in, a text filter for narrowing down the selection in the dropdown? That's not a bad idea, but I'd add it as a follow-up PR. I didn't even touch those components in this one

p-sad:
Added as issue #1103, will do right after I'm done with current task

DevNikita16:
I meant that the words here should be separated by space, you should use `camelCaseToText`
image

jodator:
> I meant that the words here(and in stages) should be separated by space, you should use `camelCaseToText`

This might be done together with #1103 if @p-sad didn't touch them yet and the plan is to do it in #1103.

p-sad:
Updated issue to include replacing camelCase with spaces

DevNikita16 (approved):

PR #1089, opened 2021-07-22
🦊 Display role's unstaking status

### LINKED ISSUE #714: Display role's unstaking state
In the Working Groups -> My Roles -> Role detail

*   [ ] Display that a role is in the unstaking process

This issue requires a research on how this can be implemented or if the schema update wouldn't be better (as Hydra could also provide this information)

A follow-up for #665.

p-sad:
...and account for `WorkerStatusLeaving` in My Roles in general.
Closes #714


packages/ui/src/working-groups/model/getRoleWarning.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function getRoleWarning(status: WorkerStatusTypename) {
+  if (status === 'WorkerStatusActive') return undefined
+  if (status === 'WorkerStatusLeaving')
+    return {
+      title: 'Unstaking period',
+      content: 'This role is in the unstaking period.',
+      isClosable: false,
+    }
+  return {
```
jodator, line 11: Maybe let's keep `{}` for ifs:

```suggestion
  if (status === 'WorkerStatusActive') {
    return undefined
  }
  
  if (status === 'WorkerStatusLeaving') {
    return {
      title: 'Unstaking period',
      content: 'This role is in the unstaking period.',
      isClosable: false,
    }
  }
  
  return {
```

jodator (approved):
Small codestyle issue to fix.

PR #1104, opened 2021-07-23
🔧 Fix the Opening statistic mismatch

### LINKED ISSUE #1030: Opening statistic mismatch
The [statistics](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4835793) of the Working Group Opening should be:

*   Time Left
*   Reward per x blocks (depending on WG)
*   Minimal stake
*   Applicants

The footer should have: Created instead of Hired

thesan:
Closes #1030

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
@@ -160,7 +160,7 @@ export const WorkingGroupOpening = () => {
       }
       footer={
         <PageFooter>
-          <BlockTime block={opening.createdAtBlock} layout="row" dateLabel="Hired" />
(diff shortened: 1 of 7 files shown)


thesan:
@jodator the design has Time Left: "1 d : 10 min" which is good but eg. "0 d : 1043 min" looks like bug. WDYT ?

jodator:
> @jodator the design has Time Left: "1 d : 10 min" which is good but eg. "0 d : 1043 min" looks like bug. WDYT ?

@thesan Yeah. I've also found that weird. I'd go with either:

1.  something that we already have in other places (relative time or something like that) 
2.  or add a way to display "best fit" format:
    *   1 day : 20 hours
    *   14 hours : 20 minutes
    *   1 week : 2 days

PR #1105, opened 2021-07-23
🪛 Fix deciding proposals voting stage tab

thesan:
Sorry about that, my IDE was hanging yesterday and I didn't realize this had not been staged my commit


jodator:
Thanks for catching that! :ok\_hand:

PR #1108, opened 2021-07-23
🤓 Validate wasm file and change typo

### LINKED ISSUE #1100: Validate WASM file in the runtime upgrade form
The WASM file type can be validated.

There's `isWasm()` helper in the `@polkadotl/util` package already.

### LINKED ISSUE #1101: Change the file attachment message
Instead of "The xxx was uploaded successfully!" the message should be "The xxx was loaded successfully" because the file is not uploaded but rather loaded into memory and used in transaction.

DevNikita16:
Closes #1100 
Closes #1101

### DIFF
--- a/packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
+++ b/packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react'
+import React, { useCallback } from 'react'
 import { useDropzone } from 'react-dropzone'
 import styled, { css } from 'styled-components'
 
@@ -28,38 +28,38 @@ interface ValidatedFile extends File {
 
 const MAX_FILE_SIZE = 3 * 1024 * 1024
 
+const getValidatedFiles = async (event: any): Promise<ValidatedFile[]> => {
+  const files = []
+
+  const fileList = event.dataTransfer ? event.dataTransfer.files : event.target.files
+  for (let i = 0; i < fileList.length; i++) {
+    const file = fileList.item(i)
+
+    const isValidWASM = await WebAssembly.validate(await file.arrayBuffer())
+    Object.assign(file, { isValidWASM })
+    files.push(file)
+  }
+
+  return files
+}
+
+const validator = (file: ValidatedFile) => {
+  if (!file.isValidWASM) {
+    return {
+      code: 'file-invalid-type',
+      message: 'not valid WASM file',
+    }
+  }
+
+  return null
+}
+
 export const RuntimeUpgrade = ({ setRuntime }: RuntimeUpgradeProps) => {
   const onDrop = useCallback(async (acceptedFiles: File[]) => {
     if (acceptedFiles.length) {
       setRuntime(await acceptedFiles[0].arrayBuffer())
     }
   }, [])
-
-  async function getValidatedFiles(event: any): Promise<ValidatedFile[]> {
-    const files = []
-
-    const fileList = event.dataTransfer ? event.dataTransfer.files : event.target.files
-    for (let i = 0; i < fileList.length; i++) {
-      const file = fileList.item(i)
-
-      const isValidWASM = await WebAssembly.validate(await file.arrayBuffer())
-      Object.assign(file, { isValidWASM })
-      files.push(file)
-    }
-
-    return files
-  }
-
-  const validator = (file: ValidatedFile) => {
-    if (!file.isValidWASM) {
-      return {
-        code: 'file-invalid-type',
-        message: 'not valid WASM file',
-      }
-    }
-
-    return null
-  }
 
   const {
     isDragActive,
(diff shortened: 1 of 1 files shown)


packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
```diff
@@ -76,24 +97,30 @@ export const RuntimeUpgrade = ({ setRuntime }: RuntimeUpgradeProps) => {
...
+                  <b>{file.name}</b> ({file.size} B) was not loaded because of "{errors.map((e) => e.message)}"
+                </AcceptedFileText>
+              </ReceivedFile>
+            ))}
+            {fileValidation && !fileValidation.isValid && (
+              <ReceivedFile key={fileValidation.file.name} valid={false}>
+                <AcceptedFileText>
+                  <b>{fileValidation.file.name}</b> ({fileValidation.file.size} B) was not loaded because it could be
+                  malicious
+                </AcceptedFileText>
```
jodator, line 120:
```suggestion
                <AcceptedFileText>
                  <b>{fileValidation.file.name}</b> ({fileValidation.file.size} B) was not loaded because it is not
                  valid WASM file.
                </AcceptedFileText>
```

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
```diff
@@ -26,9 +26,29 @@ const MAX_FILE_SIZE = 3 * 1024 * 1024
 
 export const RuntimeUpgrade = ({ setRuntime }: RuntimeUpgradeProps) => {
   const onDrop = useCallback(async (acceptedFiles: File[]) => {
-    const file = acceptedFiles[0]
-    setRuntime(await file.arrayBuffer())
+    if (acceptedFiles.length) {
+      setRuntime(await acceptedFiles[0].arrayBuffer())
+    }
   }, [])
+  const [fileValidation, setFileValidation] = useState<{ file: File; isValid: boolean } | undefined>()
```
jodator, line 33: I think that another variable is not needed here. Using [the react-dropzone way](https://react-dropzone.js.org/#section-extending-dropzone) allows using file objects with custom prop:

```javascript
interface ValidatedFile extends File {
  isValidWASM?: boolean
}
```

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
```diff
@@ -26,9 +26,29 @@ const MAX_FILE_SIZE = 3 * 1024 * 1024
...
+  async function getFiles(event: any) {
+    const files = []
+
+    const fileList = event.dataTransfer ? event.dataTransfer.files : event.target.files
+    for (let i = 0; i < fileList.length; i++) {
+      const file = fileList.item(i)
+
+      if (WebAssembly.validate(await file.arrayBuffer())) {
+        files.push(file)
+        setFileValidation({ file, isValid: true })
```
jodator, line 44: Using the loader + validator you can extend file:

```typescript
const isValidWASM = await WebAssembly.validate(await file.arrayBuffer())
Object.assign(file, { isValidWASM })
```

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
```diff
@@ -44,6 +64,7 @@ export const RuntimeUpgrade = ({ setRuntime }: RuntimeUpgradeProps) => {
     maxFiles: 1,
     maxSize: MAX_FILE_SIZE,
     multiple: false,
+    getFilesFromEvent: (event) => getFiles(event),
```
jodator, line 67: I tested this with this validator

```ts
const validator = (file: ValidatedFile) => {
  if (!file.isValidWASM) {
    return {
      code: 'file-invalid-type',
      message: 'File is not valid WASM file',
    }
  }

  return null
}
```

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
jodator: Actually, this is to be removed when using a validator.

jodator (changes requested):
I think that some code can be removed so the `fileValidation` variable using `useState()` will not be needed.

DevNikita16:
@jodator Well, you've found very good way of doing this 👏

jodator:
@DevNikita16 I've changed two things:

1.  I've moved pure functions out of the component so they will be created by the JS runtime once not per every component render ([`aff4bb4`](https://github.com/Joystream/pioneer/pull/1108/commits/aff4bb4913fa1298dbfd401f4ebdd344bf9e245b))
2.  I've removed `any` from the callback, which was actually not that easy to fix all the TS errors (acd87d7)

PR #1119, opened 2021-07-26
🚄 Update Joystream node API handling

### LINKED ISSUE #1032: No API connection info
We should display information that the user is not connected to the API.

Right now the user might be confused that some actions are not available.

The other thing that we might to do is to introduce API button that is automatically disabled when no connection to the API (Joystream node).

jodator:
Closes #1032

### DIFF
--- a/packages/ui/src/app/App.tsx
+++ b/packages/ui/src/app/App.tsx
@@ -2,7 +2,10 @@ import React from 'react'
 import { Redirect, Route, Switch } from 'react-router-dom'
 
 import { ConnectionStatus } from '@/common/components/ConnectionStatus'
+import { Loading } from '@/common/components/Loading'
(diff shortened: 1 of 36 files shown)


packages/ui/src/app/App.tsx
```diff
@@ -27,33 +29,45 @@ import { Providers } from './Providers'
...
+      </Page>
+      <GlobalModals />
+    </WaitForAPI>
     <ConnectionStatus />
     <ExtensionWarning />
-    <GlobalModals />
   </Providers>
 )
+
+export const WaitForAPI = ({ children }: { children: ReactNode }) => {
```
DevNikita16, line 65: 👍👍👍

DevNikita16 (approved):

PR #1126, opened 2021-07-27
🧻 Clear specifics after selecting other proposal type

### LINKED ISSUE #1125: Clear specifics when selecting other type of proposal
Problem ex: I want to create proposal "Create WG Lead Opening" in specifics I select WG without lead.

Then I understand that I want to create proposal "Decrease WG lead stake" - so I go back and select this type of proposal. Later I'm again on specifics, and WG without lead(selected in "Create WG Lead Opening") will be selected, but it shouldn't

DevNikita16:
Closes #1125


packages/ui/src/proposals/modals/AddNewProposal/machine.ts
```diff
@@ -206,6 +206,11 @@ export const addNewProposalMachine = createMachine<AddNewProposalContext, AddNew
         },
         SET_TYPE: {
           actions: assign({
+            specifics: (context, event) => {
+              if (context.type !== (event as SetTypeEvent).proposalType) return {}
```
jodator, line 210:
```suggestion
            specifics: (context, event) => {
              if (context.type !== (event as SetTypeEvent).proposalType) {
                return {}
              }
```

jodator: Small code style adjustment

jodator (approved):
👌 nice

Issue #1067, opened 2021-07-22
Forum milestone


jodator:
# Introduction

![](https://user-images.githubusercontent.com/247363/126659373-2782972f-4c68-466c-ac6e-e19e123e38c5.png)

An immutable, auditable, public forum is the main communication and coordination forum among platform members.

*   [Forum chapter](https://joystream.gitbook.io/joystream-handbook/subsystems/forum) in the handbook
*   [Wireframes](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A65528)

# Goals

The forum subsystem will allow users to:

1.  Browse forum (for non-members)
2.  Follow your own threads
3.  Participate in the discussion (for members)

The focus of this milestone is put in allowing non-members and members to create & participate in the discussion. The Forum Working Group Leader actions are put for later implementation.

# Goals details

Each goal has a work scope with links to GH tickets.

*   Tickets with a :purple\_circle: (and "epic" label) are considered to require ~1 week of work).
    *   The rest of the tickets should require a 0.5-1 day of work each.

### Base UI components & integration

*   #1061
*   #1065
*   #1064
*   #1068
*   #1062
*   #1063

### Browse as non-member

*   Overview
    *   #1056
    *   #1057
    *   #1084
    *   #1085
*   Categories & Threads
    *   #1058
    *   #1059
*   Search & archive
    *   #1060
    *   #1081
    *   #1082
    *   #1083

### Browse as a member

*   #1066
*   #1086

### Participate in the discussion

*   Create Thread
    *   #1069
    *   #1072
    *   #1073
*   #1074
*   #1077
*   #1166 

---

[Moved](https://github.com/Joystream/pioneer/issues/1067#issuecomment-886878772) outside the scope of the milestone:

*   #1075 
*   #1122
*   Tags
    *   #1078
    *   #1079
    *   #1080
    *   Create Thread
    *   #1070
    *   #1071
*   #1076
*   #1091 🟣 This might require a considerable effort (defined in #50 )

bedeho:
- How functional would it be to leave out the following for another iteration?
  - polls
  - reactions
  - app resource identifiers
- How much does it impact the estimate?
- If we leave it out, how do we track these so they are not forgotten for later?

jodator:
> How functional would it be to leave out the following for another iteration?

Polls – they are quite big, so the impact might be 1-2 weeks shorter milestone (in total)

[...]

PR #1128, opened 2021-07-27
🗃️ Move `ForumPost` and `ForumComment` to the forum folder

thesan:
I had put these in the wrong place so I prefer moving them before to much work is done on the forum...

See #1067


jodator:
Hmm... that's a bummer since I've also was working on forum posts. However, those are separate entities and might need to have some common interface :thinking:

ATM we need to see what a common interface would be to use the same components, but they have a bit different meaning and some properties are different.

I think that for start they should be renamed to:

*   `ProposalThread`
*   `ProposalPost`

to match forum (and query-node) naming.

thesan:
> Hmm... that's a bummer since I've also was working on forum posts. However, those are separate entities and might need to have some common interface thinking
> 
> ATM we need to see what a common interface would be to use the same components, but they have a bit different meaning and some properties are different.
> 
> I think that for start they should be renamed to:
> 
>     * `ProposalThread`
> 
>     * `ProposalPost`
> 
> 
> to match forum (and query-node) naming.

Sorry I should have done that earlier. Back then I implemented the discussion post based on the [UI/Forum/Comment on figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=1655%3A78281) because that's what was used in the design.

And since `ProposalDiscussionPost` is mostly just missing reaction AFAIK, I thought the type could be the same, but in one case it would have undefined reactions and in the other an array possibly empty...

jodator:
I'm going to merge it after the CI pass :ok\_hand:

Issue #1059, opened 2021-07-22
Thread view


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=4558%3A131866)

Related:
- components

Requirements:
- [x] Open thread by link from the category list or other places
- [x] Display threaded discussion
- [x] Display N number of replies and pagination
- [x] Suggested threads in sidebar
- [x] Post form
- [x] #1192

jodator:
@DevNikita16 For the current implementation (after the #1135):

*   remove the "tag" handling from the thread page
*   the block data should be obtained from the query node as here: https://github.com/Joystream/pioneer/blob/b7dafe9633d7b19ec378580dc7d6e218b93a19a7/packages/ui/src/proposals/types/proposals.ts#L84-L88

jodator:
@DevNikita16 So I'm going to merge the suggested threads PR but since the design was updated we still need to add  (or bring back) to it:

*   date handling
*   breadcrumbs
*   answers count
*   initial post text

p-sad:
Broke up remaining tasks into #1257, #1258

PR #1131, opened 2021-07-27
Thread view init

DevNikita16:
See #1059


jodator:
@DevNikita16 Could you rebase this PR on `main` as the branch was already merged. Without that there will be a lot of noise in the squash commit message.

PR #1135, opened 2021-07-28
✍🏿 Introduce detailed thread types and handle thread header

DevNikita16:
See #1059


packages/ui/package.json
```diff
@@ -15,7 +15,7 @@
     "test": "jest",
     "storybook": "start-storybook -p 6006 -s public --no-manager-cache",
     "build:storybook": "build-storybook -o build-storybook",
-    "generate-mocks": "ts-node dev/scripts/generateMocks.ts",
+    "generate-mocks": "ts-node dev/scripts/generateMocks.ts all",
```
jodator, line 18:
```suggestion
    "generate-mocks": "ts-node dev/scripts/generateMocks.ts",
```

We should be able to use any mocks script ;)

jodator (approved):
Nice 👌

PR #1139, opened 2021-07-28
🥦 Add first step of Create Thread modal

### LINKED ISSUE #1069: Create Thread: General step
@[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A61552)

Requirements:

- [x] Add "Create thread" base modal
  - [ ] #1138
- [ ] Validate topic length (display the length & do not allow more letters to type)
- [x] Description must be a non-empty field

p-sad:
Closes #1069 
I'm not too familiar with xstate so I wouldn't mind a closer review on that, as well as the look of modal in general

### DIFF
--- a/packages/ui/src/forum/modals/CreateThreadModal/CreateThreadDetailsModal.tsx
+++ b/packages/ui/src/forum/modals/CreateThreadModal/CreateThreadDetailsModal.tsx
@@ -1,0 +1,60 @@
+import React, { useMemo } from 'react'
+
+import { ButtonPrimary, ButtonsGroup } from '@/common/components/buttons'
+import { CKEditor } from '@/common/components/CKEditor'
+import { InputComponent, InputText } from '@/common/components/forms'
+import { Arrow } from '@/common/components/icons'
+import { Modal, ModalBody, ModalFooter, ModalHeader } from '@/common/components/Modal'
+import { RowGapBlock } from '@/common/components/page/PageContent'
+import { TextMedium } from '@/common/components/typography'
+import { useModal } from '@/common/hooks/useModal'
+
+import { GeneralFormContext } from './machine'
+
+interface Props {
+  context: GeneralFormContext
+  setTopic: (t: string) => void
+  setDescription: (d: string) => void
+  onSubmit: () => void
+}
+
+export const CreateThreadDetailsModal = ({
+  context: { topic, description },
+  setTopic,
+  setDescription,
+  onSubmit,
+}: Props) => {
+  const fieldMissing = useMemo(() => !topic && !description, [topic, description])
+  const { hideModal } = useModal()
+  return (
+    <Modal onClose={hideModal} modalSize="l">
+      <ModalHeader title="Create a thread" onClick={hideModal} />
+      <ModalBody>
+        <RowGapBlock gap={24}>
+          <h1>General</h1>
+          <RowGapBlock gap={8}>
+            <TextMedium light>Please make sure your title will be clear for users</TextMedium>
+            <InputComponent label="Topic of the thread" id="field-topic">
+              <InputText id="field-topic" value={topic} onChange={(event) => setTopic(event.target.value)} />
+            </InputComponent>
+            <InputComponent label="Description" required inputSize="auto" id="field-description">
+              <CKEditor
+                id="field-description"
+                onChange={(_, editor) => setDescription(editor.getData())}
+                onReady={(editor) => editor.setData(description)}
+              />
+            </InputComponent>
+          </RowGapBlock>
+        </RowGapBlock>
+      </ModalBody>
+      <ModalFooter>
+        <ButtonsGroup align="right">
+          <ButtonPrimary onClick={onSubmit} size="medium" disabled={fieldMissing}>
+            Next step
+            <Arrow direction="right" />
+          </ButtonPrimary>
+        </ButtonsGroup>
+      </ModalFooter>
+    </Modal>
+  )
+}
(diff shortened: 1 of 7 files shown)


packages/ui/src/forum/modals/CreateThreadModal/machine.ts
```diff
@@ -0,0 +1,61 @@
+import { createMachine, assign } from 'xstate'
+
+import { EmptyObject } from '@/common/types'
+
+export interface GeneralFormContext {
+  topic: string
+  description: string
```
jodator, line 7: For the initial state, the context should have everything optional (unless you'll have data to set initially).

Without that, the `Requred<GeneralFormContext>` will do nothing.

jodator: Oh, I see below that there's a `Partial<>` as well. I think that we can reducu the number of interfaces to have only two (initial, and later the Transaction one).

packages/ui/src/forum/modals/CreateThreadModal/machine.ts
```diff
@@ -0,0 +1,61 @@
...
+  | { type: 'SET_TOPIC'; topic: string }
+  | { type: 'SET_DESCRIPTION'; description: string }
+
+export const createThreadMachine = createMachine<CreateThreadContext, CreateThreadEvent, CreateThreadState>({
+  initial: 'requirementsVerification',
+  context: {
+    topic: '',
+    description: '',
+  },
+  states: {
```
jodator, line 33: This can be ommited when using optional parameters in the context.

packages/ui/src/forum/modals/CreateThreadModal/machine.ts
```diff
@@ -0,0 +1,61 @@
...
+        NEXT: 'generalDetails',
+        FAIL: 'requirementsFailed',
+      },
+    },
+    requirementsFailed: { type: 'final' },
+    generalDetails: {
+      on: {
+        NEXT: {
+          target: 'end',
+          cond: (context) => !!(context['topic'] && context['description']),
```
jodator, line 45:
```suggestion
          cond: (context) => !!(context.topic && context.description),
```

Usually we use the normal accessor for that and we use array access when it is not a valid JS name, like `context['some-foo']`.

packages/ui/src/forum/modals/CreateThreadModal/CreateThreadModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { createThreadMachine } from './machine'
+
+export const CreateThreadModal = () => {
+  const { active: member } = useMyMemberships()
+  const { hideModal, showModal } = useModal()
+  const [state, send] = useMachine(createThreadMachine)
+
+  useEffect(() => {
+    if (state.matches('requirementsVerification')) {
+      !member ? showModal<SwitchMemberModalCall>({ modal: 'SwitchMember' }) : send('NEXT')
```
jodator, line 19: Please use if-else if there's an action. That's because, in most cases, ternary operator would be used in assignement.

packages/ui/src/forum/modals/CreateThreadModal/CreateThreadDetailsModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+  onSubmit: () => void
+}
+
+export const CreateThreadDetailsModal = ({
+  context: { topic, description },
+  setTopic,
+  setDescription,
+  onSubmit,
+}: Props) => {
+  const fieldMissing = useMemo(() => !(topic && description), [topic, description])
```
jodator, line 27: I'm really acustomed to `boolean` variables starting with `is`, `has`, `are`, etc. It is then easier to parse.

So here it would be either, standard `isValid` -> will show that this is some sort of "validation" as well.

jodator (approved):
Generally, the use of machine is OK but there are too many types for the context IMO. There can be just one or two (you can pass the general context values serparetely as well, instead of an object).

jodator (changes requested):
Sorry, wrong button before ;)

packages/ui/src/forum/modals/CreateThreadModal/machine.ts
p-sad: I wasn't sure about reducing the number of interfaces because I wanted them to have a separate interface with a meaningful name for each step. I'll see if that'll be needed once I've got the transaction step done, and I'll clean them up.

packages/ui/src/forum/modals/CreateThreadModal/machine.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type TransactionContext = Required<GeneralFormContext>
+
+type CreateThreadContext = Partial<TransactionContext>
+
+type CreateThreadState =
+  | { value: 'requirementsVerification'; context: EmptyObject }
+  | { value: 'requirementsFailed'; context: EmptyObject }
+  | { value: 'generalDetails'; context: GeneralFormContext }
+  | { value: 'end'; context: TransactionContext }
```
jodator, line 18: One last thing, I think that one or two of those types are redundant here:

```suggestion
export interface CreateThreadContext {
  topic?: string
  description?: string
}

type CreateThreadState =
  | { value: 'requirementsVerification'; context: EmptyObject }
  | { value: 'requirementsFailed'; context: EmptyObject }
  | { value: 'generalDetails'; context: CreateThreadContext }
  | { value: 'end'; context: Required<CreateThreadContext> }
```

The `Required<CreateThreadContext>` can be brought back as well, but since it will be a very small machine we can leave that as inline type.

jodator (changes requested):
Jut to make my comment about redundant types a bit more clear – see my comment below.

packages/ui/src/forum/modals/CreateThreadModal/machine.ts
jodator: OH, I've missed your comment. Sorry for that.

I think clearing them now might be a better option anyway to have clean code and not forget about that in the future ;)

jodator (approved):

PR #1141, opened 2021-07-29
🅱️ Handle multiple files/errors validation

### LINKED ISSUE #1112: Handle multiple file validation errors
There might be multiple file validation errors in the `RuntimeUpgrade` step, merging the looks like this:

We could either display them as `... was not loaded because of: "error o", "error 2", "error 3".` or just pick the first.

ClumsyVlad:
Closes #1112


jodator (approved):
Already reviewed F2F 👍

PR #1145, opened 2021-07-29
📋 Handle thread posts with pagination and replies

DevNikita16:
See #1059


packages/ui/dev/scripts/generators/forum/generateForumThreads.ts
```diff
@@ -4,20 +4,29 @@ import { RawForumCategoryMock, RawForumPostMock, RawForumThreadMock } from '@/mo
...
-  let nextId = 0
-
   const forumThreads = forumCategories
     .map(({ id }) => {
       return [...new Array(randomFromRange(3, 10))].map(() => ({
-        id: String(nextId++),
+        id: String(nextThreadId++),
         categoryId: id,
-        isSticky: !(nextId % 5),
+        isSticky: faker.datatype.boolean(),
```
jodator, line 29: Will this be 50%? To have some _real-like_ data I think that sticky threads should be 1-4 per category. That's why it was modulo 5 before.

packages/ui/src/app/pages/Forum/ForumThread.tsx
```diff
@@ -79,13 +78,9 @@ export const ForumThread = () => {
...
     return (
-      <div>
-        {posts.map((post) => (
-          <div>
-            {post.id} | {post.text} | by {post.authorId}
-          </div>
-        ))}
-      </div>
+      <MainPanel ref={sideNeighborRef}>
+        <PostList threadId={thread?.id as string} />
```
jodator, line 82: Casting like this should be done only as a last resort of enforcing types. It is better to avoid such cast to not introduce bugs. Here it will create an invalid cast when `thread` is `undefined`. This should be either handled here or on the `PostList` component. I'd go with rendering this component here only if thread is not undefined.

packages/ui/src/common/components/BlockTime/BlockTime.tsx
```diff
@@ -74,19 +76,25 @@ const Wrapper = styled.div<BlockTimeLayoutProps>`
...
+          text-align: end;
           grid-row-gap: 8px;
+
           ${AboutText} {
             font-size: 12px;
             line-height: 18px;
             color: ${Colors.Black[500]};
             order: 1;
           }
+
```
jodator, line 89: :heart\_eyes: yeah, this reads better for me too

packages/ui/src/forum/components/PostList/PostListItem.tsx
```diff
@@ -13,55 +14,54 @@ import { spacing } from '@/common/utils/styles'
...
+        <BlockTime block={createdAtBlock} layout="reverse" />
+      </ForumPostRow>
       <MessageBody>
         {repliesTo && (
           <Reply>
             <ReplyBadge>
               <ArrowReplyIcon />{' '}
               <Badge>
-                <Link to={repliesTo.link}>Replies to {repliesTo?.author?.handle}</Link>
+                <Link to={window.location.href}>Replies to {repliesTo?.author?.handle}</Link>
```
jodator, line 38: FYI: This will link to the whole thread but it should link to the specific post. I know that this will be handled in another issue so for now it is OK.

jodator (changes requested):
Looks nice. I'd change few things, though – please see comments.

packages/ui/src/app/pages/Forum/ForumThread.tsx
DevNikita16: Well, I've included thread condition in addition to isLoading, but in general it couldn't be undefined here - if no thread then we redirect to 404

DevNikita16:
Sorry for rebase. Forgot that it breaks PR history.

jodator (approved):
Thanks! 👍

PR #1150, opened 2021-07-30
👨🏼‍💻 Upgrade CategoryListItem styles + tooltip timeout

ClumsyVlad:


ClumsyVlad:
I added a timeout for the tooltip so it doesn't render if user hover over  trigger for a moment. This also fixes an issue with tooltip centering if trigger is resized.

PR #1152, opened 2021-07-30
🪱 Add transaction step to Create Thread Modal

### LINKED ISSUE #1073: Create Thread: Sign & send transaction
@[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A65892)

Requirements:

- [ ] Prepare & sign the create thread transaction
- [ ] Handle success step (display link to thread)
- [ ] Handle error step

p-sad:
I didn't manage to create a storybook entry for it because it would involve mocking the `useSignAndSendTransaction` hook which I'm not sure how to do, or putting it outside the sign modal which would result in a large blob of conditionals and effects. So not a single person in the world knows what it looks like but it's tested so we know it works

Closes #1073


jodator (approved):
Just one follow-up actually – the other things looks OK :)

PR #1157, opened 2021-08-02
🥚 Add Forum breadcrumbs

### LINKED ISSUE #1138: Display forum breadcrumbs in Create Thread modal
[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A61552)
The breadcrumbs probably need to be passed to the modal in a modal-with-data call.

p-sad:
Closes #1138


packages/ui/test/forum/hooks/useForumCategory.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+describe('useForumCategory', () => {
+  const mockServer = setupMockServer({ noCleanupAfterEach: true })
+
+  beforeAll(() => {
+    mockCategories.map((category) => seedForumCategory(category, mockServer.server))
+  })
+
+  it('Basic details', async () => {
```
jodator, line 28:
```suggestion
  it('Category properties', async () => {
```

packages/ui/test/forum/hooks/useForumCategory.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const { forumCategory } = await renderUseForumCategory('0')
+    expect(forumCategory?.title).toBe('Category 0')
+    expect(forumCategory?.description).toBe('Root category')
+    expect(forumCategory?.id).toBe('0')
+    expect(forumCategory?.parentCategories).toEqual([])
+  })
+
+  describe('Parent categories', () => {
+    it('Query depth exceeds parents depth', async () => {
```
jodator, line 37:
```suggestion
    it('One parent', async () => {
```

jodator: Query depth is an implementation detail IMO, so just call that in a simpler way.

packages/ui/test/forum/hooks/useForumCategory.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+      const { forumCategory } = await renderUseForumCategory('1')
+      expect(forumCategory?.parentCategories.length).toEqual(1)
+      expect(forumCategory?.parentCategories[0].id).toEqual('0')
+    })
+
+    it('Parents depth exceeds query depth', async () => {
```
jodator, line 43:
```suggestion
    it('3 parents', async () => {
```

? not 100% sure

packages/ui/test/forum/hooks/useForumCategory.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+      const { forumCategory } = await renderUseForumCategory('4')
+      expect(forumCategory?.parentCategories.length).toEqual(3)
+    })
+  })
+
+  it('Subcategories list', async () => {
```
jodator, line 49: I'm not sure what this check should do?

I think that it was "parent categories order" so the test name should reflect that.

packages/ui/src/forum/queries/forum.graphql
```diff
@@ -4,6 +4,27 @@ fragment ForumCategoryFields on ForumCategory {
...
 }
 
+fragment ForumSubCategoryFields on ForumCategory {
+  id
+  title
+}
+
+fragment ForumCategoryDetailedFields on ForumCategory {
+  ...ForumCategoryFields
+  parent {
```
jodator, line 14: So, [the runtime depth](https://github.com/Joystream/joystream/blob/40962ee609326d964b3985f68ed400754729c888/runtime/src/lib.rs#L619) is `6`. However, I wonder if we should do this query with every request for the category (to double-check with the wireframes and where we need breadcrumbs).

The question here is whether to query like that or try to take advantage of the Apollo query caching. Apollo client will cache the categories that were previously fetched IIRC it will cache based on the query/response fields.

So if we have a category tree `1 -> 2 -> 3 -> 4 -> 5` we could fetch one by one in the reverse order ( `5` then `4` then `3`, etc). This might be challenging using react hooks as we can't use hooks in loops.

A simple solution would be a multi-category query, that expands while getting more parents:

*   query categories `5` then
*   query categories `5,4` then
*   query categories `5,4,3` then
*   ... up to category without a parent, so it would stop with: `5,4,3,2,1`

However, this might re-query those for every category, as the starting query will always query the initial category (`5`). So, we'd need to investigate how [Apollo's cache works](https://www.apollographql.com/docs/react/caching/overview/). To see if my assumptions are correct or not.

jodator: So for this PR I'd go with a simpler route:

*   create a dedicated `useForumBreadcrumbs()` hook
*   it will query for `GetCategoryBreadcrumbs` query (let's name it so no one would re-use it for other purposes)
*   The fields used in this query can be referenced as `ForumCategoryBreadcrumbFields` or not at all in this dedicated query:

```
query GetForumCategoryBreadcrumbs($id: ID!) {
  forumCategories(where:{id_eq: $id}) {
    id
    title
    parent {
      id
      title
      parent {
        id
        title
        parent {
          ... // up to 6th for now
        }
      }      
    }
  }
}
```

jodator: This will create a proper abstraction over how we'll fetch this data. This hook will be used with/inside the `<ForumBreadcrumbs />` and to generate breadcrumbs at the top of the forum pages as well.

packages/ui/src/forum/types/ForumCategory.ts
```diff
@@ -16,3 +20,33 @@ export const asForumCategory = (fields: ForumCategoryFieldsFragment): ForumCateg
   description: fields.description,
   subcategories: [],
 })
+
+const asSubCategory = (fields: ForumSubCategoryFieldsFragment): ForumSubCategory => ({
```
jodator, line 24: Let's call it `ForumCategoryBreadcrumb` and `asForumCategoryBreadcrumb()` (I think that dropping `Category` will be also OK: `ForumBreadcrumb` should be understandable and will directly tie to the usage. The problem with `ForumSubCategory` for me is that a sub-category is every category nested inside another and is tied (at least for me) to the whole: category, threads, own page, etc.

jodator (changes requested):
> * display e.g. 3 at most, and some kind of "etc" icon or ellipses in front of them
> 
> * find out what the [`MAX_CATEGORY_TREE_DEPTH`](https://joystream.gitbook.io/joystream-handbook/subsystems/forum#constants) is and write the query accordingly

So I think that we can start with a query that _looks_ unoptimized (6 categories). A proper solution might require digging into Apollo's cache and, at least for now, we could live with that.

I've requested some naming changes to make sure that anyone looking at the related code will know that this relates to breadcrumbs.

packages/ui/src/forum/queries/forum.graphql
jodator: I've created an issue for the performance review: #1158. I think that doing that after we have some real-world data might be better.

packages/ui/test/forum/hooks/useForumCategory.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { result, waitForNextUpdate } = renderHook(() => useForumBreadcrumbs(id), {
+    wrapper: ({ children }) => <MockQueryNodeProviders>{children}</MockQueryNodeProviders>,
+  })
+  while (result.current.isLoading) {
+    await waitForNextUpdate()
+  }
+  return result.current
+}
+
+describe('useForumBreadcrumbs', () => {
```
jodator, line 21: The file name should match the tested component.

packages/ui/src/forum/components/ForumBreadcrumbs.tsx
```diff
@@ -0,0 +1,29 @@
...
+interface Props {
+  forumBreadcrumbs: ForumBreadcrumb[]
+  currentBreadcrumb: string
+}
+
+export const ForumBreadcrumbs = React.memo(({ forumBreadcrumbs, currentBreadcrumb }: Props) => {
+  const crumbs = forumBreadcrumbs.map((crumb) => ({
+    path: `/forum/categories/${crumb.id}`,
+    breadcrumb: crumb.title,
+    key: crumb.id + crumb.title,
```
jodator, line 17: I'm not sure if that's needed for key. The `id` must be unique in the Query Node so I'd use `id` only.

packages/ui/src/forum/components/ForumBreadcrumbs.tsx
```diff
@@ -0,0 +1,29 @@
...
+  }))
+  crumbs.push({ path: '', breadcrumb: currentBreadcrumb, key: 'last:' + currentBreadcrumb })
+  return (
+    <BreadcrumbsListComponent>
+      {crumbs.map(({ path, breadcrumb, key }, index, { length }) => (
+        <BreadcrumbsItem key={key} url={path} isLink={index < length - 1}>
+          {breadcrumb}
+        </BreadcrumbsItem>
+      ))}
+    </BreadcrumbsListComponent>
```
jodator, line 27: Since the not link is always pushed at the end, maybe something like below:

```suggestion
    <BreadcrumbsListComponent>
      {forumBreadcrumbs.map(({ id, title }) => (
        <BreadcrumbsItem key={id} url={`/forum/categories/${crumb.id}`} isLink>
          {title}
        </BreadcrumbsItem>
      ))}
      <BreadcrumbsItem key="_current">
        {currentBreadcrumb}
      </BreadcrumbsItem>
    </BreadcrumbsListComponent>
```

The above was written on GH so it might need some adjustments. The `<BeradcrumbsItem>` should allow empty `url` when `isLink` is `undefined` (so not set).

jodator: The other benefit here would be one loop vs two loops for rendering this component.

packages/ui/src/forum/components/ForumBreadcrumbs.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { BreadcrumbsListComponent } from '@/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsList'
+
+import { ForumBreadcrumb } from '../types'
+
+interface Props {
+  forumBreadcrumbs: ForumBreadcrumb[]
+  currentBreadcrumb: string
+}
+
+export const ForumBreadcrumbs = React.memo(({ forumBreadcrumbs, currentBreadcrumb }: Props) => {
```
jodator, line 13: :+1: for `React.memo()`

jodator: And a simply storybook is also needed to help style this component later.

packages/ui/src/forum/types/ForumCategory.ts
```diff
@@ -13,15 +17,37 @@ export interface ForumModerator {
...
+const asSubCategory = (fields: ForumSubCategoryFieldsFragment): ForumBreadcrumb => ({
+  id: fields.id,
+  title: fields.title,
+})
+
+export const asForumBreadcrumbs = (fields: ForumCategoryBreadcrumbsFieldsFragment): ForumBreadcrumb[] => {
+  const breadcrumbs: ForumBreadcrumb[] = []
+  assignBreadcrumbs(fields, breadcrumbs)
+  return breadcrumbs
+}
```
jodator, line 44: This looks like something that could be avoided as `assingBreadcrumbs()` is not "in the spirit of functional programming" (IMO). It does side-effect to the passed params.

This could be written using a loop or `Array.reduce()` since it is a linear structure.

```typescript
return fields.reduce((acc, item) => {
  if (item.parent) {
    acc.push(asSubCategory(item.parent)) //.unshift()?
  }
  return acc
}, [] )
```

jodator (changes requested):
I've added some tips on simplifying the code (IMO).

packages/ui/src/forum/components/ForumBreadcrumbs.tsx
p-sad: I decided to make `currentBreadcrumb` optional, to make it work more like the breadcrumbs component we already have

packages/ui/src/forum/types/ForumCategory.ts
p-sad: I decided to still use a recursive function (that no longer pushes elements to an external array) as it works very well for inserting elements in a reverse order and the depth is predetermined by query depth

Issue #1042, opened 2021-07-22
Make download file with loader/progress bar


jodator:
We should check first what are the times for larger files:

- how big a runtime upgrade can be 10 MB, 1 GB?
- can apollo client return progress on bigger fetch?

jodator:
So we can't display the progress bar so the only option is to show some sort of  "Lodaing...".

Other things from the design:

*   remove the modal
*   make the button show "Fetching file..." when the query is loading

PR #1159, opened 2021-08-03
🦔 Change runtime download to inline component

p-sad:
See #1042 
Doesn't open a modal anymore, though it could use a fancy spinner still


jodator:
I'll do a F2F review with @p-sad as I want to explore `useLazyQuery()` for this component.

jodator (approved):

PR #1163, opened 2021-08-03
🗂️ Implement the Forum category page

### LINKED ISSUE #1056: Forum categories
@[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=27%3A7429)

Related:
 - Components: 

Requirements:
- [ ] Category tab of the Forum page
- [ ] Fetch data from query-node

thesan:
Closes #1056


thesan:
I originally planed to keep `<CategoryListItem />` stateless but all the queries would have triggered too many updates, so I had to change that.

Also the "Popular Thread" is currently the last updated one but I don't think it's the intended behavior

packages/ui/src/mocks/data/seedForum.ts
```diff
@@ -36,9 +37,13 @@ export interface RawForumPostMock {
 export function seedForumCategory(forumCategoryData: RawForumCategoryMock, server: any) {
   return server.schema.create('ForumCategory', {
     ...forumCategoryData,
+    moderators: seedModerators(forumCategoryData.moderators, server),
```
jodator, line 40: That shouldn't be necessary, I'll check how to handle that.

packages/ui/dev/scripts/generators/forum/generateCategories.ts
```diff
@@ -21,6 +26,7 @@ const generateCategory = (depth: number, mocks: RawForumCategoryMock[], parentId
     title: faker.lorem.words(randomFromRange(3, 5)),
     description: faker.lorem.paragraph(randomFromRange(2, 3)),
     parentId: parentId ? parentId : null,
+    moderators: faker.random.arrayElements(workers, randomFromRange(1, 20)).map(({ id }) => id),
```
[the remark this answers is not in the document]
jodator: I've updated the field name to `moderatorIds` as well. This will make seeding a bit simpler.

packages/ui/src/mocks/data/seedForum.ts
jodator: OK, so for the one-to-many associations you can use ids only. I had to check how it is stored in MirageJS since `s` use there could go both ways. The proper one here is `moderatorIds`.

packages/ui/src/mocks/data/seedForum.ts
```diff
@@ -36,9 +37,13 @@ export interface RawForumPostMock {
...
   })
 }
 
+const seedModerators = (moderatorsIds: string[], server: any) =>
+  moderatorsIds.map((workerId) => server.schema.find('Worker', workerId))
+
```
jodator, line 46: Can be removed when using `moderatorIds`.

packages/ui/src/forum/hooks/useForumCategoryDetails.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+interface CategoryDetails {
+  subcategories?: ForumSubCategory[]
+  threadCount?: number
+  topThread?: ForumThread & { postCount?: number }
+  latestPost?: ForumPost
+}
+
+export const useForumCategoryDetails = (categoryId: string): CategoryDetails => {
```
jodator, line 31: I see that there's a lot going on which is hard for me to process as some fetch/query dependencies are intervened. This makes me think that it might be better to split this hook and rewrite dependent components a bit.

*   this hook should fetch a minimal set of "core" data
*   A non-core data (that can be fetched independently) are
    *   total threads
    *   last post
    *   popular thread
    *   moderators?

I think that the `ForumCatregoryListItem` should just pass `forumCategoryId` to the other components.

```tsx
const ForumCategoryListItem = (category) => {
  const { threadCount } = useThreadCound(category.id)
  return (
    <>
      // ... Other components like title, subcategories, etc
      <ThreadCount count={threadCount} /> // this might be fetched here
      <LastPost forumCategoryId={category.id} />
      <TopThread forumCategoryId={category.id} />
    </>
  )
}
```

:point\_up:  The above is just a general idea, not the real code to be implemented.

jodator: In here the `<LastPost />` should fetch its own data using `useLastPost()` hook.

jodator: Additionally, I'm worried that every time the optional data is fetched, and the `details` is updated the whole hook make the component using it to re-render while it could be limited to the component that uses that data. That's a minor issue, though as the child components are using memo anyway, but still, a thing to keep in mind.

ps.: The above claim might not be 100% accurate. I don't recall now if the main component will be re-rendered when children are rendered. :thinking:

jodator (changes requested):
Thanks, @thesan!

The components look great. I have some requests, though.

- I'd split that hooks & components dependancy
- The  seeds code could be simplified
- Tabs behavior was suprpising, hence the suggestion to point to forum

packages/ui/src/forum/hooks/useForumCategoryDetails.ts
thesan: I think the previous code was indeed triggering more unnecessary updates. In any case these changes do make thing more readable :+1:

packages/ui/src/app/pages/Forum/components/ForumTabs.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return <Tabs tabs={tabs} />
+}
+
+export const ForumForumTabs = () => {
+  const tabs = usePageTabs([
+    ['Categories', '/forum'],
+    ['Latest threads ', '/#'],
+    ['Top threads', '/#'],
+    ['Tags', '/#'],
+    ['My threads', '/#'],
```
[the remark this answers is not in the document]
thesan: I used '/forum/#' to prevent every tabs from being active

jodator (approved):

PR #1164, opened 2021-08-04
Handle suggested threads and post form

DevNikita16:
See #1059


packages/ui/src/forum/components/threads/ThreadItem.tsx
```diff
@@ -7,56 +7,61 @@ import { Arrow } from '@/common/components/icons'
...
 import { ThreadItemContent } from './ThreadItemContent'
 import { ThreadsLayoutSpacing } from './ThreadsLayout'
 
 export interface ThreadItemProps {
-  label: string
-  count?: number
+  categoryLabel: string
+  categoryCount?: number
+  thread: ForumThread
+  withButtons?: boolean
```
jodator, line 19: I think that it makes API more convenient when using false as the default value for optional parameters.

Using default `true` requires more typing:

```typescript
interface Props {
  withButtons?: boolean // default true
}

// Hide buttons:
<ThreadItem withButtons={false} />

// Show buttons:
<ThreadItem />
```

Using default `false` tends to be more fluent:

```typescript
interface Props {
  withButtons?: boolean // default false
}

// Hide buttons:
<ThreadItem />

// Show buttons:
<ThreadItem withButtons />
```

Using this for React components also makes them similar to HTML attributes:

```html
<button disabled />
```

packages/ui/src/forum/components/threads/ThreadItem.stories.tsx
```diff
@@ -15,15 +19,17 @@ const Template: Story<ThreadItemProps> = (args) => (
...
   </ScrollBlock>
 )
 
 export const ThreadItemComponent = Template.bind({})
 
+const thread = rawThreads[randomFromRange(0, rawThreads.length - 1)]
 ThreadItemComponent.args = {
-  label: 'CategoryLabel',
-  count: 10,
+  categoryLabel: 'Lorem > Ipsum > Del',
```
jodator, line 32: Maybe not for now, but this should be a ForumBreadcrumb component.

jodator (changes requested):
Just one remark on the boolean props usage for components to be fixed now, other comments are a general things that should be done sooner than later (the base components uses mocks but we should start using real data in the implementation). Other than that the PR look good 👍 

There's room for improvement (in follow-up PRs):
- the "lorem > ipsum" should display the category breadcrumb
- there should be no tags ("badge 1", "badge 2"),
- the time should be real
- the answers count should be real

jodator (approved):

jodator:
**If CI go green, I'm going to merge it. I've left some comments here:** #1059

Issue #1077, opened 2021-07-22
Delete own post


jodator:
Requirements:

- [x] Display sign and send [transaction](https://joystream.gitbook.io/joystream-handbook/subsystems/forum#delete-post)
- [x] only allow own posts to be deleted (disable button for matching member id)
- [x] only allow deleting posts on non-archived threads/categories

jodator:
The button should be added as context menu: https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=4558%3A131866

PR #1181, opened 2021-08-06
🗑️ Add a "Delete post" option

p-sad:
See #1077 

The way the extrinsic works seems a bit more complicated than it looked at a first glance. I decided to fetch the details for it in the modal to avoid introducing too much prop drilling.
The transaction takes a `rationale` argument for which I'm using an empty string for now, though there might be a need for a details step of the modal. There is also a `hide` argument which I'm guessing in this particular use case should always be set to `true`, given [the way posts work in the runtime](https://joystream.gitbook.io/joystream-handbook/subsystems/forum#editable-non-editable). It also may mean there needs to be an option to finalize a post/disable edits sometime down the line.

Following follow-ups may apply:
* Disallow editing posts in locked/archived threads
* Check post status (whether it's editable or has been deleted from runtime)
* Optionally add a details modal for the rationale
* Add a required balance check to requirements checks


jodator:
> The way the extrinsic works seems a bit more complicated than it looked at a first glance. I decided to fetch the details for it in the modal to avoid introducing too much prop drilling.  
> The transaction takes a `rationale` argument for which I'm using an empty string for now, though there might be a need for a details step of the modal. There is also a `hide` argument which I'm guessing in this particular use case should always be set to `true`, given [the way posts work in the runtime](https://joystream.gitbook.io/joystream-handbook/subsystems/forum#editable-non-editable). It also may mean there needs to be an option to finalize a post/disable edits sometime down the line.

That's a very good summary for the changes :rocket:

jodator:
> Following follow-ups may apply:
> 
> \* Disallow editing posts in locked/archived threads
> 
> \* Check post status (whether it's editable or has been deleted from runtime)
> 
> \* Optionally add a details modal for the rationale

Yes, please create follow-ups or just add them as points to handle in #1077

packages/ui/src/forum/components/PostList/PostListItem.tsx
```diff
@@ -60,6 +65,16 @@ export const PostListItem = forwardRef<HTMLDivElement, PostProps>(({ post, isSel
           <Button square disabled={isPreview}>
             <ReplyIcon />
           </Button>
+          {isOwn && (
+            <ContextMenu
```
jodator, line 69: Could you check (and add if missing) handling for a smaller context item? I think it should allow passing `size` prop similar to button sizing.

packages/ui/src/forum/modals/DeletePostModal/DeletePostModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+      send('PASS')
+    }
+  }, [threadId, categoryId, active?.id, api])
+
+  if (state.matches('requirementsVerification')) {
+    return <WaitModal title="Please wait..." description="Checking requirements" onClose={hideModal} />
+  }
+
+  if (state.matches('transaction') && api && threadId && categoryId) {
+    const transaction = api.tx.forum.deletePosts(post.author.id, [[categoryId, threadId, post.id, true]], '')
```
jodator, line 48: This fails on the page, it looks like some types must be stricter for some extrinsics.

```suggestion
    const transaction = api.tx.forum.deletePosts(
      post.author.id,
      [[parseInt(categoryId), parseInt(threadId), parseInt(post.id), true]],
      ''
    )
```

The above worked for me :point\_up: So I was able to send the transaction. However, it might be better to experiment with `createType()` from polkadot's API.

Something like this works: `api.createType('PostId', 23)` but I think that I was using `createType()` helper somewhere.

packages/ui/src/forum/modals/DeletePostModal/DeletePostSignModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <SelectedAccount account={controllerAccount} />
+          </InputComponent>
+        </RowGapBlock>
+      </ModalBody>
+      <ModalFooter>
+        <TransactionInfoContainer>
+          <TransactionInfo
+            title="Transaction fee:"
+            value={paymentInfo?.partialFee.toBn()}
+            tooltipText={'Lorem ipsum dolor sit amet consectetur, adipisicing elit.'}
```
jodator, line 61:
```suggestion
            tooltipText="Lorem ipsum dolor sit amet consectetur, adipisicing elit."
```

Just my OCD here :)

jodator (changes requested):
Looks good 👍 One thing to fix is the types uses in the real app, not in the tests only :)

The `parseInt()` is good enough, but `createType()` might be more future-proof.

packages/ui/src/memberships/hooks/usePostParents.ts
```diff
@@ -0,0 +1,20 @@
...
+  const { api } = useApi()
+  const { threadId, categoryId } = useMemo(() => asPostParents(data?.forumPostByUniqueInput, api), [data, loading])
+  return { threadId, categoryId, parentsLoading: loading }
+}
+
+const asPostParents = (fields: ForumPostParentsFragment | null | undefined, api: ApiRx | undefined) =>
+  fields && api
+    ? {
+        threadId: api.createType('ThreadId', fields.thread.id),
+        categoryId: api.createType('CategoryId', fields.thread.category.id),
```
p-sad, line 18: This can be done using `ApiRx` or the helper from `@joystream/types`, though the helper can only accept numeric values (rather than a string):
```suggestion
        threadId: createType('ThreadId', Number.parseInt(fields.thread.id)),
        categoryId: createType('CategoryId', Number.parseInt(fields.thread.category.id)),
```
I guess it's a matter of preference

jodator: Yeah, that doesn't matter TBH. The thing I'm worried here, that this hook might be too specialised and/or it mixes Query-Node types with internal and API types. Usually, we will do them in two steps:

1. In hooks fetching data we translate external types from `...Fields` to our internal types.
2. When touching API we would t[ranslate JS types to Substrate API types](https://github.com/Joystream/pioneer/blob/9a7f4a09b9c4fb41f5105b04eb57219380ac228d/packages/ui/src/memberships/modals/utils.ts#L11)

(we however break that for BN as BN is used to store token values).

jodator (commented):
I'd change that one bit, but OTOH it might stay.

Please check the test and merge when CI will be green

jodator (approved):

Issue #1058, opened 2021-07-22
Forum Category page


jodator:
@[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A52493)

Display the forum category threads:

- [x] Fetch threads for given category
- [ ] Display pinned items on top
- [x] Display liste of sub-category

Issue #1118, opened 2021-07-26
Members search is gone


jodator:
The search bar on the members page is gone.

PR #1188, opened 2021-08-06
🕴️ Filter the Forum Categories result by parent id

### LINKED ISSUE #1169: Forum categories filter
- [ ] Use the search bar to filter categories
- [ ]  Display the number of results on the "categories" tab

thesan:
Closes #1169
See #1058

### DIFF
--- a/packages/ui/src/mocks/server.ts
+++ b/packages/ui/src/mocks/server.ts
@@ -52,8 +52,8 @@ export const fixAssociations = (server: Server<AnyRegistry>) => {
 
   const forumCategoryModel = schema.modelFor('forumCategory')
   // "Mirage: The forum-category model has multiple possible inverse associations for the forum-category.parent association."
-  forumCategoryModel.class.prototype.associations.parent.opts.inverse = 'forumcategoryparent'
-  forumCategoryModel.class.prototype.associations.forumcategoryparent.opts.inverse = 'parent'
+  forumCategoryModel.class.prototype.associations.parent.opts.inverse = null
+  forumCategoryModel.class.prototype.associations.forumcategoryparent.opts.inverse = null
 }
 
 export const makeServer = (environment = 'development') => {
--- a/packages/ui/src/mocks/server.ts
+++ b/packages/ui/src/mocks/server.ts
@@ -73,7 +73,7 @@ export const makeServer = (environment = 'development') => {
               appliedOnOpeningEvents: getWhereResolver('AppliedOnOpeningEvent'),
               budgetSetEvents: getWhereResolver('BudgetSetEvent'),
               budgetSpendingEvents: getWhereResolver('BudgetSpendingEvent'),
-              forumCategories: getWhereResolver('ForumCategory'),
+              forumCategoriesConnection: getConnectionResolver('ForumCategoryConnection'),
               forumCategoryByUniqueInput: getUniqueResolver('ForumCategory'),
               forumThreads: getWhereResolver('ForumThread'),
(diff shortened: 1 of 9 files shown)


packages/ui/src/mocks/server.ts
```diff
@@ -52,8 +52,8 @@ export const fixAssociations = (server: Server<AnyRegistry>) => {
 
   const forumCategoryModel = schema.modelFor('forumCategory')
   // "Mirage: The forum-category model has multiple possible inverse associations for the forum-category.parent association."
-  forumCategoryModel.class.prototype.associations.parent.opts.inverse = 'forumcategoryparent'
-  forumCategoryModel.class.prototype.associations.forumcategoryparent.opts.inverse = 'parent'
+  forumCategoryModel.class.prototype.associations.parent.opts.inverse = null
+  forumCategoryModel.class.prototype.associations.forumcategoryparent.opts.inverse = null
```
thesan, line 56: @jodator, @p-sad in my understanding `forumCategoryModel` is all the parents/ancestors of a category and not the list not the subcategories is this correct ?

thesan:
What happened to the search bars ? I wanted to reuse them but I can't find any in the app :confused:

jodator:
> What happened to the search bars ? I wanted to reuse them but I can't find any in the app 😕

Yeah... I've noticed that too (#1118) The component is there somewhere, but it looks like one of the refactorings removed that. We'll need to fix that sooner or later.

packages/ui/src/mocks/server.ts
jodator: AFAICS they are stored in the same "table" so the relations are both ways.

So the field `parent` will be the direct parent. While `forumcategoryparent` is the other direction of the same relation. So it will hold "subcategories". Naming is automatic probably so it looks "weird".

So Having categories:

*   TopCategory(id=1)
    *   Sub A(id=2)
        *   SubASub1 (id=4)
        *   SubASub2 (id=5)
    *   Sub B(id=3)

The fields would be:

```
TopCategory
  parent: null
  forumcategoryparent:[2, 3]

Sub A
  parent: 1
  forumcategoryparent:[4,5]

Sub A Sub 1
  parent: 2
  forumcategoryparent:[]
```

jodator: @thesan & @p-sad  I don't see if this change breaks something (fetching related categories) so it should be safe.

However, if you see that some of the subcategories/parents is not fetched, please see into that.

jodator (approved):
AFAICS it does the trick and we can fetch top-level categories.

packages/ui/src/mocks/server.ts
thesan: I see. So it shouldn't be used to get the root categories then :disappointed:. Now it's only working because I implemented the `forumcategoryparent` as a list of parents.

jodator: OH.. but using `parent` there should work. I was sure I was testing that with `parent` but, no :D

jodator: OH.. but using `parent` there should work. I was sure I was testing that with `parent` but, no :D

jodator: A simple `null` will do the trick:

```graphql
query {
  forumCategories(where:{parent:null}) {
    id
    title
    parentId
  }
}
```

jodator: Can I ask you to fix that later and add @p-sad as a reviewer?

thesan: Got it, I will do that. Thanks

Issue #1197, opened 2021-08-09
Thread Filter component


thesan:
@[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A52493)

See #1058

Issue #1196, opened 2021-08-09
Threads List Component


thesan:
@[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A52493).
See #1058.

PR #1198, opened 2021-08-10
🎩 Implement the top part of the category page

thesan:
Still requires #1196 and #1197 to complete #1058


p-sad (approved):
Looks solid to me

PR #1201, opened 2021-08-11
Implement the `<ThreadListItem />` component

thesan:
See #1196


packages/ui/src/common/types/helpers.ts
```diff
@@ -1 +1,8 @@
 export type Reducer<Accumulator, Value = Accumulator> = (acc: Accumulator, value: Value, index: number) => Accumulator
+
+// From https://stackoverflow.com/a/54487392/4065560
+export type DataFields<T> = T extends Record<'__typename', any>
+  ? Omit<{ [K in keyof T]: DataFields<T[K]> }, '__typename'>
+  : T extends (infer U)[]
+  ? DataFields<U>[]
+  : T
```
p-sad, line 8: If I'm understanding this correctly, this type is for copying GQL `...FieldsFragment` types but without typenames, right? What's the intended use?

thesan: Yes.
In tests and storybook, I sometime need to directly pass mocked data missing `__typename` to the casting functions which require it (in this case a [`RawForumThreadMock`](https://github.com/Joystream/pioneer/blob/88053d1141adc82ef6ca6b33572e6fac69ba23e4/packages/ui/src/forum/components/threads/ThreadListItem.stories.tsx#L46) to [`asForumThread`](https://github.com/Joystream/pioneer/blob/88053d1141adc82ef6ca6b33572e6fac69ba23e4/packages/ui/src/forum/types/ForumThread.ts#L27)). Usually I uses a simple `Omit<XXXFieldsFragment, '__typename'>` but here there was the `createdInEvent` also missing `__typename`. So I added this to easily solve this kind of problem in the future. Do you think it might impact the compile time ?

p-sad (approved):

PR #1203, opened 2021-08-12
© Use copy button instead of copying on click

ClumsyVlad:


packages/ui/src/proposals/modals/VoteRationale/VoteRationale.tsx
```diff
@@ -39,8 +37,7 @@ export const VoteRationale = React.memo(() => {
     }
   }
 
-  const getVoteLink = () =>
-    copyValue(`${window.location.origin}/#${ProposalsRoutes.preview}/${vote?.proposalId}?showVote=${voteId}`)
+  const getVoteLink = `${window.location.origin}/#${ProposalsRoutes.preview}/${vote?.proposalId}?showVote=${voteId}}`
```
thesan, line 40: There is an extra "}" at the end of the link

thesan (changes requested):
Hey, there's only a problem with the VoteRationale modal link

thesan (approved):

PR #1220, opened 2021-08-16
🥊 Add thread status to mocks

p-sad:
So that we can move on to tasks that use it. Bad news is that `DataFields` type can no longer be used for `asForumThread` since the thread status is determined by the status's `__typename` in GQL — unless we want to write something sketchy looking like
```ts
type ThreadStatus = ForumThreadFieldsFragment['status']
export const asForumThread = (fields: DataFields<ForumThreadFieldsFragment> & { status: ThreadStatus }): ForumThread => ({
```
Personally, I think the readability of a core part of the code like these typecasts is more important than readability of storybook entries.


thesan (commented):
Agree about the core vs the stories readability.

Just about the `ForumThread.status`. Since other types (e.g `Proposal`, `Worker`) use a union type on their `status` field instead of a string. Maybe you could export the `ThreadStatus` type you defined in `forum/helpers/storybook.ts` in `forum/types/ForumThread.ts` and use it for `ForumThread.status`. WDYT ?

packages/ui/src/forum/types/ForumThread.ts
```diff
@@ -11,6 +10,7 @@ export interface ForumThread {
   categoryId: string
   tags: ForumThreadTag[]
   visiblePostsCount: number
+  status: string
```
thesan, line 13: I mean for this `status` field

thesan (commented):
Agree about the core vs the stories readability.

Just about the ForumThread.status. Since other types (e.g Proposal, Worker) use a union type on their status field instead of a string. Maybe you could export the ThreadStatus type you defined in forum/helpers/storybook.ts in forum/types/ForumThread.ts and use it for ForumThread.status. WDYT ?

packages/ui/src/forum/types/ForumThread.ts
p-sad: You're absolutely right, I think I put it in as a string for the meantime at first and then just forgot

thesan (approved):

Issue #1192, opened 2021-08-09
Post history in modal


p-sad:
[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=4809%3A128851)
Follow up:
- [x] Make the stepper light up for post currently on-screen
- [x] Clicking on bullet points on the left should scroll to their respective edits

PR #1221, opened 2021-08-17
🚯 Make edit at the top of history modal active

p-sad:
See #1192


packages/ui/src/forum/modals/PostHistoryModal/PostHistoryModal.tsx
```diff
@@ -24,7 +25,23 @@ export const PostHistoryModal = React.memo(() => {
...
+  const [activeEdit, setActiveEdit] = useState(0)
+
+  const editsInView = useMemo(() => new Array<boolean>(edits?.length ?? 0), [edits?.length])
+
+  const pickActiveEdit = useCallback(
+    (index: number) => (isCurrentInView: boolean) => {
+      editsInView[index] = isCurrentInView
+      const newActiveEdit = editsInView.reduceRight(
+        (currentTopIndex, isEditInView, index) => (isEditInView ? index : currentTopIndex),
+        0
```
krzysztof-jelski, line 37: As we're simply looking for the first occurrence of `true`, this can be simplified to:
```suggestion
      const newActiveEdit = editsInView.findIndex((value => value))
```

packages/ui/src/forum/modals/PostHistoryModal/PostHistoryModal.tsx
```diff
@@ -24,7 +25,23 @@ export const PostHistoryModal = React.memo(() => {
...
+      )
+      setActiveEdit(newActiveEdit)
+    },
+    [editsInView]
+  )
+
+  const viewport = useRef<HTMLDivElement>(null)
```
krzysztof-jelski, line 44: Seems like the viewport's ref does not change at all. Not sure we need it.

krzysztof-jelski (changes requested):
Remarks in separate comments.

packages/ui/src/forum/modals/PostHistoryModal/PostHistoryModal.tsx
p-sad: I forgot to assign it to the `StepperBody` - I wanted to consider margins so that points light up a bit before edits hit the absolute top of the view

krzysztof-jelski (approved):

PR #1137, opened 2021-07-28
🛍️ Implement the CategoryListItem component

thesan:
See #1065 

Closes #1064

PR #1229, opened 2021-08-19
🤢 Remove tooltip delay

### LINKED ISSUE #1222: Member role info box glitching
Upon hovering over [this info box](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=3420%3A2), it starts blinking seemingly randomly

ClumsyVlad:
Closes #1222

### DIFF
--- a/packages/ui/src/common/components/Tooltip/Tooltip.tsx
+++ b/packages/ui/src/common/components/Tooltip/Tooltip.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react'
+import React, { useState } from 'react'
 import ReactDOM from 'react-dom'
 import { usePopper } from 'react-popper'
 import { Link } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


ClumsyVlad:
Previously, I added a delay to delay displaying the tooltip when hovering over tooltip's trigger. Glitch appeared because of this. To keep a delay, but remove the glitch, is needed, as I think, one complex or several less complex event listeners. So I just removed this delay 🤷‍♂️

thesan (approved):
I think removing the delay is the right think to do, but now the issue from [#1137](https://github.com/Joystream/pioneer/pull/1137#discussion_r679117065) is back: when placing the mouse over an avatar from `<MemberStack />` then moving it over the next avatar on the right, the avatar moves but the tooltip stays in the wrong place.

Still it's more subtle than I remember and it only affects this particular case.

PR #1236, opened 2021-08-19
Thread category empty state

### LINKED ISSUE #1063: Overview threads browser
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=4361%3A122577)

Implement the overview thread browser:

- [ ] Display list of overview threds or polls
- [ ] Optional count of threads / polls
- [ ] Allow to browse next/previous (using connection query)
- [ ] Handle empty state (depends on usage)

ClumsyVlad:
Closes #1063

### DIFF
--- a/packages/ui/src/forum/components/threads/ThreadBrowser.tsx
+++ b/packages/ui/src/forum/components/threads/ThreadBrowser.tsx
@@ -46,6 +46,21 @@ export const ThreadBrowser = ({ label }: ThreadBrowserProps) => {
     setCurrentItemsGroup(currentItemsGroup + 1)
   }
 
+  if (!items) {
+    return (
+      <ThreadBrowserStyles>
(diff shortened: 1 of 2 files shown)


packages/ui/src/forum/components/threads/ThreadBrowser.tsx
```diff
@@ -46,6 +46,29 @@ export const ThreadBrowser = ({ label }: ThreadBrowserProps) => {
     setCurrentItemsGroup(currentItemsGroup + 1)
   }
 
+  if (!items) {
```
p-sad, line 49: There is a lot of duplicated code here, but I suppose that stems from not being sure how it's going to work with the query node. I think it can be simplified alongside coding in querying for these threads.

p-sad (approved):

PR #1239, opened 2021-08-19
⛓ Forum subcategories links

### LINKED ISSUE #1242: `validateDOMNesting` warning from subcategories links being nested in the category link
Also nesting the subcategories links in the category links causes a `validateDOMNesting` warning (still when using react dev tools) but this can be addressed in another issue.

_Originally posted by @thesan in https://github.com/Joystream/pioneer/issues/1239#issuecomment-902608344_

### LINKED ISSUE #1272: validateDOMNesting warning from subcategories links being nested in the category link
created from #1242

gitmazya:
Closes #1272 
Closes #1242

### DIFF
--- a/packages/ui/src/common/components/RouterLink.tsx
+++ b/packages/ui/src/common/components/RouterLink.tsx
@@ -11,7 +11,6 @@ export const RouterLink = styled(Link)<Props>`
 `
 
 export const GhostRouterLink = styled(Link)`
-  display: contents;
   color: unset;
   font-family: unset;
 `


packages/ui/src/common/components/buttons/LinkButtons.tsx
```diff
@@ -91,7 +91,7 @@ const LinkButtonInnerWrapper = styled.span<LinkButtonSizingProps>`
 `
 
 interface LinkButtonStyleProps extends Omit<LinkButtonProps, 'square'> {
-  $square: LinkButtonProps['square']
+  square?: LinkButtonProps['square']
 }
```
thesan, line 95: The `$square` (from #1034) is to prevent `square` to be passed down to the Link component which causes: `Warning: Received true for a non-boolean attribute square` when using React dev tools (it doesn't append with `<ButtonXXX />`). I don't know styled-components very well, is there another way to prevent this warning ?

thesan (changes requested):

thesan:
Also nesting the subcategories links in the category links causes a `validateDOMNesting` warning (still when using react dev tools) but this can be addressed in another issue.

packages/ui/src/common/components/buttons/LinkButtons.tsx
ClumsyVlad: Where can I find this error?

thesan: I'm not sure where on the app, but you can see it in the `Common/Buttons/LinkButtons` story (it first needs to be wrapped in `<MemoryRouter />` though)

PR #1240, opened 2021-08-19
🔌 Add minimum fee requirement to Create Thread modal

### LINKED ISSUE #1155: Add required balance check to create thread modal
If we check this here, we need to show some information about the required funds here.

This can be a follow-up ticket as well since we might ask the design team how to show that.

The other thing is that we _probably_ need the required funds to check as well – this is also a follo-up candidate to discuss if that check is needed.

ps.: Both comments might be treated as one follow-up actually :)

_Originally posted by @jodator in https://github.com/Joystream/pioneer/pull/1152#discussion_r679946152_

p-sad:
Closes #1155

### DIFF
--- a/packages/ui/src/forum/modals/CreateThreadModal/utils.ts
+++ b/packages/ui/src/forum/modals/CreateThreadModal/utils.ts
@@ -1,0 +1,5 @@
+import BN from 'bn.js'
+
+export const getMessage = (fee?: BN) => {
+  return `Insufficient funds to cover the thread creation. You need at least ${fee?.toString()} JOY on your account for this action.`
+}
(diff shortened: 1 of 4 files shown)


packages/ui/src/forum/modals/CreateThreadModal/utils.ts
```diff
@@ -0,0 +1,5 @@
+import BN from 'bn.js'
+
+export const getMessage = (fee?: BN) => {
+  return `Insufficient funds to cover the thread creation. You need at least ${fee?.toString()} JOY on your account for this action.`
```
krzysztof-jelski, line 4: Do we have some JOY formatting method? Not sure if `BN.toString()` would be consistent with currency formatting in other places.

krzysztof-jelski (approved):

packages/ui/src/forum/modals/CreateThreadModal/utils.ts
p-sad: That's how it's done in warnings in other parts of the site, so I'm inclined to leave it as is for now.

PR #1245, opened 2021-08-20
💙 Like button component + styles

ClumsyVlad:


packages/ui/src/forum/components/PostList/PostListItem.tsx
```diff
@@ -76,24 +78,19 @@ export const PostListItem = forwardRef<HTMLDivElement, PostProps>(
...
         <ForumPostRow>
+          <ButtonsGroup>{reaction && <LikeButton counter={reaction.length} />}</ButtonsGroup>
           {isThreadActive && !editing && (
             <ButtonsGroup>
-              {reaction && (
-                <Button size="small">
-                  <HeartIcon />
-                  {!!reaction.length && reaction.length}
-                </Button>
-              )}
```
thesan, line 86: I think the like button should be hidden when the thread is not active (see #1224)

DevNikita16:
@thesan We'd like to handle all opened PRs till Friday weekly. That's why I handled changed myself

thesan (approved):

PR #1264, opened 2021-08-24
✍🏿 Edit thread title functionality

### LINKED ISSUE #1166: Edit thread title
The UX should be the same as GH's edit post title action ([design](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=4813%3A135559)):

*   [x] When clicking "edit title" show the inline field and "save" button and "cancel" (button or link depending on the design).
*   [x] On save show "sign and send" transaction modal

DevNikita16:
Closes #1166 
I've handled all functionality, @ClumsyVlad will handle design part on Thursday - creating follow-up for this #1263


packages/ui/src/forum/components/Thread/ThreadTitle.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const formInitializer: TitleFormFields = {
+    title: thread.title,
+  }
+  const { fields, changeField } = useForm<TitleFormFields>(formInitializer, FormSchema)
+
+  const transaction = useMemo(() => {
+    if (isMyThread && threadAuthor && connectionState === 'connected' && api) {
+      return api.tx.forum.editThreadTitle(threadAuthor.id, thread.categoryId, thread.id, fields.title)
+    }
+  }, [fields.title, threadAuthor, isMyThread, connectionState])
```
p-sad, line 45: Personally, I'd instantiate the transaction in the modal instead, and keep this component relatively light (here it will be created even for non-OP as well). It feels like a variable that belongs in the modal where the transaction takes place

DevNikita16: Nice catch, done ✅

packages/ui/src/forum/components/Thread/ThreadTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const FormSchema = Yup.object().shape({})
+
+export const ThreadTitle = ({ thread }: ThreadTitleProps) => {
+  const { members: myMembers } = useMyMemberships()
+  const [isEditTitle, setEditTitle] = useState<boolean>(false)
+  const { showModal, hideModal } = useModal<EditThreadTitleModalCall>()
+
+  const isMyThread = thread && myMembers.find((member) => member.id === thread.id)
```
p-sad, line 30:
```suggestion
  const isMyThread = thread && myMembers.find((member) => member.id === thread.authorId)
```

p-sad: I think it should be something like this instead

p-sad (changes requested):

packages/ui/src/forum/components/Thread/ThreadTitle.tsx
DevNikita16: Oops

p-sad (approved):

Issue #1082, opened 2021-07-22
Search forum controls


jodator:
Add dedicated controls for searching in the [archived threads](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A51085):

- [ ] Author
- [ ] Archived date

No search by tags.

thesan:
See [figma comment](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7?node-id=5294:140700#104746410)

PR #1269, opened 2021-08-25
👀 Search through forum threads

### LINKED ISSUE #1268: Search for forum threads
[@figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=4558%3A126544)

thesan:
Closes #1268


thesan:
@p-sad I just went through the remaining issues, should this also close #1082 or do you think this search should also display results in a popup like on the categories page ? I'm hesitant because the category and archived category pages already have a list of threads with filters :thinking:

PR #1301, opened 2021-08-30
◼️ Assorted empty states

### LINKED ISSUE #1255: Show empty state in thread search results

p-sad:
Our empty states are kind of all over the place, so in addition to adding a state for empty threads list, I went over some other lists that I remembered being there. Now most of them look like an Apple advert - not the final look by any means, but I wanted them all to use the same style so it's easier to wrangle them all into proper shape later on.
Outliers include:
* My Memberships, which have a prompt for applying for a membership
* My Accounts, because I figured there might be a prompt asking for creating an account through the extension or something

Closes #1255


packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
```diff
@@ -13,18 +14,21 @@ export interface ProposalListProps {
...
-    </ProposalsListHeaders>
-    <List as="div">
-      {proposals.map((proposal) => (
-        <ProposalListItem key={proposal.id} proposal={proposal} isPast={isPast} />
-      ))}
-    </List>
-  </RowGapBlock>
-)
+export const ProposalList = ({ proposals, isPast }: ProposalListProps) =>
+  proposals.length ? (
```
jodator, line 18: I'd avoid big ternary operator blocks. An `if` will add some visual space, which I find easier to read:

```tsx
() => {
  if (!proposals.length) {
    return <NotFoundText>No.... </NotFoundText>
  }
  
  return (
    <RowGapBlock>
      //...
    </RowGapBlock>
  )
}
```

jodator (changes requested):
I'd change that one ternary operator to if-else statement.

jodator (approved):

Issue #1057, opened 2021-07-22
Forum overview


jodator:
@[figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A65528)

Related
- components: 

Implement the forum overview page:

- [ ] Recent polls
- [ ] Top 3 categories
- [ ] Forum activity

PR #1305, opened 2021-08-31
🦐 Add my threads browser

p-sad:
Related to #1057 
It feels a bit unwieldy wrapping the threads browser inside the component that fetches data for it; there's also an option of passing a collection of thread IDs to the browser component and fetching the actual threads from there, but that comes with its own host of problems.


packages/ui/src/forum/components/SuggestedThreads.tsx
```diff
@@ -18,14 +18,7 @@ export const SuggestedThreads = () => {
...
         {threads.map((thread) => (
-          <ThreadItem
-            key={thread.id}
-            answers={[{ answer: 'Text' }]}
-            badges={[{ badge: 'Text' }]}
-            content="Paragraph Medium. Will one day be essential for ensuring that the petabytes of media items uploaded to Joystream are formatted correctly and compreh..."
-            date={new Date().toString()}
-            title={thread.title}
-          />
+          <ThreadItem key={thread.id} thread={thread} answers={[{ answer: 'Text' }]} badges={[{ badge: 'Text' }]} />
```
jodator, line 21:
```suggestion
          <ThreadItem key={thread.id} thread={thread} />
```

Let's remove that as well. We don't have polls nor tags for now.

packages/ui/src/forum/components/threads/MyThreadsBrowser.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { useState } from 'react'
+
+import { useMyThreads } from '@/forum/hooks/useMyThreads'
+
+import { ThreadBrowser } from './ThreadBrowser'
+
+export const MyThreadsBrowser = () => {
+  const [currentPage, setCurrentPage] = useState(1)
+  const threadsPerPage = 2
+  const { threads, pageCount, totalCount, isLoading } = useMyThreads({ page: currentPage, threadsPerPage })
```
jodator, line 10:
```suggestion
  const { threads, pageCount, totalCount, isLoading } = useMyThreads({ page: currentPage, threadsPerPage: 2 })
```

I think we can remove that variable.

packages/ui/src/forum/components/threads/RandomThreadsBrowser.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const RandomThreadsBrowser = ({ label, maxThreads }: RandomThreadsBrowserProps) => {
+  const [currentPage, setCurrentPage] = useState(1)
+  const threadsPerPage = 2
+  const { threads, pageCount, totalCount, isLoading } = useRandomPaginatedThreads({
+    page: currentPage,
+    threadsPerPage,
+    maxThreads,
+  })
```
jodator, line 19:
```suggestion
  const { threads, pageCount, totalCount, isLoading } = useRandomPaginatedThreads({
    page: currentPage,
    threadsPerPage: 2,
    maxThreads,
  })
```

Similarly here.

packages/ui/src/forum/components/threads/ThreadItem.tsx
```diff
@@ -18,21 +20,21 @@ interface ThreadAnswerProps {
...
+  thread: ForumThread
   badges?: ThreadBadgeProps[]
   answers?: ThreadAnswerProps[]
   halfSize?: boolean
   empty?: boolean
 }
 
-export const ThreadItem = ({ title, date, content, badges, answers, halfSize, empty }: ThreadItemContentProps) => {
+export const ThreadItem = ({ thread, badges, answers, halfSize, empty }: ThreadItemContentProps) => {
+  const originalPost = useThreadOriginalPost(thread.id)
```
jodator, line 31: There's a weird delay while browsing the threads. I'm wondering what we could do about that.

The simplest solution would be to delay loading title+text until we have all the data.

packages/ui/src/forum/hooks/useThreadOriginalPost.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const useThreadOriginalPost = (threadId: string) => {
+  const { data, loading } = useGetForumPostsQuery({
+    variables: {
+      where: { thread: { id_eq: threadId } },
+      orderBy: ForumPostOrderByInput.DeletedAtAsc,
+      limit: 1,
+    },
+  })
+  return useMemo(() => data?.forumPosts[0] && asForumPost(data.forumPosts[0]), [data, loading])
```
jodator, line 16: Why did you use `useMemo()` hook here? AFAIR we don't do that in other hooks.

jodator (changes requested):
Overall good job :)

Besides small stylistic changes, I'm worried about the "odd" behavior of loading post text _after_ the title. Probably we can't do anything to fetch them together so maybe a delay loding the whole until we have everything could fix that.

Issue #1216, opened 2021-08-16
Allow preview on post edit


krzysztof-jelski:
When a user finishes editing a post, we should show post preview in confirm transaction modal.

PR #1306, opened 2021-09-01
👮🏻‍♂️ Edit proposal discussion post

### LINKED ISSUE #1087: Edit Proposal post
Requirements are the same as for forum posts: #1074

DevNikita16:
Closes #1087 

I have couple thoughts about this PR:
1) As You can see I've used modal from forum(not to duplicate code), so mb it should be in common now? WDYT
2) Wondering why `api.tx.proposalsDiscussion.updatePost` method is not using memberId? Probably we should ask Joystream team if it is correct.


packages/ui/src/forum/modals/PostActionModal/DeletePostModal/DeletePostModal.tsx
```diff
@@ -21,29 +22,37 @@ import { DeletePostModalCall } from '.'
...
+  const transaction = useMemo(() => {
+    if (api) {
+      if (type === 'forum' && forumPostData.categoryId && forumPostData.threadId) {
+        return api.tx.forum.deletePosts(
+          createType('ForumUserId', Number.parseInt(post.author.id)),
+          [[forumPostData.categoryId, forumPostData.threadId, post.id, true]],
+          ''
+        )
+      }
+      if (type === 'proposal' && proposalPostData.threadId) {
```
jodator, line 46: I'm thinking about this. I like the idea of not duplicating the code since there are a few places to change something. The problem I see with this solution is that it will require editing this code in the _potential_ future if we'd like to add similar actions to other places.

To combat we could introduce a way of changing the behavior from outside (let's call it dependency injection).

If that's a transaction only that needs to be updated then maybe pass it as a prop (or as a callback that returns a transaction).

jodator (changes requested):
Please see my comment about the solution. As for moving the logic to common, I'm not 100% sure what's the best solution. On one hand, it makes sense to move it as it will not be dependant on a specific domain but OTOH I'd imagine that someone would search that in the forum.

However, if you think that it makes more sense to move it common then I don't have anything against that :)

packages/ui/src/forum/modals/PostActionModal/DeletePostModal/DeletePostModal.tsx
DevNikita16: So, I've extracted this logic to PostEditor(Edit transaction) and PostContext(Delete transaction).

DevNikita16: I will add text check back to tests in this PR #1216

PR #1308, opened 2021-09-01
🌻 Display posts count in threads list

### LINKED ISSUE #1258: Count thread replies
In threads list and suggested threads

p-sad:
Closes #1258 
Looks like "answers" in suggested threads refer to poll answers (though in case they actually refer to replies, they should be trivial to implement)


p-sad:
Will use `visiblePostsCount` instead once it's in the API

PR #1315, opened 2021-09-02
🦻 Factor in post status when displaying options

p-sad:
For the longest time I had no idea how to test these buttons because they have no text in them, but then I found out I can add titles to them, which I think we should be doing anyway for the sake of accessibility


jodator (approved):
Looks OK

Issue #1062, opened 2021-07-22
Overview Thread component


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=4361%3A122577)

Implement base thread component (poll tracked here: #1122):

- [x] handle thread 
- [x] display title
- [x] display category breadcrumbs
- [x] display "time ago" field

For the Thread variant:
- [x] description excerpt
- [x] number of replies (answers)

PR #1322, opened 2021-09-02
🍔 Display thread overview details

p-sad:
See #1062 
Thread breadcrumbs sometimes don't fit in there - mostly in suggested threads - this could be addressed in a follow-up, though I'm not sure what's the best way to do it


DevNikita16 (approved):
👍

PR #1339, opened 2021-09-03
🦈 Create a post in thread

### LINKED ISSUE #1317: Post to a thread
Active threads should have a reply functionality. That's what the editor at the bottom is meant to convey I think

p-sad:
Closes #1317
This is a version that creates a locked (non-editable) post. Handling editable posts to be done in a follow-up, and we'll probably need a design for a modal step that explains what this does and what the post deposit is for.
Issue for an editable post can be found here: #1341

### DIFF
--- a/packages/ui/src/forum/components/Thread/ThreadNewPost.tsx
+++ b/packages/ui/src/forum/components/Thread/ThreadNewPost.tsx
@@ -5,11 +5,17 @@ import { CKEditor } from '@/common/components/CKEditor'
 import { InputComponent } from '@/common/components/forms'
 import { TextBig } from '@/common/components/typography'
 import { useModal } from '@/common/hooks/useModal'
+import { CreatePostModalCall } from '@/forum/modals/PostActionModal/CreatePostModal'
+import { ForumThread } from '@/forum/types'
 import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
 
 import { EditorWrap } from '../PostList/PostEditor'
 
-export const NewThreadPost = () => {
+interface NewPostProps {
+  thread: ForumThread
+}
+
+export const NewThreadPost = ({ thread }: NewPostProps) => {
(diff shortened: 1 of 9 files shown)


packages/ui/src/forum/components/Thread/ThreadNewPost.tsx
```diff
@@ -0,0 +1,55 @@
...
+  if (!active) {
+    return <TextBig>Pick an active membership to post in this thread</TextBig>
+  }
+
+  return (
+    <EditorWrap>
+      <InputComponent inputSize="auto">
+        <EditorMemo setNewText={setText} />
+      </InputComponent>
+      <ButtonsGroup>
```
jodator, line 32: This or the input should be wrapped in something (RowGapBlock?) in order to give some space to the button.

jodator, line 32: This or the input should be wrapped in something (RowGapBlock?) in order to give some space to the button.

packages/ui/src/forum/components/Thread/ThreadNewPost.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ForumThread } from '@/forum/types'
+import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
+
+import { EditorWrap } from '../PostList/PostEditor'
+
+interface NewPostProps {
+  thread: ForumThread
+}
+
+export const NewThreadPost = ({ thread }: NewPostProps) => {
```
jodator, line 18: A basic test for this is welcome as well :)

jodator, line 18: A basic test for this is welcome as well :)

jodator (changes requested):
This works well :) We would need a test for that, however.

PR #1348, opened 2021-09-03
🏎 Add post edit preview and refactor edit/delete modals

### LINKED ISSUE #1216: Allow preview on post edit

DevNikita16:
Closes #1216

### DIFF
--- a/packages/ui/src/forum/modals/PostActionModal/PostActionSignModal.tsx
+++ b/packages/ui/src/forum/modals/PostActionModal/PostActionSignModal.tsx
@@ -68,7 +68,9 @@ export const PostActionSignModal = ({
         <ModalBody>
           <RowGapBlock gap={24}>
             <RowGapBlock gap={16}>
-              <TextMedium>You intend to {action} your post.</TextMedium>
+              <TextMedium>
+                {action === 'edit' ? 'You intend to edit your post.' : 'You intend to delete your post.'}
+              </TextMedium>
               <TextMedium>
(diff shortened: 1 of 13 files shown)


packages/ui/src/forum/modals/PostActionModal/PostActionSignModal.tsx
```diff
@@ -40,33 +63,50 @@ export const PostActionSignModal = ({ transaction, service, controllerAccount, a
...
-            <TextMedium>{actionText}</TextMedium>
-            <TextMedium>
-              A fee of <TokenValue value={paymentInfo?.partialFee} /> will be applied to the transaction.
-            </TextMedium>
+    <>
+      <TransactionModal onClose={hideModal} service={service}>
+        <ModalBody>
+          <RowGapBlock gap={24}>
+            <RowGapBlock gap={16}>
+              <TextMedium>You intend to {action} your post.</TextMedium>
```
jodator, line 71:
```suggestion
              <TextMedium>You intend to {action} your post.</TextMedium>
```

We shouldn't do that for text fields. ATM we're using English, but there might be a need to translate the app.

Because of this I'd use two separate strings or just allow to pass a text as before. The reason for this is that it will be easier to translate a whole text.

jodator (changes requested):
I've left one comment about how we should handle texts for future translations.

Also, try to avoid "and" in PRs – it might be a reason to create two, smaller PRs. So if you feel adding "and" it might mean that the PR should be split into two.

jodator (approved):

PR #1359, opened 2021-09-06
🥖 Truncate breadcrumbs in thread items

### LINKED ISSUE #1062: Overview Thread component

p-sad:
Closes #1062 
Final touch for thread overview - breadcrumbs would get massive and take up half the space in the component, so they need to be truncated if they're too verbose. The style was changed site-wide but in practice we shouldn't see it much outside forum overview, and if someone's screen is narrow enough the crumbs should sit neatly in a single line anyway.
Also removed the thread title breadcrumb from there since the thread title is at the top of the component anyway.

### DIFF
--- a/packages/ui/src/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsItem.tsx
+++ b/packages/ui/src/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsItem.tsx
@@ -1,6 +1,6 @@
 import React, { ReactNode } from 'react'
 import { Link } from 'react-router-dom'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 
 import { Colors, Transitions, Fonts } from '../../../../constants'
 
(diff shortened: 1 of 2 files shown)


jodator:
>   
> Final touch for thread overview - breadcrumbs would get massive and take up half the space in the component, so they need to be truncated if they're too verbose. The style was changed site-wide but in practice we shouldn't see it much outside forum overview, and if someone's screen is narrow enough the crumbs should sit neatly in a single line anyway.  
> Also removed the thread title breadcrumb from there since the thread title is at the top of the component anyway.

Looks OK :) I wonder only if we shouldn't remove some parts of the path to hide the middle steps. But I think it will be better to see real-world data and fix them then.

PR #1360, opened 2021-09-06
⛵ Make the forum navigable with the search pane

### LINKED ISSUE #1326: Remove the gray overlay for search results
Or maybe the dark overlay could be removed?

thesan:
Since the search pane also allows to explore categories with each results breadcrumbs, you can now navigate through the all forum with the search result pane open. The overlay disappears once the route changes.

Closes #1326


jodator:
> Since the search pane also allows to explore categories with each results breadcrumbs, you can now navigate through the all forum with the search result pane open. The overlay disappears once the route changes.

I wonder if this is the best UX. However, I don't have a strong opinion on that. I'll close that and we will discuss that with the design team later. ATM it is OK how it works IMO

PR #1372, opened 2021-09-08
🪛 Fix the root categories and archived query MirageJS result

### LINKED ISSUE #1362: MirageJS mock server issues
MirageJS doesn't work with `id_eq: null` (Categories in forum are not displayed).

thesan:
Closes #1362


jodator (approved):
Thanks!

PR #1421, opened 2021-09-09
␛🪝 Escape event hook

### LINKED ISSUE #1357: Create a hook to close elements on 'esc'

ClumsyVlad:
Closes #1357


jodator (approved):
Looks OK a test would be awesome, but for now, we can leave without it.

Issue #1280, opened 2021-08-26
Expand `excludePaths` and `knownRoutes` lists for breadcrumbs


p-sad:
We haven't taken a look at `src/app/constants/breadcrumbs.ts` for a while and a bunch of dead/incorrect routes appeared in breadcrumbs while the site was expanding. I think someone needs to go through available pages and make sure only links to valid pages are left

jodator:
Let's see if anyone submit an issue for this. Looks like handled in #1429

PR #1429, opened 2021-09-09
🧽 Tidy up some routes/breadcrumbs

p-sad:
See #1280 
I'm not sure if I should close that issue because keeping an eye on breadcrumbs is more of a continuous task, given that each page we add adds one.


jodator (approved):
Thanks!

PR #1449, opened 2021-09-10
🐫 Multi transaction modals updates

### LINKED ISSUE #1171: Update multi-transaction handling
ATM we have two transactions that need two different signers (requires `api.tx.members.addStakingAccountCandidate()` API call that needs to be signed by the account that will be used for staking):

*   create proposal
*   apply for role

Right now we're issuing always two transactions. But there's another scenario, in which an account was bounded already so we don't need (or can't) invoke it again. In such a scenario, we should go directly to the second transaction.

The second missing thing is that we don't send `api.tx.members.confirmStakingAccount()` with the main transaction (it should be done in batch).

Requirements:

*   [x] Add confirm staking account to the main transaction and use batch for:
    *   create proposal
    *   apply for role
*   [x] Check if the selected account is bound and display only in the main transaction

jodator:
Closes #1171

This PR adds the real-world handling of the two-step modals (mostly those that require staking). It is not ideal as it has some code duplication for the modal and for the machine as well. However, I thought that it is OK for now as we have only two modals that require staking.

It would be nice that you could verify (for instance for proposals) that this works. I was able to test that using two steps:

*   at a first run, I used an account that was not bounded and I've passed two transactions (staking + proposal creation)
*   for the second check, I've used the same account for bonding (so it was already bounded) so I had only one transaction (proposal creation)

### DIFF
--- a/packages/ui/src/working-groups/modals/ApplyForRoleModal/ApplyForRoleModal.tsx
+++ b/packages/ui/src/working-groups/modals/ApplyForRoleModal/ApplyForRoleModal.tsx
@@ -14,6 +14,7 @@ import { useApi } from '@/common/hooks/useApi'
 import { useModal } from '@/common/hooks/useModal'
 import { getEventParam, metadataToBytes } from '@/common/model/JoystreamNode'
 import { getSteps } from '@/common/model/machines/getSteps'
+import { useMember } from '@/memberships/hooks/useMembership'
 import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
 import { BindStakingAccountModal } from '@/memberships/modals/BindStakingAccountModal/BindStakingAccountModal'
 import { SwitchMemberModalCall } from '@/memberships/modals/SwitchMemberModal'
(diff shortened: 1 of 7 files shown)


p-sad:
Tried to replicate the steps described in PR description; if I understand correctly, when I make a proposal the 2nd time around and use the same account for staking, it should skip the 1st transaction and go straight to the "Create proposal" transaction. It didn't though, and the extension returned `members.StakingAccountIsAlreadyRegistered` error once I tried to bind the same account for the second time.

packages/ui/src/working-groups/modals/ApplyForRoleModal/ApplyForRoleModal.tsx
```diff
@@ -108,41 +122,50 @@ export const ApplyForRoleModal = () => {
...
     })
 
+    let transaction: SubmittableExtrinsic<'rxjs', ISubmittableResult>
+
+    if (stakingStatus === 'confirmed') {
+      transaction = applyOnOpeningTransaction
+    } else {
+      transaction = api.tx.utility.batch([
+        applyOnOpeningTransaction,
+        api.tx.members.confirmStakingAccount(activeMember?.id, stake.account.address),
```
p-sad, line 165: How does batch processing work exactly? I'm assuming it's sequential - shouldn't the `confirmStakingAccount` transaction be first in batch?

jodator: Oh... nice catch :D

jodator: I'd assume that it should work as you described.

PR #1450, opened 2021-09-10
🥖 Keep the thread item breadcrumbs on one line

### LINKED ISSUE #1446: Long thread breadcrumbs are multi-line
This is how the breadcrumbs are displayed in the right side of the page:

It may be worse when there's even bigger nesting (e.g. 4 levels or more) and/or text are long.

thesan:
Closes #1446

### DIFF
--- a/packages/ui/src/forum/components/threads/ThreadItemBreadcrumbs.tsx
+++ b/packages/ui/src/forum/components/threads/ThreadItemBreadcrumbs.tsx
@@ -1,25 +1,94 @@
-import React, { memo } from 'react'
+import React, { memo, useEffect, useMemo, useRef, useState } from 'react'
 import styled from 'styled-components'
 
 import {
   BreadcrumbsItem,
+  BreadcrumbsItemComponent,
   BreadcrumbsItemLink,
   BreadcrumbsItemText,
 } from '@/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsItem'
 import { BreadcrumbsListComponent } from '@/common/components/page/Sidebar/Breadcrumbs/BreadcrumbsList'
 import { Colors, Fonts } from '@/common/constants'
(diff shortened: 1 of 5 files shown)


thesan:
@jodator the suggested thread section is too narrow to fit many of the long breadcrumbs. Should the middle segments sometime be replace by a "..." depending on the available space ?

jodator:
> @jodator the suggested thread section is too narrow to fit many of the long breadcrumbs. Should the middle segments sometime be replace by a "..." depending on the available space ?

I'll ask the design team and post the answer. I see that we either do "..." in the middle or allow to scroll the breadcrumbs.

Both are not ideal IMO

*   "..." can be tricky to implement correctly
*   scroll might be ugly

thesan:
I didn't manage to test to the logic properly. But at least the story present the different cases.

packages/ui/src/forum/components/threads/ThreadItemBreadcrumbs.tsx
```diff
@@ -1,43 +1,129 @@
...
+
+      const widthDifference = containerBox.width - childrenWidth
+
+      if (widthDifference >= 0) {
+        setKeptItemOnEachSide(undefined)
+      } else {
+        const totalExtraChildren = Math.ceil(-widthDifference / MIN_BREADCRUMB_WIDTH)
+        setKeptItemOnEachSide(Math.floor((children.length - 2 - totalExtraChildren) / 2))
+      }
+    }, 1000)
```
jodator, line 82: Wow, that actually quite complex. I see that this works quite well. However, I feel that this is a bit too much for our needs ATM. Since it is already here, I don't see a reason to remove that.

jodator (commented):
Other than I'd try doing it a simpler mechanism for start, the fully fledged bounding rect solution is OK :)

PR #1454, opened 2021-09-13
🍋 Post in proposal discussion

### LINKED ISSUE #1427: Create post in proposal discussion

p-sad:
Closes #1427 
A couple of notes about this one: 
* I decided to reuse components from forum; now the transaction gets injected into the modal via modal data, much like in the Edit Post modal. The only downside to that is that params passed to the modal cannot be tested that well anymore, as functions can't be matched as easily as data objects.
* The const used for the post deposit should be different based on type (forum vs proposal discussion) but the const for discussion post isn't exposed in the current version of the API. We probably need an issue open to look out for it with the next version of the API. Currently the const from forum is used.


packages/ui/test/forum/components/NewThreadPost.test.tsx
```diff
@@ -48,44 +50,102 @@ describe('UI: Add new post', () => {
...
+      useMyMemberships.setActive(getMember('alice'))
+      renderEditor(props)
+      expect(await getButton('Post a reply')).toBeDisabled()
+    })
 
-  it('Disables the post button if text is empty', async () => {
-    useMyMemberships.setActive(getMember('alice'))
-    renderEditor()
-    expect(await getButton('Post a reply')).toBeDisabled()
+    it('Opens the modal', async () => {
```
jodator, line 67:
```suggestion
    it('...', async () => {
```

Should describe the action or state not what happens

packages/ui/test/forum/components/NewThreadPost.test.tsx
```diff
@@ -48,44 +50,102 @@ describe('UI: Add new post', () => {
...
-  })
+  describe('In a forum thread', () => {
+    const props: NewPostProps = { type: 'forum', thread: { id: '1', categoryId: '1', title: 'thread' } }
+
+    it('Hides editor if no membership is selected', async () => {
+      renderEditor(props)
+      expect(await screen.findByText('Pick an active membership to post in this thread')).toBeDefined()
+    })
+
+    it('Disables the post button if text is empty', async () => {
```
jodator, line 61: As above, this should describe the state which we test

packages/ui/test/forum/components/NewThreadPost.test.tsx
```diff
@@ -48,44 +50,102 @@ describe('UI: Add new post', () => {
     useMyMemberships.active = undefined
   })
 
-  it('Hides editor if no membership is selected', async () => {
-    renderEditor()
-    expect(await screen.findByText('Pick an active membership to post in this thread')).toBeDefined()
-  })
+  describe('In a forum thread', () => {
```
jodator, line 53:
```suggestion
  describe('Forum thread', () => {
```

A bit shorter ;)

packages/ui/src/forum/components/Thread/NewThreadPost.tsx
```diff
@@ -1,29 +1,71 @@
...
   if (!active) {
     return <TextBig>Pick an active membership to post in this thread</TextBig>
   }
 
+  if (
+    props.type === 'proposalDiscussion' &&
+    props.thread.mode === 'closed' &&
+    !props.thread.whitelistIds?.includes(active.id)
+  ) {
+    return <TextBig>The discussion of this proposal is closed; only select members can comment on it.</TextBig>
```
DevNikita16, line 66:
```suggestion
    return <TextBig>The discussion of this proposal is closed. Only whitelist members can comment on it.</TextBig>
```

jodator (changes requested):
I think that the logic when something is rendered or what transaction to call should be totally outside the base components. So actually the component that renders `<NewTrheadPost />` should decide what transaction to use and when display the edit box for some special cases (e.g. whitelist). The handling of an active member should stay as it is now as it is a general check IMO.

p-sad:
One more comment: proposal discussion logic completely does not take council members into consideration, because as far as I know we don't have much implemented regarding council. Council members should be able to post in discussions regardless of whitelist status.

jodator:
> One more comment: proposal discussion logic completely does not take council members into consideration, because as far as I know we don't have much implemented regarding council. Council members should be able to post in discussions regardless of whitelist status.

Please add a ticket for that as well :+1:

jodator (approved):
Cool 😎  I think that it is a bit better now :)

PR #1457, opened 2021-09-13
🙈 Hide pagination component when there's only one page

### LINKED ISSUE #1439: Hide pagination component when there's only one page
I think that showing pagination when there's only one page doesn't make sense.

We could make the pagination component render `null` when pages count is `1`.

DevNikita16:
Closes #1439

### DIFF
--- a/packages/ui/src/app/pages/Forum/ForumMyThreads.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumMyThreads.tsx
@@ -14,17 +14,12 @@ export const ForumMyThreads = () => {
   const [page, setPage] = useState(1)
   const { threads, pageCount, isLoading } = useMyThreads({ page, threadsPerPage: 5 })
 
-  const pagination = useMemo(
-    () => !isLoading && !!pageCount && <Pagination pageCount={pageCount} handlePageChange={setPage} page={page} />,
-    [isLoading, pageCount, page]
-  )
-
   const displayThreads = () => {
     return (
       <RowGapBlock gap={24}>
-        {pagination}
+        <Pagination pageCount={pageCount} handlePageChange={setPage} page={page} />
         <ThreadList threads={threads} onSort={() => null} isLoading={isLoading} />
-        {pagination}
+        <Pagination pageCount={pageCount} handlePageChange={setPage} page={page} />
       </RowGapBlock>
     )
   }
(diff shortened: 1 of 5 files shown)


packages/ui/src/app/pages/Forum/ForumMyThreads.tsx
```diff
@@ -14,16 +14,17 @@ export const ForumMyThreads = () => {
   const [page, setPage] = useState(1)
   const { threads, pageCount, isLoading } = useMyThreads({ page, threadsPerPage: 5 })
 
+  const pagination = useMemo(
+    () => !isLoading && !!pageCount && <Pagination pageCount={pageCount} handlePageChange={setPage} page={page} />,
```
p-sad, line 18: The `!!pageCount` check seems redundant here and in `Members.tsx`, seeing that `pageCount <2` is performed in the Pagination component itself. Suggestion in below comment-

packages/ui/src/common/components/Pagination/Pagination.tsx
```diff
@@ -11,27 +11,33 @@ interface PaginationProps {
...
-      nextLabel={<Arrow direction="right" />}
-      nextLinkClassName="pagination__link"
-      previousLinkClassName="pagination__link pagination__link--previous"
-      onPageChange={(value) => handlePageChange(value.selected + 1)}
-      forcePage={page && page - 1}
-    />
-  </StyledPaginateContainer>
-)
+export const Pagination: FC<PaginationProps> = ({ pageCount, handlePageChange, page }) => {
+  if (pageCount < 2) {
```
p-sad, line 15:
```suggestion
  if (pageCount ?? 0 < 2) {
```
This way we could catch nullish pageCount in here instead of parent component, making it a bit more concise.
I don't have a strong opinion on this though, can be left as is.

p-sad (approved):

packages/ui/src/app/pages/Forum/ForumMyThreads.tsx
DevNikita16: Nice catch!

packages/ui/src/common/components/Pagination/Pagination.tsx
DevNikita16: @p-sad Than is loading also redundant, because pageCount also will be undefined/null

PR #1470, opened 2021-09-15
✍🏼 Display post edits properly

### LINKED ISSUE #1464: The edited versions are out of order
The [editing history](https://pioneer-2.netlify.app/#/forum/thread/10?post=40) events seem out of order:

DevNikita16:
Closes #1464 

Also edited text now depends of last edit time, because `updatedAt` can represent other post data updates as it was on Olympia


packages/ui/src/forum/queries/forum.graphql
```diff
@@ -204,7 +207,7 @@ query GetForumPosts($where: ForumPostWhereInput!, $orderBy: [ForumPostOrderByInp
 }
 
 query GetForumPostEdits($id: ID!) {
-  edits: postTextUpdatedEvents(where: { post: { id_eq: $id } }) {
+  edits: postTextUpdatedEvents(where: { post: { id_eq: $id } }, orderBy: [createdAt_ASC]) {
```
p-sad, line 210: I'd assume the most recent edit would go to the top instead, as it's the most relevant one. Not sure what the design is though

p-sad (approved):
Looks good in the way that they're now ordered, but I'm not sure if this is the right order

PR #1471, opened 2021-09-15
🎟️ Hide the "transfer invites" button of other users

### LINKED ISSUE #1441: Hide transfer invites button
On the member details page there's a "transfer invites" button. It should be visible only for the user's membership.

thesan:
Closes #1441

### DIFF
--- a/packages/ui/src/memberships/components/MemberProfile/MemberDetails.tsx
+++ b/packages/ui/src/memberships/components/MemberProfile/MemberDetails.tsx
@@ -1,5 +1,7 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import styled from 'styled-components'
+
+import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
 
 import { BlockTime } from '../../../common/components/BlockTime'
 import { TransferSymbol } from '../../../common/components/icons/symbols'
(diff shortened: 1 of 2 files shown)


packages/ui/src/memberships/components/MemberProfile/MemberDetails.tsx
```diff
@@ -22,6 +24,12 @@ type Props = { member: Member }
 export const MemberDetails = React.memo(({ member }: Props) => {
   const { member: memberDetails, isLoading } = useMember(member.id)
 
+  const myMemberships = useMyMemberships()
+  const isMyMembership = useMemo(() => myMemberships.members.some(({ id }) => id === member.id), [
+    member.id,
+    myMemberships,
+  ])
```
jodator, line 31: WDYT about exposing `isMyMember()` in the `useMyMemberships()` hook? I think that we potentially have other places that require a similar check.

I'm thinking about something like this:

```typescript
const { isMyMembership } = useMyMemberships()

return (
  <>
    {isMyMembership(member.id) ? "It's you!" : "Sorry, I can't do this Dave."}
  </>
)
```

jodator (commented):
Generally OK. I've added an optional idea to help in other places.

packages/ui/src/memberships/components/MemberProfile/MemberDetails.tsx
thesan: Makes sense :ok_hand:

PR #1472, opened 2021-09-15
🦇 Pass correct transaction parameters in Edit Post

### LINKED ISSUE #1463: The edit post transaction is not working
I was using "edit post" action on the [testnet](https://pioneer-2.netlify.app/#/forum/thread/10?post=40) and it produces weird results.

The Post was edited from

> 2 0I might like to edit this post in the future…. 

To:

> 0I might like to edit this post in the future…. aaaand I did

But it produced: 

> 0I might like to edit this post in the future…. 

It needs to be checked whether it is a bug on Pioneer side (e.g. Wrong transaction parameters) or an upstream side.

p-sad:
Closes #1463 
As it turns out, the edited text was missing from the memo deps list.
I opted to use `useCallback` instead, as to not render too many hooks too often


DevNikita16 (approved):
image
Works on testnet 🤩

PR #1473, opened 2021-09-15
🗜️ Fix the thread item breadcrumbs expanding on resize

thesan:
Sorry I was done with this but then I realized the behavior was sometime strange whenever the window would resize


jodator (approved):
Thanks for noticing that :) I see that the code get a bit simpler, so that's a plus 👍

Issue #1465, opened 2021-09-13
Submitting the transaction should clear editing state


jodator:
This might be tricky to implement, so a minimal requirement is to clear the editing state of a post right after the user hits "save".

The drawback is that user could close the transaction window in order to re-edit the data. This however, is a smaller issue than seeing this data still in the edit field.

jodator:
@DevNikita16 I'm closing your #1488 PR but we need a similar handling when

*   [ ] Responding to a thread (It can clear the editor state when transaction is done)
*   [ ] Editing post title (Let's keep that simple: click on :white\_check\_mark: will close edit box)

PR #1483, opened 2021-09-16
🔐 Show transaction error modal when Polkadot extension is closed

### LINKED ISSUE #1482: Show transaction error modal when closing Polkadot extension

DevNikita16:
Closes #1482

### DIFF
--- a/packages/ui/src/common/modals/TransactionModal/TransactionModal.tsx
+++ b/packages/ui/src/common/modals/TransactionModal/TransactionModal.tsx
@@ -20,8 +20,6 @@ export interface TransactionModalProps {
 export const TransactionModal = ({ onClose, children, service, title, useMultiTransaction }: TransactionModalProps) => {
   const [state] = useActor(service)
 
-  console.log(state)
-
   if (state.matches('prepare')) {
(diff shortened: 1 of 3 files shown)


DevNikita16:
Wondering how commit from previous PR appeared here

DevNikita16:
@jodator also what should be instead of `any` for `onTransactionError` type. I was playing with types, but haven't succeeded

jodator (commented):
It would be actually quite good to show a different error message. I think that we need to have a different event for that case.

If you do this now, we could introduce better error handling.

DevNikita16:
> It would be actually quite good to show a different error message. I think that we need to have a different event for that case.
> 
> If you do this now, we could introduce better error handling.

I'd like to dive into my current [issue](https://github.com/Joystream/pioneer/issues/1465). So if we could just create issue for later would be awesome 🤩

jodator:
> @jodator I'd like to dive into my current [issue](https://github.com/Joystream/pioneer/issues/1465). So if we could just create issue for later would be awesome 🤩

OK, so please create an issue for that.

PR #1488, opened 2021-09-16
🥴 Handle actions on success and canceled/failed post editing

### LINKED ISSUE #1465: Submitting the transaction should clear editing state

DevNikita16:
See #1465

Alright, it was pretty hard to make perfect UX. 

I've come up with these solutions:
1) On success - we update text and time of post
2) On error or on canceling modal - we focus user back to inline editor. Previously it was looking really weird.

Regarding 2nd point - that was the big fight against react ref types. But it works nice as I think

### DIFF
--- a/packages/ui/src/forum/components/Thread/NewThreadPost.tsx
+++ b/packages/ui/src/forum/components/Thread/NewThreadPost.tsx
@@ -50,7 +50,7 @@ export const NewThreadPost = React.forwardRef(
                 </Badge>
               </div>
               <div>
-                <ButtonPrimary size="small" onClick={removeReply}>
+                <ButtonPrimary size="small" square onClick={removeReply}>
                   <CrossIcon />
                 </ButtonPrimary>
               </div>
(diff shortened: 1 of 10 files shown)


packages/ui/src/forum/components/Thread/NewThreadPost.tsx
```diff
@@ -50,7 +50,7 @@ export const NewThreadPost = React.forwardRef(
                 </Badge>
               </div>
               <div>
-                <ButtonPrimary size="small" onClick={removeReply}>
+                <ButtonPrimary size="small" square onClick={removeReply}>
```
DevNikita16, line 53: This is out of PR scope. Just Vlad noticed that I need to add `square` here

jodator:
> On error or on canceling modal - we focus user back to inline editor. Previously it was looking really weird.

:+1:

jodator (approved):
I didn't catch anything suspicious, so merging ;)

PR #1512, opened 2021-09-17
🍬 Make the Create Thread modal nicer

### LINKED ISSUE #1447: Make Markdown editor fields higher
ATM, we start with a single line of the editing fields, which doesn't look good.

I think that we should update the initial height of the Markdown editor.

### LINKED ISSUE #1448: Make the Create Thread modal nicer
Right now there's too much space: 

I think that making #1447 & removing the "General" title would help.

thesan:
Closes #1447
Closes #1448


jodator:
@thesan The fix looks cool :) Could you add two more things to the editor

*   max-height handling
*   allow passing (optional) min/max height to the component (but use defaults if not provided)

jodator (changes requested):
As in the comment above ☝️ Adding those two will give us sane defaults and the flexibility to adjust sizes in various places.

PR #1516, opened 2021-09-20
🥝 Calculate worker's total earnings

### LINKED ISSUE #1495: Display Worker earnedTotal
We have `useMyEarnings()` hook that calculates earnings.

A similar hook should be used to handle [`Worker.earnedTotal`](https://github.com/Joystream/pioneer/blob/ef731424d6a09817977768ef0e252bb2c003868b/packages/ui/src/working-groups/types/Worker.ts#L73):

*   remove `earnedTotal` field from `Worker`
*   properly calculate the value using hook

p-sad:
Closes #1495 
This looks nice for now, but seeing how payouts take place every ~24h, I'm a little worried they'll start adding up really quickly - each of one's roles might end up having to fetch thousands of events before counting, after just three years.

### DIFF
--- a/packages/ui/src/memberships/components/MemberProfile/MemberRoleToggle.stories.tsx
+++ b/packages/ui/src/memberships/components/MemberProfile/MemberRoleToggle.stories.tsx
@@ -28,7 +28,6 @@ Default.args = {
   role: {
     id: '123',
     runtimeId: 12,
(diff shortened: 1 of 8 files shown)


jodator (approved):
Nice :)

👍 For tests and small PR

Issue #1385, opened 2021-09-08
Display council members


jodator:
@[figma](https://www.figma.com/file/Z3KcRoAll3JGEY622o0iZu/Council?node-id=220%3A22645)

Display a list of council members

- [x] Fetch council members from the query-node
- Handle fields
  - [x] Member info
  - [x] Own reward
  - [x] Voters stake
  - [x] Elected times count
- [x] Click on the council member row opens member details
- [x] Allow sorting by the above
- [x] Handle empty state

thesan:
@jodator what comes to my mind now is the various common errors which can happen when mocking data, eg:
- `Mirage: The xxx model has multiple possible inverse associations for the xxx.xxx association.`
- The error when an object is passed as an attribute in the seeding process instead of creating a model...
- Or even, no resolver being found when q new query hasn't been added to the Mirage server

thesan:
@jodator is the second field the "Owed reward" which would be the `CouncilMember.unpaidReward` from schema? I haven't found anything about a "Own" reward in the handbook

jodator:
> @jodator is the second field the "Owed reward" which would be the `CouncilMember.unpaidReward` from schema?

I think yes. It might be a typo. IMO it should be "owed" as unpaid [reward](https://joystream.gitbook.io/joystream-handbook/governance/council#rewards).

PR #1525, opened 2021-09-22
🏛️ Add the councilor list to the app

thesan:
See #1385


packages/ui/src/common/components/typography/TokenValue.tsx
```diff
@@ -15,7 +17,7 @@ interface ValueProps extends ValueSizingProps {
 }
 
 export const TokenValue = React.memo(({ className, value, size }: ValueProps) => {
-  if (!value) {
+  if (value === null || !isDefined(value)) {
```
jodator, line 20: Oh that was unexpected :D But I see that we had `number` there...

packages/ui/src/council/queries/council.graphql
```diff
@@ -0,0 +1,24 @@
+fragment CouncilMemberFields on CouncilMember {
+  id
+  member {
+    ...MemberFields
+    councilMembers {
+      __typename
+    }
```
jodator, line 7: This looks odd. Why do we need to query this? I couldn't find where it was used.

packages/ui/src/council/hooks/useElectedCouncil.ts
```diff
@@ -0,0 +1,8 @@
+import { useGetElectedCouncilsQuery } from '@/council/queries'
+import { asCouncil } from '@/council/types'
+
+export const useElectedCouncil = () => {
+  const { loading, data } = useGetElectedCouncilsQuery({ variables: { where: { deletedAt_all: false } } })
```
thesan, line 5: @jodator my interpretation is that the `ElectedCouncil` gets "deleted" once the council is replaced, is that correct?

jodator (commented):
Really nice 🚀 

I have only one question about `__typename` use.

packages/ui/src/council/hooks/useElectedCouncil.ts
jodator: I think that this comes from Hydra and is automatically added to all `where` variables.

Maybe `endedAtBlock` (when null) is better. ATM I don't see other field that could indicate currently elected council.

jodator: An alternative way is to query the API:

```
api.query.council.councilMembers()
```

packages/ui/src/council/queries/council.graphql
thesan: So far the only way I found to get all the `numberOfTerms` in one request (number of time a member was elected council), is to count the `member.councilMembers`. So there's no need for any real data in the `councilMembers`

jodator: Ah right, thanks!

packages/ui/src/common/components/typography/TokenValue.tsx
thesan: What do you mean?

jodator: Just a comment that this bug was a hidden one :) So when `0` was passed instead of `new BN(0)` the value was `-`.

Anyway, closing this.

packages/ui/src/council/hooks/useElectedCouncil.ts
thesan: Right! I always forget that passing `null` is an option

PR #1532, opened 2021-09-23
🧑‍🤝‍🧑 Sort councilors and handle empty state

### LINKED ISSUE #1385: Display council members

### LINKED ISSUE #1526: Sort councilor
See #1385

### LINKED ISSUE #1527: Handle the empty council state
See #1385

thesan:
Closes #1385
Closes #1526
Closes #1527

### DIFF
--- a/packages/ui/src/app/pages/Council/Council.tsx
+++ b/packages/ui/src/app/pages/Council/Council.tsx
@@ -1,18 +1,21 @@
-import React, { useMemo, useRef } from 'react'
+import React, { useMemo, useRef, useState } from 'react'
 
 import { PageHeaderRow, PageHeaderWrapper, PageLayout } from '@/app/components/PageLayout'
 import { MainPanel } from '@/common/components/page/PageContent'
 import { PageTitle } from '@/common/components/page/PageTitle'
 import { SidePanel } from '@/common/components/page/SidePanel'
-import { CouncilList } from '@/council/components/councilList'
+import { CouncilList, CouncilOrder } from '@/council/components/councilList'
 import { useElectedCouncil } from '@/council/hooks/useElectedCouncil'
+import { Councilor } from '@/council/types'
 
 import { CouncilTabs } from './components/CouncilTabs'
 
 export const Council = () => {
   const { council, isLoading } = useElectedCouncil()
-  const councilors = useMemo(() => council?.councilors ?? [], [council])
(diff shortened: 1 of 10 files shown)


packages/ui/src/app/pages/Council/Council.tsx
```diff
@@ -25,11 +29,22 @@ export const Council = () => {
...
 
   return <PageLayout header={header} main={main} sidebar={sidebar} />
 }
+
+const sortBy = ({ key, isDescending }: CouncilOrder): ((a: Councilor, b: Councilor) => number) => {
+  const direction = isDescending ? -1 : 1
+  return key === 'member'
+    ? (a, b) => a.member.handle.localeCompare(b.member.handle) * direction
+    : (a, b) => (a[key] - b[key]) * direction
+}
```
thesan, line 50: @jodator in my understanding the current council will have a small number of member, so I'm thinking that sorting the data locally makes more sense than sending queries. WDYT?

jodator:
@thesan WDYT about adding tests for this page?

Writing every case would be time-consuming, but I think that from the business logic perspective the most important cases could be:

*   display proper council list (having a past council in the DB as well) – just to be sure we query for the right councilors
*   maybe empty state as well

Loading, sorting might be something to pass on tests.

Also, I was thinking about whether to test: page or `useElectedCouncil()` hook. I think that testing on the page level _might_ be easier and will be also an integration test (page + hook).

jodator (commented):
As in the comment above ☝️ 

The code looks good, but there's _a bit_ of business logic so it might make sense to write a simple test for that page.

packages/ui/src/app/pages/Council/Council.tsx
jodator: Yeah, I saw that too. I'm for sorting locally since we would get already the whole council as you wrote.

jodator:
Thank you!

PR #1540, opened 2021-09-24
🪜 Fix the Markdown editor height

thesan:
See #1447


thesan:
Sorry about that, what happened is that [the last of of #1512](https://github.com/Joystream/pioneer/pull/1512/commits/c85fa9fa838fa8028f6932fb6e694c8543f4201f) was supposed to disable `minRows` for inline editors. I used the `.ck-editor__editable_inline` to target those but this class actually applies to all editor :facepalm:

jodator:
> Sorry about that, what happened is that [the last commit](https://github.com/Joystream/pioneer/pull/1512/commits/c85fa9fa838fa8028f6932fb6e694c8543f4201f) of #1512 was supposed to disable `minRows` for inline editors. I used the `.ck-editor__editable_inline` to target those but this class actually applies to all editors

Oh, that was easy to miss. Thanks for the quick fix :)

PR #1545, opened 2021-09-24
💳 Transaction status notification component

ClumsyVlad:


jodator (approved):
I wanted to give a bigger review, but actually, it looks OK.

I would change how the steps are passed, but your solution is flexible and allows you to add/remove steps in the future 👍

Issue #1388, opened 2021-09-08
Candidacy preview – base component


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=1927%3A101798)

Implement the candidate preview.

- [x] Display base candidate details (membership)
- [x] Display candidate number out of total candidates
- [x] Allow cycling throughout candidates
- [x] The preview should work in "preview mode" as well (link to issue: https://github.com/Joystream/pioneer/issues/1406 )

Note that this view mixes most of the membership preview modal. The existing views should be re-usable in this component.

PR #1554, opened 2021-09-27
🍦 Candidacy preview modal skeleton

p-sad:
See #1388 
I tried to add some sort of a mock data hook for a Candidate object but I wasn't very happy with it because Candidate in the GQL schema isn't completed yet. Instead, I focused on reusing the member preview as best as possible, for now


jodator (approved):
Sorry that it took so long. AFAICS everything is OK 👍

PR #1570, opened 2021-09-29
🦵 Set discussion mode when creating a proposal

### LINKED ISSUE #1509: Check the proposal parameters sent to node
It looks like that some of the general proposal attributes are not sent in the extrinsic.

*   [x] discussion mode & whitelist
*   [x] proposal start trigger

p-sad:
Closes #1509 
Added a transaction for setting the discussion mode; the transaction modal is located in a separate dir called `ChangeThreadMode`, since I expect that a standalone functionality for changing the thread mode will be added as well.


jodator (approved):
Nice 👍

Issue #1389, opened 2021-09-08
Candidacy preview – Candidacy details


jodator:
@[figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=1927%3A101798)

@[figma](https://www.figma.com/file/Z3KcRoAll3JGEY622o0iZu/Council?node-id=104%3A5407)

Fetch and display candidacy details:

- [ ] Title
- [ ] Past election outcomes dashboard
- [ ] Link to candidacy discussion thread 
- [ ] Number of messages and last message details
- [ ] Candidate summary preview (image + markdown)
- [ ] Staked amout
- [ ] Staking account

jodator:
Things left for later are the title & summary are not in the schema yet and must be taken from "note"

PR #1571, opened 2021-09-29
🍸 Fetch candidacy details

p-sad:
See #1389 
Certain fields are missing from the schema, others I'm not yet sure how to look for (past election stats).
These I'm hoping to complete in follow-up PRs

### DIFF
--- a/packages/ui/src/council/types/Candidate.ts
+++ b/packages/ui/src/council/types/Candidate.ts
@@ -9,8 +9,9 @@ export interface Candidate {
   stakingAccount: string
   rewardAccount: string
   stake: BN
-  note: string
   title: string
+  summary: string
+  description: string[]
   member: Member
 }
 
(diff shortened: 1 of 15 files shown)


packages/ui/src/council/types/Candidate.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { CandidateFieldsFragment } from '../queries'
+
+export interface Candidate {
+  id: string
+  stakingAccount: string
+  rewardAccount: string
+  stake: BN
+  note: string
+  title: string
```
jodator, line 13: So the candidacy will have:

*   title
*   summary
*   short description (bullet points, so `string[]`)

The `note` will be something completely different on the GraphQL probably (like `metadata`).

We could either:

*   add the fields below to the `Candidate` type
*   create `CandidateNote` type for the `Candidate.note` field

What will happen later on GraphQL should be contained in casting. Our components should use this type.

jodator (changes requested):
> Certain fields are missing from the schema, others I'm not yet sure how to look for (past election stats).
> These I'm hoping to complete in follow-up PRs

I'd change one thing already – the `note` handling

For other concerns – please look at the designs. If those queries happen to be present somewhere else we probably might need a separate query for those stats.

ps.: I think that @thesan implemented "times being a council" on the member details tab, so you can look there for the query.

jodator:
@p-sad Please rebase on main – will be better to avoid conflicts from my changes.

jodator (approved):
Thanks!

PR #1575, opened 2021-10-01
#️⃣ Numeric ID in application item

ClumsyVlad:


packages/ui/src/working-groups/components/ApplicationsList.tsx
```diff
@@ -47,7 +47,7 @@ const ApplicationListItem = ({ application, past }: { application: WorkingGroupA
     <ApplicationItemWrap past={past}>
       <ApplicationItemInfo>
         <ToggleableItemInfoTop>
-          <ApplicationID title={application.id}>ID: {application.id}</ApplicationID>
+          <ApplicationID title={application.id}>ID: {application.id.replace(/\D/g, '')}</ApplicationID>
```
jodator, line 50: That's something that I have to think about. We could either:

*   use `runtimeId`
*   don't do that

The reason is that the ID is "weird" for applications (AFAIR it is group name + runtime id number).

jodator (commented):
I'll check that later. For now, I'm not 100% sure.

packages/ui/src/working-groups/components/ApplicationsList.tsx
ClumsyVlad: Then in OpeningListItem.tsx and UpcomingOpeningListItem.tsx this part should be fixed too

PR #1579, opened 2021-10-01
🐊 Display previous candidacy stats in Candidacy Preview

p-sad:
See #1389 
Counting failed candidacies is not perfect yet; subtracting successful and withdrawn from total gives a certain approximation, but this way a member's *current* candidacies are also counted as failed.


DevNikita16 (approved):
👍 code looks fine!

Just one catch for candidate **occaecati_aut_31** there is **-1** failed. So either you think it is mocks problem or calculation problem?

PR #1586, opened 2021-10-04
📭 Handle title and bullet points for announce candidacy

### LINKED ISSUE #1404: Announce candidacy – Profile
@[figma](https://www.figma.com/file/Z3KcRoAll3JGEY622o0iZu/Council?node-id=119%3A232)

Handle the Candidate profile step

- [x] Profile title
- [x] Validate max profile title length
- [x] Short description
- [x] Check requirements for short description (points vs free-text).
- [x] Validate max short description length

DevNikita16:
Closes #1404

### DIFF
--- a/packages/ui/src/council/modals/AnnounceCandidacy/components/TitleAndBulletPointsStep.tsx
+++ b/packages/ui/src/council/modals/AnnounceCandidacy/components/TitleAndBulletPointsStep.tsx
@@ -1,8 +1,10 @@
 import React from 'react'
+import * as Yup from 'yup'
 
-import { InputComponent, InputText, InputTextarea } from '@/common/components/forms'
+import { InputComponent, InputText } from '@/common/components/forms'
 import { Row } from '@/common/components/Modal'
 import { RowGapBlock } from '@/common/components/page/PageContent'
+import { useForm } from '@/common/hooks/useForm'
 
 interface TitleAndBulletPointsStepProps {
   title?: string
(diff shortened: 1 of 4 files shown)


packages/ui/src/council/modals/AnnounceCandidacy/components/TitleAndBulletPointsStep.tsx
```diff
@@ -0,0 +1,134 @@
...
+        break
+      case 'bulletPoint1':
+      case 'bulletPoint2':
+      case 'bulletPoint3': {
+        const points = [fields.bulletPoint1, fields.bulletPoint2, fields.bulletPoint3].map((point, index) =>
+          index + 1 === Number(field.replace('bulletPoint', '')) ? value : point
+        )
+        const values = points.filter(
+          (bulletPoint) => bulletPoint && !!bulletPoint?.trim() && isValidString(bulletPoint, 'bulletPoint')
+        ) as string[]
```
DevNikita16, line 56: This is done to avoid empty bullet points

jodator: I think that this should be possible to achieve with `Yup` or by using static validation. I'll see what we can do here with Yup.

I'm worried that this part looks a bit hackish. And some simpler solution should be possible.

packages/ui/src/council/modals/AnnounceCandidacy/components/TitleAndBulletPointsStep.tsx
```diff
@@ -0,0 +1,134 @@
...
+  bulletPoint1?: string
+  bulletPoint2?: string
+  bulletPoint3?: string
+}
+
+type StringType = 'title' | 'bulletPoint'
+
+const typeLength: { [key in StringType]: number } = { title: 60, bulletPoint: 120 }
+
+const FormSchema = Yup.object().shape({})
```
jodator, line 27: I think that in this case, we could actually use `Yup` for validation.

packages/ui/src/council/modals/AnnounceCandidacy/components/TitleAndBulletPointsStep.tsx
```diff
@@ -0,0 +1,134 @@
+import React from 'react'
+import * as Yup from 'yup'
+
+import { InputComponent, InputText } from '@/common/components/forms'
+import { Row } from '@/common/components/Modal'
+import { RowGapBlock } from '@/common/components/page/PageContent'
+import { useForm } from '@/common/hooks/useForm'
```
jodator, line 7: I think that we could re-visit `useForm()` here. It looks like it is not used in 100%

jodator (changes requested):
Let's schedule some pair programming to see if we can improve form validation.

DevNikita16:
> Let's schedule some pair programming to see if we can improve form validation.

Sure, I'll investigate on `Yup` and `useForms()`. Probably some PP will also be useful.

jodator (commented):
Thanks!

PR #1587, opened 2021-10-04
🔀 Move useCouncilSize hook to council

jodator:
Refactor only


thesan (approved):
Sorry for the delay, it's all good !

PR #1592, opened 2021-10-06
⛓ Allow linking to candidate preview

### LINKED ISSUE #1590: Allow linking to candidate preview
This involves creating a new routing, or adding params to an existing routing.
Displaying candidate requires the candidate's ID and the election cycle number.

p-sad:
Closes #1590 
One possible inconsistency I can see with this is that the Elections page can't normally be navigated to if the user isn't connected to Joystream node, but I'm not sure if this can potentially break the page, and how.

### DIFF
--- a/packages/ui/src/council/modals/CandidacyPreview/CandidacyPreview.tsx
+++ b/packages/ui/src/council/modals/CandidacyPreview/CandidacyPreview.tsx
@@ -67,7 +67,7 @@ export const CandidacyPreview = React.memo(() => {
           <CopyButtonTemplate
             square
             size="small"
-            textToCopy={`${window.location.host}/#/council/election?candidate=${candidate?.member.id}&cycle=${modalData.cycleId}`}
+            textToCopy={`${window.location.host}/#/council/election?candidate=${candidate?.id}&cycle=${modalData.cycleId}`}
             icon={<LinkIcon />}
           />
         </SidePaneTopButtonsGroup>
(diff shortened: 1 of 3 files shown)


packages/ui/src/council/modals/CandidacyPreview/CandidacyPreview.tsx
```diff
@@ -67,7 +67,7 @@ export const CandidacyPreview = React.memo(() => {
           <CopyButtonTemplate
             square
             size="small"
-            textToCopy={`${window.location.host}/#/council/election?candidate=${candidate?.member.id}&cycle=${modalData.cycleId}`}
+            textToCopy={`${window.location.host}/#/council/election?candidate=${candidate?.id}&cycle=${modalData.cycleId}`}
```
jodator, line 70: Let's keep only `candidate` id in the link.

jodator (commented):
Should work now, but I think about how to link current & past candidates.

One idea is to have two links:
- `/council/election?candidateId=0`
- `/council/past-election/candidateId=0`
Then we could have a redirect in the modal to the proper page for the candidate.

Piotr's idea was to have a "temporary" page like `council/candidate?id=0` and from that redirect to either election or past eleciton.

p-sad:
What I've decided would be the easiest was making the links always go to `election`, and then the modal can redirect to `past-election`. I'm assuming that in the final version one will be able to browse past elections by cycle ID, so once that's done we could add the candidate's `cycleId` at the end of the redirected address so it redirects to their election too.

PR #1597, opened 2021-10-06
🥠 Improve types of getDataFromEvent function

### LINKED ISSUE #1593: Cannot create a membership
@jodator I can now transfer JOY from the accounts, but when I try create a membership I get:
```
Uncaught TypeError: memberId is undefined   BuyMembershipModal.tsx:46:72
```

_Originally posted by @thesan in https://github.com/Joystream/pioneer/issues/1588#issuecomment-935859243_

jodator:
Closes #1593 

So it looks like the event name has changed. I've fixed that and introduced better types for `getDataFromEvent()`.

Additionally, I fixed some types used in tests as well.


jodator:
@thesan I added the types for events for now in order to be future-proof with that kind of change. In other words, using TypeScript should allow detecting when a particular event is removed or renamed in the future.

PR #1703, opened 2021-10-29
🐊 Validate other staking account

### LINKED ISSUE #1658: Member second candidacy announcement transaction would probably fail
Currently `api.tx.members.confirmStakingAccount` is batch with `api.tx.council.announceCandidacy` in [`AnnounceCandidacyModa`](https://github.com/Joystream/pioneer/blob/4479930d0663f16f93b1f01141da2bf6d12f2765/packages/ui/src/council/modals/AnnounceCandidacy/AnnounceCandidacyModal.tsx#L102).

But `api.tx.members.confirmStakingAccount` apears to fail with: `members.StakingAccountAlreadyConfirm` the second time it's called with the same MemberId. Which would make the whole transaction fail.

jodator:
Closes #1658.


jodator:
I think that the work on this is done, so merging it. I think that I wanted to refactor the stake steps but let's keep that for later.

PR #1707, opened 2021-11-02
♻️ Add Recover vote stake transaction modal

### LINKED ISSUE #1417: Recover stake transaction
@[figma](https://www.figma.com/file/Z3KcRoAll3JGEY622o0iZu/Council?node-id=180%3A16361)

Display the recover stake modal and handle the stake recovering for voting.

Note: This should be already handled by the recover all button on accounts page.

p-sad:
Closes #1417

### DIFF
--- a/packages/ui/src/accounts/components/AccountLockInfo.tsx
+++ b/packages/ui/src/accounts/components/AccountLockInfo.tsx
@@ -3,38 +3,49 @@ import React from 'react'
 import styled from 'styled-components'
 
 import { lockIcon } from '@/accounts/components/AccountLocks'
-import { AccountRow } from '@/common/components/Modal'
+import { TableListItem } from '@/common/components/List'
 import { TokenValue } from '@/common/components/typography'
 import { Colors, Sizes } from '@/common/constants'
 
-import { Account } from '../types'
+import { Account, LockType } from '../types'
 
 import { AccountInfo } from './AccountInfo'
 
 interface SelectedAccountProps {
   account: Account
-  stake: BN
+  amount: BN
+  lockType: LockType
 }
 
-export const AccountLockInfo = ({ account, stake }: SelectedAccountProps) => {
+export const AccountLockInfo = ({ account, amount, lockType }: SelectedAccountProps) => {
   return (
-    <LockInfoContainer>
+    <LockInfoContainer $colLayout={lockInfoLayout}>
       <AccountInfo account={account} />
       <div>
-        {lockIcon('Voting')}
-        Stake from council
+        {lockIcon(lockType)}
+        {lockType} stake
       </div>
-      <TokenValue value={stake} />
+      <TokenValueWrapper>
+        <TokenValue value={amount} />
+      </TokenValueWrapper>
     </LockInfoContainer>
   )
 }
 
-export const LockInfoContainer = styled(AccountRow)`
+export const lockInfoLayout = '3fr 2fr 2fr'
+
+export const LockInfoContainer = styled(TableListItem)`
   padding: 8px 0 8px 16px;
-  grid-template-columns: 1fr 1fr 1fr;
+  width: 100%;
   height: ${Sizes.selectHeight};
   min-height: ${Sizes.selectHeight};
   max-height: ${Sizes.selectHeight};
   background-color: ${Colors.Black[75]};
   border: 1px solid ${Colors.Black[200]};
+`
+
+export const TokenValueWrapper = styled.div`
+  display: flex;
+  justify-content: right;
+  padding-right: 44px;
 `
--- a/packages/ui/src/accounts/components/AccountLockInfo.tsx
+++ b/packages/ui/src/accounts/components/AccountLockInfo.tsx
@@ -11,13 +11,13 @@ import { Account, LockType } from '../types'
 
 import { AccountInfo } from './AccountInfo'
 
-interface SelectedAccountProps {
+export interface AccountLockInfoProps {
   account: Account
   amount: BN
   lockType: LockType
 }
 
-export const AccountLockInfo = ({ account, amount, lockType }: SelectedAccountProps) => {
+export const AccountLockInfo = ({ account, amount, lockType }: AccountLockInfoProps) => {
   return (
     <LockInfoContainer $colLayout={lockInfoLayout}>
       <AccountInfo account={account} />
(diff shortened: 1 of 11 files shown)


packages/ui/src/accounts/components/AccountLockInfo.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { AccountInfo } from './AccountInfo'
+
+interface SelectedAccountProps {
+  account: Account
+  amount: BN
+  lockType: LockType
+}
+
+export const AccountLockInfo = ({ account, amount, lockType }: SelectedAccountProps) => {
```
jodator, line 20: Could you add a storybook for this?

packages/ui/src/council/modals/RecoverVoteStake/RecoverVoteStakeSuccessModal.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <TextMedium margin="l" light>
+          Your stake amount was recovered successfully.
+        </TextMedium>
+      </SuccessModalBody>
+      <ModalFooter>
+        <ButtonGhost size="medium" onClick={() => history.push(ProfileRoutes.profile)}>
+          Your account
+        </ButtonGhost>
+      </ModalFooter>
+      p
```
jodator, line 28: Typo? :stuck\_out\_tongue\_winking\_eye:

packages/ui/src/council/modals/RecoverVoteStake/RecoverVoteStakeModal.tsx
```diff
@@ -0,0 +1,53 @@
...
+  useEffect(() => {
+    if (state.matches('requirementsVerification') && isDefined(feeInfo?.canAfford)) {
+      send(feeInfo?.canAfford ? 'PASS' : 'FAIL')
+    }
+  }, [feeInfo, state.value])
+
+  if (state.matches('success')) {
+    return <RecoverVoteStakeSuccessModal />
+  }
+  if (state.matches('error')) {
```
jodator, line 34:
```suggestion
  }

  if (state.matches('error')) {
```

Just to give a bit structure to file

packages/ui/src/council/modals/RecoverVoteStake/RecoverVoteStakeModal.tsx
```diff
@@ -0,0 +1,53 @@
...
+    )
+  }
+
+  if (!feeInfo || !transaction) return null
+
+  if (state.matches('requirementsFailed')) {
+    return <InsufficientFundsModal onClose={hideModal} address={address} amount={feeInfo.transactionFee} />
+  }
+
+  if (transaction && state.matches('transaction')) {
```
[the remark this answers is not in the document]
jodator: You already have a type guard above, so maybe this one is not needed?

packages/ui/src/council/modals/RecoverVoteStake/index.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { ModalWithDataCall } from '@/common/providers/modal/types'
+import { Address } from '@/common/types'
+
+export * from './RecoverVoteStakeModal'
+export type RecoverVoteStakeModalCall = ModalWithDataCall<
+  'RecoverVoteStake',
+  {
+    address: Address
+    stake: BN
```
jodator, line 11: AFAICS `stake` is not used in the final modal. If so, please remove it from type.

packages/ui/src/council/modals/RecoverVoteStake/RecoverVoteStakeModal.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { RecoverVoteStakeSuccessModal } from './RecoverVoteStakeSuccessModal'
+
+export const RecoverVoteStakeModal = () => {
+  const [state, send] = useMachine(recoverVoteStakeMachine)
+  const { api } = useApi()
+  const {
+    hideModal,
+    modalData: { address },
+  } = useModal<RecoverVoteStakeModalCall>()
+  const transaction = useMemo(() => api?.tx.referendum.releaseVoteStake(), [])
```
jodator, line 23: Probably would not fail, but the deps of `useMemo()` will cause to render the hook only once.

It should be: `[api]`, because it is using the `api` inside memo.

jodator (changes requested):
Looks OK, but some finishing touches are needed.

packages/ui/src/council/modals/RecoverVoteStake/RecoverVoteStakeModal.tsx
krupka4soft: see: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
 
If you specify a list of dependencies as the last argument to useEffect, useLayoutEffect, useMemo, useCallback, or useImperativeHandle, it must include all values that are used inside the callback and participate in the React data flow. That includes props, state, and anything derived from them

packages/ui/src/council/modals/RecoverVoteStake/index.ts
p-sad: `stake` is displayed in the transaction modal as the amount the user is freeing, in the footer and in the lock info.

packages/ui/src/council/modals/RecoverVoteStake/RecoverVoteStakeSignModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+  return (
+    <TransactionModal onClose={hideModal} service={service}>
+      <ModalBody>
+        <TextMedium light>
+          You intend to recover <TokenValue value={stake} /> stake locks from account.
+        </TextMedium>
+
+        <RowGapBlock gap={8}>
+          <ListHeaders $colLayout={lockInfoLayout}>
+            <div />
```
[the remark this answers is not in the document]
p-sad: The `<div />` is there to fill the first column, as only the 2nd and the 3rd have headers in the design. Not sure how to do it better and still use the same `$colLayout` between headers and row.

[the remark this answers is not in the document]
jodator: OK, that's fine. Maybe create a `<EmptyHeader />` component for this (in common, where other list components are present) so we would have a clear understanding of what it is. Even though, it will be an empty `<div />` it will be fluent enough.

jodator (approved):

PR #1710, opened 2021-11-02
🤖 Generate chain specs which force an election stage

thesan:
It needs some documentation.

In the meantime (with all commands ran from the repository root):
1. The chain spec (`packages/ui/dev/chain-spec/data/chain-spec.json`) is at `Announcing` by default. To switch the stage:
   - `yarn workspace ui run set-chain-spec <stage>` (the stages are lowercase)
2. The node can be run either with:
   - docker compose: `docker-compose up node`
   - or `<path to the runtime> --tmp --alice --validator --unsafe-ws-external --unsafe-rpc-external --rpc-cors=all --chain packages/ui/dev/chain-spec/data/chain-spec.json --log runtime`
3. When the runtime is changed the default chain spec might need to be rebuilt (I used docker for this):
   - `docker-compose run --rm build`

Both voting and revealing periods are extra long regardless of the chosen `olympia` runtime thanks to the `referendumInstance1.stage.{STAGE}.started` value.

Currently the docker image is [`joystream/node:9079be466b42a95e18830711ef520c91c614b772`](https://hub.docker.com/layers/joystream/node/9079be466b42a95e18830711ef520c91c614b772/images/sha256-26614a099b3cc35b3471dba1890706c369fe7328d4922783d61602b770e20c86) which might be outdated but it worked fine for me so far.

Finally no Member/Candidate/Vote is set for now.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,4 @@
+COMPOSE_PROJECT_NAME=pioneer-joystream-node
+COMPOSE_FILE=packages/ui/dev/chain-spec/docker-compose.yml
+
+RUNTIME=9079be466b42a95e18830711ef520c91c614b772
--- a/packages/ui/dev/chain-spec/configs/announcing.ts
+++ b/packages/ui/dev/chain-spec/configs/announcing.ts
@@ -1,0 +1,19 @@
+export default {
+  genesis: {
+    runtime: {
+      council: {
+        stage: {
+          stage: {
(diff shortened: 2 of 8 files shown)


jodator (approved):
looks awesome :) I checked how the stage changing works and it is pretty nice. The chain spec worked with the olympia runtime that I have built some time ago.

PR #1716, opened 2021-11-03
🪵 Display councillor vote on proposal page

### LINKED ISSUE #1547: Display Councilor vote on the proposal page
Handle councilor actions & status on the [proposal preview page](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=335%3A42618):

*   [x] Add the "Vote on Proposal" button
    *   [x] It should be disabled with the text "Already voted" when already voted
*   [x] Display "You voted for: ..." on the right side, above the votes section
*   [x] The buttons & vote outcome must be only visible for the active councilor member

Voted state:

Not yet voted state:

p-sad:
Closes #1547


jodator:
I was thinking about whether display the vote even for the past voter. Let say you view the past proposal as well. But maybe it is OK in this form. anyway, looks good.

PR #1717, opened 2021-11-03
📓 Document the chain specs generation

### LINKED ISSUE #1714: Document the chain specs generation
From #1710

thesan:
Closes #1714

### DIFF
--- a/docs/mocks.md
+++ b/docs/mocks.md
@@ -23,6 +23,26 @@ To show help:
 ```shell
 yarn node-mocks --help
 ```
+
+#### Chain spec
+
+Another way to influence the on-chain state for testing purpose, is to provide a customize `chain-spec.json` file when running a Joystream node:
+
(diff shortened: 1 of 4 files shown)


docs/mocks.md
```diff
@@ -23,6 +23,26 @@ To show help:
...
 yarn node-mocks --help
 ```
+
+#### Chain spec
+
+Another way to influence the on-chain state for testing purpose, is to provide a customize `chain-spec.json` file when running a Joystream node:
+
+1. Create `packages/ui/dev/chain-spec/data/chain-spec.json` if it does not exist:
+   - Run `docker-compose run --rm build`
+   - The first time the `chain-spec.json` file is generated this way, the file ownership might have to be fixed.
```
jodator, line 33: Could you add the info about building the chain spec using node as well? This is easy to find in docker-file, but I think it will be easier just to write it here as well.

jodator (commented):
Just one change and ready to merge :)

Thanks!

jodator (approved):

PR #1719, opened 2021-11-04
🍜 Change fetching policy

### LINKED ISSUE #1683: Fetch updates from the query node
Right now we cache everything on the front-end. We only refresh my members but we should refresh everything.

This situation is visible in particular when we add a new entity, like a proposal, or vote, to the system and the state does not refresh.

DevNikita16:
Closes #1683

I've forced watch queries to have `cache-and-network` policy and common queries to have `standby` - seems working nice for our case.

 fetchPolicy determines where the client may return a result from. The options are:
 * cache-first (default): return result from cache. Only fetch from network if cached result is not available.
 * cache-and-network: return result from cache first (if it exists), then return network result once it's available.
 * cache-only: return result from cache if available, fail otherwise.
 * no-cache: return result from network, fail if network call doesn't succeed, don't save to cache
 * network-only: return result from network, fail if network call doesn't succeed, save to cache
 * standby: only for queries that aren't actively watched, but should be available for refetch and updateQueries.

 https://medium.com/@galen.corey/understanding-apollo-fetch-policies-705b5ad71980
 

Also thread posts were not sorted and displayed in random order.
In this PR I've forced them to order by `createdAt` desc

### DIFF
--- a/packages/ui/src/app/providers/QueryNodeProvider.tsx
+++ b/packages/ui/src/app/providers/QueryNodeProvider.tsx
@@ -49,5 +49,18 @@ const getApolloClient = (network: 'local' | 'olympia-testnet') => {
     link: from([errorLink, httpLink]),
     cache: new InMemoryCache(),
     connectToDevTools: true,
+    defaultOptions: {
+      watchQuery: {
+        fetchPolicy: 'cache-and-network',
+        errorPolicy: 'all',
+      },
+      query: {
+        fetchPolicy: 'standby',
+        errorPolicy: 'all',
+      },
+      mutate: {
+        errorPolicy: 'all',
+      },
+    },
   })
 }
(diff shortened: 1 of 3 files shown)


jodator:
> Also thread posts were not sorted and displayed in random order.  
> In this PR I've forced them to order by `createdAt` desc

I see that tests fails because of that probably ;)

packages/ui/src/app/providers/QueryNodeProvider.tsx
```diff
@@ -49,5 +49,18 @@ const getApolloClient = (network: 'local' | 'olympia-testnet') => {
...
+        errorPolicy: 'all',
+      },
+      query: {
+        fetchPolicy: 'standby',
+        errorPolicy: 'all',
+      },
+      mutate: {
+        errorPolicy: 'all',
+      },
+    },
```
jodator, line 64: :rocket:

jodator (approved):

PR #1720, opened 2021-11-04
🐀 Use working group IDs as they appear in the API

### LINKED ISSUE #1560: Add other working groups
Right now only a subset of working groups is set as "known", however, all of them are fetched from the query-node.

In order to do this, we should update how the working groups are handled. Ideally, making TypeScript types use the working group definitions from the polkadot's node API.

Additionally:

*   read reward period from the node's API (const)

p-sad:
Closes #1560 
Also replaced reward period constants with constants fetched from the API.


packages/ui/src/working-groups/types/WorkingGroup.ts
```diff
@@ -1,11 +1,16 @@
+import { ApiRx } from '@polkadot/api'
 import BN from 'bn.js'
 
 import { getAverageStake } from '@/working-groups/model/getAverageStake'
 
 import { WorkingGroupDetailedFieldsFragment, WorkingGroupFieldsFragment } from '../queries'
 
+export type GroupIdName = Extract<
```
jodator, line 8: :rocket: I was searching for this :)

packages/ui/src/working-groups/types/locks.ts
```diff
@@ -1,15 +1,15 @@
...
+import { GroupIdName } from '@/working-groups/types/WorkingGroup'
 
-const GROUP_TO_LOCK_ID: Record<GroupName, LockType> = {
-  forum: 'Forum Worker',
-  'content directory': 'Content Directory Worker',
-  membership: 'Membership Worker',
-  storage: 'Storage Worker',
-  gateway: 'Gateway Worker',
-  operations: 'Operations Worker',
+const GROUP_TO_LOCK_ID: Record<GroupIdName, LockType> = {
```
jodator, line 4: :+1:

jodator (approved):
I briefly scanned all the files, but assuming the main change of types we should be now TS safe in regard to `GrupIdName`. Thanks!

PR #1724, opened 2021-11-05
🏄🏻 Introduce hook to get user onboarding status

### LINKED ISSUE #1563: On-boarding status
Implement the on-boarding status handling.

The on-boarding process requires to detect and handle states:

*   no Polkadot plugin is available
*   no accounts available
*   no JOY tokens on selected account (any account)
*   no membership

Remember, that user might be at any stage of the on-boarding depending on his/her background. For instance, the user that is totally new to Polkadot would not have the extension (step 1 - no extension), but an experienced Polkadot user might have the extension and some accounts already (step 3 - no JOY tokens).

DevNikita16:
Closes #1563 

I will create follow-up to handle free tokens step and status.


jodator (approved):
The new hook looks promising. Waiting for the next steps :)

Issue #1568, opened 2021-09-28
Disable transaction button


jodator:
Because of the transaction status updates (#1558), we need to have a base "transaction button" that all action buttons will use. This enables us to disable any "action" button while another transaction is pending.

The button should:

*   allow any variant (square, ghost, etc)
*   automatically disable itself regardless of `disabled` state when another transaction is pending
*   display a tooltip on hover when disabled by other transactions with a message "please wait..."
    *   optionally check if adding a (?) icon would not cause layout shift

PR #1728, opened 2021-11-05
🥚 Check current transaction status in transaction buttons

p-sad:
See #1568 

We've added a rudimentary transaction context that it controlled by the `observeTransaction` function. While this is not 100% adherent to the function's responsibility, this function is very good at doing this job because it's an asynchronous function and a hook that could otherwise stop being rendered once a modal is closed. The context is very basic at the moment, only exposing information about whether a transaction is currently pending or not. This can be extended by checking the transaction status more accurately at a later stage.

The new transaction button component can't be always used with existing transaction buttons as they sometimes have their own styles and/or dedicated components, but we tried to reuse components as much as possible with the transaction button wrapper.


jodator (changes requested):
I think that PR as-is is OK, but some minimal testing/usage example are welcome:

- storybook (could just simulate a transaction behavior using [controls](https://storybook.js.org/docs/react/essentials/controls))
- a test for the ( disabled ) status for general button

Issue #1566, opened 2021-09-28
Allow user to change height of the editor


jodator:
The markdown editor in various places have a defined min & max height. However, it would be good to allow user to change the editor height using a drag&drop, similar to native textarea controls:

![](https://user-images.githubusercontent.com/247363/135063604-3b1efa53-0087-4a92-aebb-d1137ff1539b.png)

jodator:
ps.: work ok with min & max height:

https://user-images.githubusercontent.com/247363/135065553-caa92e3d-ee98-4858-ac82-909b93e06388.mov

PR #1733, opened 2021-11-07
1566 - Allow user to change height of the editor

oleksanderkorn:
I propose a fix for the Issue https://github.com/Joystream/pioneer/issues/1566

Add `resize: vertical;` to CKEditor to allow changing CKEdit height.
Add `max-height: 800px;` and  `overflow: auto;` to the modal to be able to scroll the modal in case when CKEditor is resized and modal is bigger than the screen.


jodator:
Thank you! I think that this works nice. I tested on the storybook preview and this behaves as expected.

Issue #1480, opened 2021-09-16
Hiring complete is wrongfully displayed


jodator:
The "We are very sorry, you haven’t been chosen." is displayed for the selected member.

PR #1734, opened 2021-11-07
1480 - Hiring complete is wrongfully displayed

oleksanderkorn:
Issue: https://github.com/Joystream/pioneer/issues/1480
Show warning only in case there is an application and a member is not hired.


jodator:
@oleksanderkorn Thank you! I think that this PR doesn't require a test, so I'm merging it.

Issue #1497, opened 2021-09-17
Application side-modal updates


jodator:
<img width="607" alt="Screenshot 2021-09-24 at 15 48 06" src="https://user-images.githubusercontent.com/247363/134685348-224bdcae-352a-4857-a599-cb405339a02f.png">

The Application preview modal should:

*   [ ] The title should be "Application" for other member's applications. The word "My Application" should be used only for applications made by the user controller members (`useIsMyMember()`hook).
*   [ ] Add a link to the opening in the "General Detail" tab (click on this link should close the modal)

dmtrjsg:
Done,checked on vercel env.

PR #1735, opened 2021-11-07
1497 - Application side-modal updates

oleksanderkorn:
Issue https://github.com/Joystream/pioneer/issues/1497

N.B. I'm unable to test this because I need some openings, apply to them, and check cases when my application was hired and was not hired. I cannot do that by myself on Olympia, and if I try to run the node in dev mode according to the docs, UI is not working properly, I see a bunch of errors in console, and UI crashes to a white screen.
```
joystream-node --dev --tmp --unsafe-ws-external --unsafe-rpc-external --rpc-cors=all --log runtime
```

### DIFF
--- a/packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
+++ b/packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
@@ -14,8 +14,8 @@ import {
 import { Tabs } from '@/common/components/Tabs'
 import { useEscape } from '@/common/hooks/useEscape'
 import { useModal } from '@/common/hooks/useModal'
-import { useApplication } from '@/working-groups/hooks/useApplication'
 import { useIsMyMembership } from '@/memberships/hooks/useIsMyMembership'
+import { useApplication } from '@/working-groups/hooks/useApplication'
 
 import { FormDetails } from './FormDetails'
 import { GeneralDetails } from './GeneralDetails'
--- a/packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
+++ b/packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
@@ -15,6 +15,7 @@ import { Tabs } from '@/common/components/Tabs'
 import { useEscape } from '@/common/hooks/useEscape'
 import { useModal } from '@/common/hooks/useModal'
 import { useApplication } from '@/working-groups/hooks/useApplication'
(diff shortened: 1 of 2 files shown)


jodator:
@oleksanderkorn The PR looks OK AFAICS, but could you run locally `yarn lint:fix` form the `ui` package before creating a PR. I'll more info about this in the [Contributing guide](https://github.com/Joystream/pioneer/blob/main/CONTRIBUTING.md). But for now, you often should do:  
 

```
yarn workspace ui lint:fix
yarn workspace ui test
```

To be sure that most of the CI requirements are passing (linter & tests). The `lint:fix` will automatically fix most of the requirements. Any "error"  should be fixed, while "warning" outside the files that you change can be safely ignored.

oleksanderkorn:
@jodator Thanks for the tip, fixed.

oleksanderkorn:
@jodator is there an existing hook for detecting the route has been changed? This is needed to close the modal on the link click. Because i missed that part.

oleksanderkorn:
> @jodator is there an existing hook for detecting the route has been changed? This is needed to close the modal on the link click. Because I missed that part. I guess one option is to store location from `useLocation` and close modal it it has changed.

Found the way history is used to hide the modal:
```
useEffect(
    () =>
      history.listen((location) => {
        if (activeTab === 'FORUM' && location.pathname.startsWith(ForumRoutes.forum)) {
          setHasOverlay(false)
        } else {
          hideModal()
        }
      }),
    []
  )
```

oleksanderkorn:
@jodator tested on netlify, seems to working fine, link works, modal is closed.

packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
```diff
@@ -28,6 +30,8 @@ export const ApplicationDetailsModal = React.memo(() => {
     modalData: { applicationId },
   } = useModal<ApplicationDetailsModalCall>()
   const { isLoading, application } = useApplication(applicationId)
+  const history = useHistory()
+  const isMyMembership = useIsMyMembership(application?.applicant?.id || '')
```
jodator, line 34: I think that we have a bug here. This looks valid from the TS POV, but it does not render correctly since we do not query for that variable. I'll create a separate issue for that.

jodator (approved):
Other than a bug that I've discovered (not related to your PR) it looks OK. Thanks!

packages/ui/src/working-groups/modals/ApplicationDetailsModal/ApplicationDetailsModal.tsx
jodator: See #1742.

PR #1736, opened 2021-11-08
🪄 Show non stored revealed votes in Revealing stage

### LINKED ISSUE #1722: Show non stored revealed votes in Revealing stage
@p-sad I think whenever `voteFor` is available (i.e the vote was revealed) retrieving the vote from the query nodes is the best way.

But when this field is not available the vote can't be shown with the current design. Maybe there could be a way to signal to the user that `N` of it's votes were cast but not reveal-able (in which case there could be an alternate way to reveal those votes, see #1633). 

So I think `votesPerCandidate.myVotes` from `useElectionVotes` should be the voting attempts which match some `CastVote` merged together with the `CastVote` with a set `voteFor` (instead of only being the voting attempts which match some `CastVote`), WDYT ?

_Originally posted by @thesan in https://github.com/Joystream/pioneer/issues/1694#issuecomment-954901545_

thesan:
Closes #1722


p-sad (approved):
Looks good

PR #1738, opened 2021-11-08
🎃 Add recover balance button to my profile

### LINKED ISSUE #790: Handle recoverable balance

jodator:
Closes #790:

*   we merge two stake recovery into one modal with lock parameter
*   we added helpers for `useMyMeberships()` - that introduced some problems with tests, however, looks like usable helper
*   we fetch bound accounts for members so many mocks needed to be regenerated


p-sad (approved):
Looks alright, though one part of the modal got thrown out and I'm not sure whether it was intentional or not.

packages/ui/src/council/modals/RecoverVoteStake/RecoverVoteStakeSignModal.tsx
```diff
@@ -1,87 +0,0 @@
...
-            <ListHeader>Recoverable stake</ListHeader>
-          </ListHeaders>
-          <InputComponent inputSize="l" disabled>
-            <AccountLockInfo
-              account={accountOrNamed(allAccounts, address, 'Account')}
-              amount={stake}
-              lockType={'Voting'}
-            />
-          </InputComponent>
-        </RowGapBlock>
```
p-sad, line 63: I think I didn't submit my review comment; it's this part that didn't make it.

jodator: Oh, we could add this actually instead of what we displayed there. Thanks!

Issue #1562, opened 2021-09-28
On-boarding top bar


jodator:
Implement the onboarding top bar:

![](https://user-images.githubusercontent.com/247363/135056924-53f742ee-fae0-4859-a330-7d376bd4b2eb.png)

The top bar:

*   [x] should be visible only for a non-member
*   [x] should be displayed on every page
*   [x] Should have the show-more section with steps
*   [x] Display on-boarding status (see: )

PR #1740, opened 2021-11-08
👁  On-boarding top bar

stepien4soft:
Closes: #1562


DevNikita16:
👀

packages/ui/src/app/components/OnboardingOverlay/OnBoardingOverlay.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const onBoardingSteps = asOnBoardingSteps(steps, status)
+
+  return (
+    <MainWrapper>
+      <Wrapper>
+        <TextContainer>
+          <TextHuge bold>Become a member</TextHuge>
+          <TextSmall onClick={() => setIsOpen((prev) => !prev)}>
+            Show how <ArrowDownExpandedIcon /> {/*todo add icons and make them rotate on open*/}
```
DevNikita16, line 86: Probably should be handled in this PR :D

packages/ui/test/app/pages/components/OnBoardingOverlay.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+      expect(getByText('How to become a member?')).toBeDefined()
+    })
+
+    it('Install plugin', async () => {
+      useMyAccounts.error = 'EXTENSION'
+
+      const { getByText } = await renderComponent()
+      const pluginCircle = getByText('Add Polkadot plugin')?.parentElement?.previousElementSibling
+
+      expect(pluginCircle).toHaveStyle(`background-color: ${Colors.Blue[500]}`)
```
DevNikita16, line 74: 👍👍

DevNikita16 (changes requested):
Overall looks fine! 🚀 Just handle that icon

packages/ui/src/app/components/OnboardingOverlay/OnBoardingOverlay.tsx
```diff
@@ -0,0 +1,183 @@
...
+    if (index === activeIndex) {
+      return { ...step, type: 'active' }
+    }
+    return { ...step, type: 'next' }
+  })
+}
+
+export const OnBoardingOverlay = () => {
+  const { isLoading, status } = useOnBoardingStatus()
+  const [isOpen, setIsOpen] = useState(false)
```
DevNikita16, line 72: Here we should use `useToggle` hook

WRadoslaw: Fixed

packages/ui/src/app/components/OnboardingOverlay/OnboardingOverlay.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+export default {
+  title: 'App/OnboardingOverlay',
+  component: OnBoardingOverlay,
+} as Meta
+
+const Template: Story = (args) => (
+  <TemplateBlock>
+    <OnBoardingOverlay {...args} />
+  </TemplateBlock>
+)
```
jodator, line 16: I tried to test this on the storybook and, at least for me, it wasn't rendering anything. I would add some controls here to simulate the change in the onboarding status. This way we could test or change the UI quickly.

jodator (changes requested):
The component works very well :) I had some trouble with testing it, but after reading the requirements it was easy :D 

However, I wasn't able to test it using storybook and I think that would be nice to have it. This way we could check or change the UI quicker. I think that it should be doable to add a control to change the onboarding status and simulate the required state for the component using contexts. WDYT?

packages/ui/src/app/components/OnboardingOverlay/OnboardingOverlay.stories.tsx
WRadoslaw: Yeah, I forgot to update story after hook integration, I've take care of it and it should be working now :)

DevNikita16 (approved):

PR #1743, opened 2021-11-09
🎛️ Set candidates and custom stage duration in the chain spec

thesan:
The votes casting sometime works and sometime does not. But it doesn't seems to be link to the chain spec so I will keep looking into it.

Also I wanted to set the duration as a positional option argument but I'm having some issue getting yargs to do what I want !

### DIFF
--- a/docs/mocks.md
+++ b/docs/mocks.md
@@ -36,8 +36,8 @@ Another way to influence the on-chain state for testing purpose, is to provide a
       ```
 
 2. _(optional)_ Change the starting Council/Referendum stage (the default is `Announcing`):
-   - Run `yarn workspace ui run set-chain-spec <stage>` (the stages are lowercase)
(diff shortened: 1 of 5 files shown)


docs/mocks.md
```diff
@@ -36,8 +36,8 @@ Another way to influence the on-chain state for testing purpose, is to provide a
       ```
 
 2. _(optional)_ Change the starting Council/Referendum stage (the default is `Announcing`):
-   - Run `yarn workspace ui run set-chain-spec <stage>` (the stages are lowercase)
-   - Note that the voting and revealing periods are also extended thanks to `referendumInstance1.stage.{STAGE}.started` being set to a future block.
+   - Run `yarn workspace ui run set-chain-spec <stage> [-d <duration>]` (the stages are lowercase)
+   - For the `Voting` and `Revealing` stages the `duration` option set the number of blocks this stage will last.
```
jodator, line 40: :+1:

jodator (approved):
Nice 🚀

PR #1755, opened 2021-11-10
🧑‍🔧 Fix voting buttons

### LINKED ISSUE #1744: Fix vote button on `CandidateCard` not showing

thesan:
Closes #1744


jodator (approved):
Sorry for the late review. AFAICS this change looks OK. I would like to test it together with the query-node locally, but that needs to wait, so I'm merging it right now and do the testing during this week.

PR #1761, opened 2021-11-12
🏗️  Context implementation upon onboarding

WRadoslaw:


packages/ui/src/common/providers/onboarding/types.ts
```diff
@@ -0,0 +1,16 @@
+interface MembershipOnBoarding extends UseMembershipOnBoarding {
+  setFreeTokens?: (value: string) => void
+  isModalOpen: boolean
+  toggleModal: () => void
+}
+
+export interface UseOnBoarding {
+  membership: MembershipOnBoarding
+}
```
jodator, line 9: I think that designing an interface without a clear need should be at least postponed. Right now we only have one concept of onboarding, and I'm not 100% sure that we would have another in the future.

packages/ui/src/common/providers/onboarding/provider.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const [membershipModal, toggleMembershipModal] = useToggle()
+  const [membershipOnBoardingValue, setMembershipOnBoardingValue] = useLocalStorage<string>('onboarding-intro')
+  const membership = useMembershipOnBoarding(membershipOnBoardingValue)
+
+  const value = {
+    membership: {
+      ...membership,
+      setFreeTokens: setMembershipOnBoardingValue,
+      isModalOpen: membershipModal,
+      toggleModal: toggleMembershipModal,
```
jodator, line 25: I think that those two are not related to the onboarding status. So I'd see them outside this hook.

Looking at how the components are used together _it should be_ possible to have the `useToggle()` hook defined in `<MembershipOnBoarding />` or in `<OnBoardingOverlay />`.

jodator (changes requested):
I like the general change of adding the context and wrapping it around the overlay only.

However, I think that introducing multiple onboarings is premature at this stage. It adds complexity to the code, there's more code just to add the `membership` namespaces. ATM I don't see a need for other onboardings. I'm worried that we would stick with the single `membership` onboarding.

Sorry for that change, I think that I didn't fully understand the proposed changes during the F2F talk. However, I'm a fan of not adding code that is not immediatly needed. Also, postponning such refactorings for later ussually gives a better results IMO as we don't always know what it will be needed in the future.

packages/ui/src/common/providers/onboarding/context.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { UseOnBoarding } from '@/common/providers/onboarding/types'
+
+export const OnBoardingContext = createContext<UseOnBoarding>({
+  membership: {
+    isLoading: false,
+    status: undefined,
+    toggleModal: () => {
+      /**/
+    },
```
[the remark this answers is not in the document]
WRadoslaw: Fixed

packages/ui/src/common/providers/onboarding/types.ts
WRadoslaw: Fixed

packages/ui/src/common/providers/onboarding/provider.tsx
WRadoslaw: Transferred to parent component of modal and overlay

jodator:
I forgot that this was based off previous PR. I've rebases this branch onto main so it should be good now. Let's see what the CI would tell

jodator (approved):

PR #1764, opened 2021-11-13
🙈  Hide table headers when there is no past elections

### LINKED ISSUE #1747: Hide list headers in past elections
The list headers on the past elections page should be hidden when there are no data.

WRadoslaw:
Closes #1747


jodator (approved):
Thanks!

PR #1765, opened 2021-11-15
🦀 Add proposal parameters verification

### LINKED ISSUE #1510: Validate proposal fields
The API sets constraints on the proposal properties like:

*   max proposal description length
*   max proposal title length
*   etc

We should handle them on the proposal mode because otherwise the transaction fails.

p-sad:
Closes #1510 
In addition to validating title/description length, the max active proposals number is also checked when displaying the button.


jodator (approved):
Nice 👍

PR #1767, opened 2021-11-15
🎭 Debug revealing stage

### LINKED ISSUE #1745: Check why the voting transaction sometime fails

thesan:
This adds votes to the chain specs, and fix some issues with the chain spec generation.

Closes #1745


p-sad:
Leaving this comment here so we have it on record - the reason this didn't work on local testnet is because the generated chainspec JSON defined council and referendum modules before members. Normally the order of keys in a JSON doesn't make a difference, but the joystream runtime parses the file in order, top to bottom - so the votes were not seeded to the chain because their related member IDs were defined later.

thesan:
@p-sad the `council.announcementPeriodNr` solved it (regardless of the fields order) and AFAICS it even solved the vote casting which sometime mysteriously failed :partying_face:. I also fixed some details like the 0 at `vote_for` and the missing `changed_at`. Thanks again :pray:

jodator (approved):
🚀

Issue #1549, opened 2021-09-24
Vote on proposal modal


jodator:
Add the ["vote on proposal"](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=338%3A12916) modal.

![](https://user-images.githubusercontent.com/247363/134696353-f0b5bd4c-c2e3-4d3b-a264-4a504e5c99e0.png)

The modal:

*   [x] Should validate required stake for transaction
*   [x] Vote Step
    *   [x] Should have 3 buttons: Approve, Reject, Abstain
    *   [x] Only one button can be selected at a time
    *   [x] Selecting Reject should show also the "Slash Proposal" checkbox as well (default "no")
*   [x] Proposal general parameters on the left side (title, rationale)
*   [x] Rationale input
    *   [x] check & validate maximum rationale length
*   [x] Sign and send transaction step

Authorize step:

![](https://user-images.githubusercontent.com/247363/134696496-4428141b-c06d-49f9-957a-234e32dd5352.png)

PR #1769, opened 2021-11-16
🥸 Additional info proposal voting 

stepien4soft:
see: #1549


jodator (approved):
Thanks 👍  Looks OK, so merging it to unblock proposal details PR.

PR #1771, opened 2021-11-16
🦹‍♂️  Action and modal for insufficient worker stake

### LINKED ISSUE #1130: Handle add stake for role
Link to [figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4838976).

On the my roles page it should be possible to add stake to the role's account when the stake is below required.

WRadoslaw:
Closes #1130


packages/ui/src/working-groups/modals/AddStakeModal/machine.ts
```diff
@@ -0,0 +1,59 @@
+// import { DecoratedEvents } from '@joystream/types/augment/augment-api-events'
+import { EventRecord } from '@polkadot/types/interfaces/system'
+import { assign, createMachine } from 'xstate'
```
jodator, line 3:
```suggestion
import { EventRecord } from '@polkadot/types/interfaces/system'
import { assign, createMachine } from 'xstate'
```

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    })
+
+    it('Setup minimal stake', async () => {
+      renderModal()
+
+      const input = await getStakeInput()
+      const submitButton = await getButton('Add Stake')
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      expect(input?.value).toBe('200')
```
jodator, line 83:
```suggestion
      expect(input.value).toBe('200')
```
I would strongly encourage you to not do that. This rule suggest that you might be working with wrong type. The TS ignore is the simplest fix – by not looking that the error exists. That's why we also have the ESLint rule for that ;)

A better solution would be just explicitly tell TS that we're sure that this is a `HTMLInputElement` by type assertion: `const input = (await getStakeInput()) as HTMLInputElement` but fortunately, the testing library got us covered 👇

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      const signButton = await getButton('Sign transaction and Stake')
+      signButton.click()
+    }
+
+    return modal
+  }
+
+  const getStakeInput = () => {
+    return screen.getByTestId('amount-input')
+  }
```
jodator, line 179: ☝️ from the above:

```suggestion
  const getStakeInput = () => {
    return screen.getByTestId<HTMLInputElement>('amount-input')
  }
```

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+  beforeEach(async () => {
+    stubDefaultBalances(api)
+    transfer = stubTransaction(api, 'api.tx.storageWorkingGroup.increaseStake')
+  })
+
+  describe('Form step', () => {
+    it('Renders modal', () => {
+      const { queryByText } = renderModal()
+
+      expect(queryByText('Add stake')).toBeDefined()
```
jodator, line 73:
```suggestion
     renderModal()

      expect(screen.queryByText('Add stake')).toBeDefined()
```
We have some mixup in the tests, but I think that we should stick to one way of querying (by using `screen`). This makes changing query methods faster (no need to update the destructuring object returned from `renderModal()`.

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <MockKeyringProvider>
+        <ApiContext.Provider value={api}>
+          <MockQueryNodeProviders>
+            <ModalContext.Provider value={mockModalContext}>
+              <AddWorkerStakeModal />
+            </ModalContext.Provider>
+          </MockQueryNodeProviders>
+        </ApiContext.Provider>
+      </MockKeyringProvider>
+    )
```
jodator, line 192:
```suggestion
   render(
      <MockKeyringProvider>
        <ApiContext.Provider value={api}>
          <MockQueryNodeProviders>
            <ModalContext.Provider value={mockModalContext}>
              <AddWorkerStakeModal />
            </ModalContext.Provider>
          </MockQueryNodeProviders>
        </ApiContext.Provider>
      </MockKeyringProvider>
    )
```
If the only reason to return render result is to get `get|find|queryBy*()` query, I think that is better to use `screen.get|find|queryBy*()`.

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    const submitButton = await getButton('Add Stake')
+    submitButton.click()
+
+    if (shouldSign) {
+      const signButton = await getButton('Sign transaction and Stake')
+      signButton.click()
+    }
+
+    return modal
+  }
```
jodator, line 175:
```suggestion
    renderModal()

    const submitButton = await getButton('Add Stake')
    submitButton.click()

    if (shouldSign) {
      const signButton = await getButton('Sign transaction and Stake')
      signButton.click()
    }
  }
```
continued on removing return value from render

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+        expect(await screen.queryByText('Failure')).toBeDefined()
+      })
+    })
+  })
+
+  const renderModalAndProceedToSign = async (shouldSign = false) => {
+    const modal = renderModal()
+
+    const submitButton = await getButton('Add Stake')
+    submitButton.click()
```
jodator, line 167: This should rather use testing-library's `fireEvent` because it will do mock DOM interactions rather then calling a `click` on element. See here: https://testing-library.com/docs/guide-events/#interactions-vs-events

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    describe('Pending', () => {
+      it('Renders wait for transaction', async () => {
+        stubTransactionPending(transfer)
+
+        await renderModalAndProceedToSign(true)
+
+        expect(screen.queryByText('Pending transaction')).toBeDefined()
+      })
+    })
```
jodator, line 139: I think that pending step might be omitted. This is a general mechanism, so testing for success/error should be enough.

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      it('Renders transaction failure', async () => {
+        stubTransactionFailure(transfer)
+        await renderModalAndProceedToSign(true)
+
+        expect(await screen.queryByText('Failure')).toBeDefined()
+      })
+    })
+  })
+
+  const renderModalAndProceedToSign = async (shouldSign = false) => {
```
jodator, line 163: The usage outside looks like this:
```ts
renderModalAndProceedToSign(true)
// or:
renderModalAndProceedToSign()
```

It is not totally clear what it will do, a more fluent example would be:

```ts
renderModalAndProceedToSign()
signTransaction()
```
OR
```ts
renderModal()
passStakeStep()
signTransaction()
```

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
jodator: AFAIR, you might need to wrap each `fireEvent.click()` in `act()` as well.

packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRole.tsx
```diff
@@ -156,6 +162,11 @@ export const MyRole = () => {
                       Move Excess Tokens
                     </TransactionButton>
                   ))}
+                {isActive && shouldIncreaseStake && (
+                  <TransactionButton style="primary" size="small" onClick={onAddStakeClick}>
+                    Add Stake
+                  </TransactionButton>
+                )}
```
jodator, line 169: Just a note on modal system. This might need a second thought probably later as we might not followed this rule strictly. If your modal is used in one place (I think that here is such case) you can just use `useToggle()` + render the modal right after the button `{showModal && <AddWorkerStakeModal />`.

Since, you already added this to global modals, there's no need to change it in this PR and I think that using global modal is OK for now.

packages/ui/src/working-groups/modals/AddStakeModal/AddWorkerStakeModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  useEffect(() => {
+    changeField('amount', amount)
+  }, [amount])
+
+  const onSubmit = () => {
+    send({ type: 'DONE', form: fields })
+  }
+
+  if (state.matches('transaction')) {
+    const transaction = api?.tx[group.id].increaseStake(runtimeId, new BN(state.context.form.amount || 0))
```
jodator, line 62: FYI, we have: `getGroup(api, group.id)` which does basically the same. I think that for now, it would be better to stick to the `getGroup(api, group.id)`.

If you think that using `api.tx[group.id]` is better (for me both are OK now), please create a new PR that removes `getGroup()` all together.

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+}
+
+jest.mock('@/accounts/hooks/useMyAccounts', () => {
+  return {
+    useMyAccounts: () => useMyAccounts,
+  }
+})
+
+describe('UI: AddStakeModal', () => {
+  setupMockServer()
```
jodator, line 37: I think that we do not use query-node inside modal. If that's correct then you could remove `setupMockServer()` and query-node context from tests (should be a bit faster after that).

jodator (changes requested):
Overall I think that the PR is OK I added more comments on tests as I found some improvements there.

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
WRadoslaw: Totally forget about type casting. Thanks for improvement!

packages/ui/test/working-groups/modals/AddWorkerStakeModal.test.tsx
WRadoslaw: Changed to fireEvent based click. About wrapping fireEvent with act I found this in testing-library docs `Passes it to the @testing-library/dom fireEvent. It's also wrapped in act so you don't need to worry about doing it.`, so fortunately there is no need to `act` 😅

DevNikita16 (approved):
PR looks good 👏🏼, but from my POV _"Add stake"_ should be _"Increase stake"_.

jodator:
> PR looks good 👏🏼, but from my POV _"Add stake"_ should be _"Increase stake"_.

Just to be sure, you mean the code naming, button titles?

jodator (approved):

DevNikita16:
> > PR looks good 👏🏼, but from my POV _"Add stake"_ should be _"Increase stake"_.
> 
> Just to be sure, you mean the code naming, button titles?

Mostly I was talking about button titles, but code names maybe is one more thing to rename

jodator:
So for me either forms are OK and since it is tie @WRadoslaw can decide :)

From my POV we should follow some rules for naming those and maybe follow name of the main extrinsic should give a clear guidance. So here, if it is `increaseWorkerStake` the modal should be `IncreaseWorkerStakeModal`.

jodator:
Thanks!

PR #1772, opened 2021-11-16
📉 Calculate spent on proposals stat for councils

### LINKED ISSUE #1695: Spend on proposals calculations
It looks that we're unable to calculate those. We need to investigate this further.

See incoming links from other issues for details on where those should be added.

DevNikita16:
Closes #1695 

So from what I get we should calculate _Total Spent On Proposals_ from the approved funding requests' amounts.
Other thing I did in this PR is refactoring hooks for Past Council List, and placing loading items properly


packages/ui/src/council/types/PastCouncil.ts
```diff
@@ -21,21 +23,31 @@ export interface PastCouncilWithDetails extends PastCouncil {
 
 export const asPastCouncil = (fields: PastCouncilFieldsFragment): PastCouncil => ({
   id: fields.id,
-  endedAtBlock: fields.endedAtBlock as number,
+  endedAtBlock: fields.endedAtBlock as number
```
jodator, line 26: That change looks odd. Looks like wrong code style setting in IDE

packages/ui/src/council/queries/council.graphql
```diff
@@ -301,12 +320,21 @@ query GetCouncilBlockRange($where: ElectedCouncilWhereUniqueInput!) {
   }
 }
 
-query GetCouncilProposalsStats($startBlock: Int!, $endBlock: Int!) {
-  approved: proposalExecutedEventsConnection(where: { inBlock_gt: $startBlock, inBlock_lt: $endBlock }) {
+query GetPastCouncilListStats($startBlock: Int!, $endBlock: Int!) {
```
jodator, line 323:
```suggestion
query GetPastCouncilStats($startBlock: Int!, $endBlock: Int!) {
```
WDYT? For me, the `List` part is redundant.

jodator (changes requested):
Overall 👍 

One thing to consider and linter settings to check ;)

jodator (approved):

PR #1776, opened 2021-11-17
🏚 Use a correctly built query in proposal filters

### LINKED ISSUE #1752: Table memberships specified more than once
Some filter combinations, mostly liftetime + proposer, like below:

Returns an error on testnet:

This may be either:

*   error in past proposal filters
*   error in Hydra implemention

Also, check other filters, like:

*   archived forum categories

p-sad:
Closes #1752 

In addition to generating a garbled query, the date ranges specified in `OR` clauses were incorrect. Fixed them up now so they can correctly catch all proposals intersecting the given date range.

More on Hydra's OR clauses and the explanation why the query didn't work can be found in a comment written under the issue.


jodator:
@krupka4soft closing this, but please take a look on that change.

Issue #1260, opened 2021-08-24
Archived `<CategoryListItem />` shouldn't query Moderators


thesan:
Loading the categories is already a bit slow, so since the archived CategoryListItem don't show the moderators it would be better not to query these categories moderators data.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203239107006407) by [Unito](https://www.unito.io)

jodator:
> So the way to fix is to create and use another query in [ForumArchived](https://github.com/oleksanderkorn/pioneer/blob/c42469f5c461f779b29c40377bbb833750395fb5/packages/ui/src/app/pages/Forum/ForumArchived.tsx#L23) that does not fetch moderators, do you agree with that approach?

I think that this might be a good approach. Just be sure that you follow the [query-node guide](https://github.com/Joystream/pioneer/blob/main/docs/README.md#adding-queries). The change should originate from [`council.graphql` file](https://github.com/Joystream/pioneer/blob/main/packages/ui/src/forum/queries/forum.graphql#L148-L152).

oleksanderkorn:
@jodator as i understand from the docs command `yarn run queries:generate` should also generate hooks, but when i add new Fragment to `forum.graphql` and run the rask, i see that it generates new fragments, but there are no new hook i can use:

![image](https://user-images.githubusercontent.com/9252017/142285027-585b6c35-8956-453f-a477-1cabbd8019c0.png)

![image](https://user-images.githubusercontent.com/9252017/142285149-5e22a13a-6baf-4d98-a52d-0eea7b6f3cb2.png)

Should i add the hook manually, or i am doing something wrong?

ivanturlakov:
✅ Tested on https://dao-git-carthage-joystream.vercel.app/#/forum
wss://44.201.112.198.nip.io/ws-rpc

@traumschule @oleksanderkorn  Let me know if there is a specific scenario I need to test

https://user-images.githubusercontent.com/46903215/199476248-0aa3da5d-3f30-4a87-9725-fc1219c567ba.mov

PR #1777, opened 2021-11-17
1260 - Archived <CategoryListItem /> shouldn't query Moderators

oleksanderkorn:
Issue https://github.com/Joystream/pioneer/issues/1260

For Archived Categories moderators array is always empty now:
image

### DIFF
--- a/packages/ui/src/forum/hooks/useForumCategories.ts
+++ b/packages/ui/src/forum/hooks/useForumCategories.ts
@@ -1,29 +1,39 @@
-import { useGetForumCategoriesQuery } from '@/forum/queries'
-import { asForumCategory, CategoryStatusType } from '@/forum/types'
+import { useGetForumCategoriesQuery, useGetArchivedForumCategoriesQuery } from '@/forum/queries'
+import { asForumCategory, asArchivedForumCategory, CategoryStatusType } from '@/forum/types'
 
 interface Props {
   isRoot?: boolean
-  isArchive?: boolean
 }
 
 export const ActiveStatus: CategoryStatusType = 'CategoryStatusActive'
 export const ArchivedStatus: CategoryStatusType = 'CategoryStatusArchived'
 
-export const useForumCategories = ({ isRoot, isArchive }: Props = {}) => {
+export const useForumCategories = ({ isRoot }: Props = {}) => {
   const { loading, data } = useGetForumCategoriesQuery({
     variables: {
       where: {
         ...(isRoot ? { parent: { id_eq: null } } : {}),
-        status_json: { isTypeOf_eq: isArchive ? ArchivedStatus : ActiveStatus },
-        ...(isArchive && !isRoot
-          ? { OR: [{ parent: { id_eq: null } }, { parent: { status_json: { isTypeOf_eq: ActiveStatus } } }] }
-          : {}),
+        status_json: { isTypeOf_eq: ActiveStatus },
+      },
+    },
+  })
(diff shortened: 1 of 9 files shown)


oleksanderkorn:
@jodator please review :)

packages/ui/src/forum/hooks/useForumCategories.ts
```diff
@@ -1,29 +1,39 @@
...
     },
   })
-
   return {
     isLoading: loading,
     forumCategories: data?.forumCategories.map(asForumCategory),
   }
 }
+
+export const useArchivedForumCategories = () => {
```
jodator, line 26: We tend to have one main function/component per file so this might be extracted to `useArchivedForumCategories.ts` together with `ArchiveStatus`.

packages/ui/src/forum/queries/forum.graphql
```diff
@@ -27,6 +27,32 @@ fragment ForumCategoryFields on ForumCategory {
...
+    __typename
+    ... on CategoryStatusArchived {
+      categoryArchivalStatusUpdatedEvent {
+        createdAt
+        inBlock
+        network
+      }
+    }
+  }
+}
```
jodator, line 54: I see that those duplicates a bit with `FourmCategoryBaseFields` so maybe we could change current `ForumCategoryFields`:

*   remove `moderators` from `ForumCategoryBaseFields`
*   remove `ArchivedForumCategoryBaseFields` (it should be the same as `ForumCategoryBaseFields`)
*   add `moderators` to `ForumCategoryFields`

This way we could have the same `Fields` for `ForumCategoryFields` and `ArchivedForumCategoryFields`.

packages/ui/src/forum/types/ForumCategory.ts
```diff
@@ -47,12 +49,33 @@ export const asBaseForumCategory = (fields: ForumCategoryFields): Omit<ForumCate
...
+})
+
 export const asForumCategory = (fields: ForumCategoryFields): ForumCategory => ({
   ...asBaseForumCategory(fields),
   subcategories:
     fields.forumcategoryparent?.map((fields) => ({ ...asSubCategory(fields), status: fields.status.__typename })) ?? [],
 })
 
+export const asArchivedForumCategory = (fields: ArchivedForumCategoryFields): ForumCategory => ({
+  ...asBaseArchivedForumCategory(fields),
```
jodator, line 74: Changing types in GraphQL queries would probably enforce changing some casting here as well.

I'd expect that `asBaseArchivedForumCategory()` would not be needed (the `asBaseForumCategory()` should be enough) and the `moderators` casting will probably be moved to `asForumCategory()` casting.

jodator (changes requested):
Hello @oleksanderkorn, great PR.

I would change:

1. how the base fields are derived for archived and standard category
2.extract a hook to its own file

packages/ui/src/forum/types/ForumCategory.ts
oleksanderkorn: Thanks for the tips, I applied the proposed changes.

packages/ui/src/forum/queries/forum.graphql
oleksanderkorn: Thanks, done.

oleksanderkorn:
@jodator i have applied suggested changes

jodator (approved):
Thanks!

Issue #1633, opened 2021-10-15
Expose and backup voting salt


bedeho:
## Background

Currently, the voting [salt](https://joystream.gitbook.io/joystream-handbook/governance/council#reveal-vote) is generated in the browser and stored in local storage. This is obviously not great, because it means that a browser state reset will cause you to loose your ability to reveal your vote, and you also cannot use a different browser or device.

## Proposal

We should make it possible for the user to backup and import their salt when voting. The backup could be to simply display the salt, or to allow a file download - as those likely have an easier time persisting. There also needs to be an input or import step during revealing.

## User Stories

**Backup salt**
- [ ] On the first vote member is displayed with a salt modal
   - Concept of salt (tooltip) "Voting salt is a random string, generated for a given member id and used for securing the vote to be published without exposing the contents of the vote, and contents to be securely revealed during reveal stage. It is stored locally in the browser and all votes submitted on this device and browser can be revealed using this salt. Voting salt is _required_ to reveal your vote. It can be lost forever, if cookies are cleared. Backup your voting salt to ensure your vote can be revealed later, even from a different browser or device."
   - Input for [meta data ](https://joystream.gitbook.io/joystream-handbook/governance/proposals#vote-1)(pre-populated; proposal; The type of vote; rationale; councilor)
   - Preview of filename (using meta inputs as a mask, editable)
   - "Display" Salt -> so can be copy-pasted ❓ 
    - "Checkbox" to acknowledge that vote can be lost if salt is lost, default value = false
   - "Checkbox" Export Voting Salt file, default set = true
   - CTA "view announcement"
   - 💡  would be great if there is some sort of self-validation in the export format, so that it describes what specific vote, voter member id, in what election and so on, the salt is for, so that when the user tries to import, they are warned if they have the wrong salt, and perhaps explains what it is for, e.g. this salt is not for your membership, this salt is not for this vote, etc.

Figma link to the [Vote Submitted Modal](https://www.figma.com/file/Z3KcRoAll3JGEY622o0iZu/Council?node-id=218%3A32146)

 - [ ] Subsequent votes from same device+browser could have a modification of this window, since salt is already stored by browser and may have been exported.

**Reveal Vote flow when Salt missing**
- [ ] User logged in from a diff browser and device shown an additional modal to import voting salt as part of [reveal vote flow](https://www.figma.com/file/Z3KcRoAll3JGEY622o0iZu/Council?node-id=218%3A27572) 
[...]

PR #1779, opened 2021-11-18
📠 Add a FileInput component

thesan:
See #1633

Here are the wireframes: [@figma](https://www.figma.com/file/Z3KcRoAll3JGEY622o0iZu/Council?node-id=1752%3A36146)
Because only one file is needed to restore the votes I didn't implement the delete file/dust bin button.


ClumsyVlad (approved):
I've added a few fixes and we can merge it 👍🏻

Issue #888, opened 2021-06-24
External link support for tooltips


thesan:
Currently tooltips external `tooltipLinkURL` are not accessible because `<Tooltip />` always renders a react-router `<Link />`.

Fixing this would also remove the need to use `<MemoryRouter />` in stories with an external `tooltipLinkURL`

oleksanderkorn:
Proposed solution: 
If link starts with `/` or `#` using the React Router Link, otherwise use `styled.a`:
![image](https://user-images.githubusercontent.com/9252017/142502834-2b0ccfdd-caf2-4625-82a3-ea6dff1d96a8.png)

Example:
![image](https://user-images.githubusercontent.com/9252017/142502644-0cef7d6c-718b-43fd-9529-2b5335ec9ef1.png)

PR #1789, opened 2021-11-18
888 - External link support for tooltips

oleksanderkorn:
Issue https://github.com/Joystream/pioneer/issues/888

### DIFF
--- a/packages/ui/src/common/components/Tooltip/Tooltip.tsx
+++ b/packages/ui/src/common/components/Tooltip/Tooltip.tsx
@@ -88,7 +88,8 @@ export const Tooltip = ({
     onMouseLeave: mouseLeft,
   }
 
-  const isExternalLink = () => tooltipLinkURL && (tooltipLinkURL.startsWith('#') || tooltipLinkURL.startsWith('/'))
+  const isExternalLink = () =>
+    tooltipLinkURL && (tooltipLinkURL.startsWith('http://') || tooltipLinkURL.startsWith('https://'))
 
   return (
     <TooltipContainer absolute={absolute}>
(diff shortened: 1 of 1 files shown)


oleksanderkorn:
@jodator please review

packages/ui/src/common/components/Tooltip/Tooltip.tsx
```diff
@@ -121,11 +121,16 @@ export const Tooltip = ({
               >
                 {tooltipTitle && <TooltipPopupTitle>{tooltipTitle}</TooltipPopupTitle>}
                 <TooltipText>{tooltipText}</TooltipText>
-                {tooltipLinkURL && (
+                {tooltipLinkURL && (tooltipLinkURL.startsWith('#') || tooltipLinkURL.startsWith('/')) ? (
```
jodator, line 124:
```suggestion
                {tooltipLinkURL && (isExternalLink(tooltipLinkUrl)) ? (
```

To make this condition a bit easier to read I would extract this to a helper function.

I think that there should be additional bracket around both components. ATM if `tooltipLinkURL` is falsy the `<TooltipExternalLink />`  will be rendered.

I think it should be something like:

```tsx
{tooltipLinkUrl && (
  isExternalLink(tooltipLinkURL) ? (
   <TooltipLink /> 
  ) : (
   <TooltipExternalLink /> 
  )
)} // <- two closing brackets
```

jodator (changes requested):
Thank you for the PR 😃 

See my comment for small changes. Overall 👍

packages/ui/src/common/components/Tooltip/Tooltip.tsx
oleksanderkorn: Makes sense, pushed the update.

oleksanderkorn:
Updated the check as well:
```
const isExternalLink = () =>
    tooltipLinkURL && (tooltipLinkURL.startsWith('http://') || tooltipLinkURL.startsWith('https://'))
```

jodator (approved):

jodator:
Thanks!

PR #1799, opened 2021-11-19
🪆 Fix tooltip nesting

### LINKED ISSUE #1753: Invalid DOM nesting of Button + Tooltip
The `<Tooltip />` component uses `<button>` to display the tooltip icon. It will fail in a scenario where a button has a tooltip:

```tsx
<ButtonPrimary>
  Button with Help <Tooltip .../>
</ButtonPrimary>
```

The React dev mode error:

```
Warning: validateDOMNesting(...): <button> cannot appear as a descendant of <button>.
```

I think that we should refactor the Tooltip component to not use `<button>`.

ClumsyVlad:
Closes #1753


ClumsyVlad (commented):
Checked

PR #1801, opened 2021-11-22
🦖  Move Excess Tokens action and TransferModal refactor

### LINKED ISSUE #1129: Move excess tokens do not work
Link to [figma](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4838716).

The "move tokens" button has no action handler. It should allow transferring free tokens (show transfer modal) to other accounts.

### LINKED ISSUE #1797: Infinite loop and runtime error inside TransferFromModal
There is infinite loop inside `TransferFromModal` when there are 2 accounts picked for transfer. Seems like `useTransactionFee` combined with `useEffect` is causing it. 

Additionally when user picks that he want to transfer to given account and selects source/from account that has no balance it will cause white screen.

WRadoslaw:
Closes #1129 
Closes #1797


p-sad (commented):
Looks okay, but don't we need additional tests for the newly introduced props (`minValue`, `transactionFactory`, etc)?
Also, why remove the "Use max" button?

jodator:
> Looks okay, but don't we need additional tests for the newly introduced props (`minValue`, `transactionFactory`, etc)?

:+1: 

> Also, why remove the "Use max" button?

That's a good question. The short answer is that "max" button doesn't work (in regard to transaction fees).

The longer answer: @WRadoslaw please create a follow-up issue that address that and link here. We should have that kind of decisions documented. AFAIR only you & me knows why this button was removed.

WRadoslaw:
Follow-up #1816

jodator (changes requested):
I'm sorry, but after testing the "increase stake" flow I need to ask you to bring back the old modal.

The thing that I did not see earlier is that the previous modal has a bit better UX:
- it was clearly stated that the modal is to "increase stake" and not "transfer tokens" (modal title)
- it was clear what amount to set. With reusing the general purpose transfer modal we removed that validation

image

vs: 

image

Again, I'm sorry for the late change, but I think that it would be easier to revert the old modal than add more features to the transfer modal.

WRadoslaw:
`Again, I'm sorry for the late change, but I think that it would be easier to revert the old modal than add more features to the transfer modal.` 

I think I would still consider making Transfer Modal more generic. With one input I think it will be pretty easy to pass schema by props or even adjust schema base on `minValue` and/or `maxValue`, then UX will improve a lot with text error for input, plus it would require small test refactor to test it. 

`it was clearly stated that the modal is to "increase stake" and not "transfer tokens" (modal title)`

Making modal title generic and testing it, again will require less work and test refactor than bringing back 2 new modals for increase/decrease stake plus tests for each one.

WDYT?

jodator:
> I think I would still consider making Transfer Modal more generic.

That might be OK with me, but remember that we might not gonna need that 😉 

> Making modal title generic and testing it, again will require less work and test refactor than bringing back 2 modals for increase/decrease stake plus tests for each one.

I only want to bring back 1 modal: "Increase stake". The main reason for that is that I think that the design of this modal should be different than the transfer modal.

For now transfer modal is responsible only for transfers and I think that this should stay this way. Because:

1. The UX flow of transfer is a bit different than other modals. Especially for the "success" step which shows a transfer.
2. I forgot also that transfer transaction is not processed by the query-node so it has a different confirmation process (different hooks are used for that). While the "increase stake" should be processed by the query-node to display the proper value of the updated state on the worker's page.

Handling other cases on Transfer modal makes it more complex than needed.

jodator (approved):

jodator (changes requested):
Please check tests. I have failing test locally as well.

PR #1806, opened 2021-11-23
🐙 Update query node schema

jodator:
Updated query-node schema to the latest changes: https://github.com/Joystream/joystream/pull/2672#issuecomment-976153007


packages/ui/test/council/pages/Election.test.tsx
```diff
@@ -40,9 +40,13 @@ jest.mock('../../../src/memberships/hooks/useMemberCandidacyStats', () => ({
 const aliceMemberId = getMember('alice').id
 const bobMemberId = getMember('bob').id
 
+// Easier to test if both ID are the equal
```
thesan, line 43:
```suggestion
// NOTE in these tests, ids of candidates and members/options are used interchangeably for simplicity
```
It's a detail but I got a little confused when I saw that `castVote()` was passed candidate ids for `optionId`s. Maybe this could make it a little clearer WDYT ?

thesan (approved):
Looks great :+1: 

I just have a suggestion about the election test comment

PR #1817, opened 2021-11-24
🏡 Use local query-node instance

jodator:
This is a simple update to the network settings:

- added new network type `local-mocks` (will work as current `local`)
- removes MirageJS mocks from `local` (to enable `query-node` actions)
- updated settings page to handle any network type


DevNikita16 (approved):
Works for me 🙌🏼

Issue #1809, opened 2021-11-23
Cancel Working Group Lead Opening proposal


stepien4soft:
https://joystream.gitbook.io/joystream-handbook/governance/proposals#cancel-working-group-lead-opening

PR #1820, opened 2021-11-24
👁 Cancel working group proposal

stepien4soft:
close: #1809


packages/ui/src/mocks/data/raw/openings.json
```diff
@@ -36,7 +36,7 @@
     "stakeAmount": 6000,
     "rewardPerBlock": 400,
     "version": 1,
-    "type": "LEAD",
+    "type": "LEADER",
```
jodator, line 39: Please add this to the mocks generators as well.

The `all` option should regenerate all mocks now using proper dependencies.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/WorkingGroupLeadOpening/types.ts
```diff
@@ -11,3 +11,7 @@ export interface WorkingGroupAndOpeningDetailsParameters {
   shortDescription?: string
   groupId?: string
 }
+
+export interface CancelWorkingGroupLeadOpeningParameters extends WorkingGroupAndOpeningDetailsParameters {
+  openingId?: string
+}
```
jodator, line 17: This should be a new type. We only need `groupId` and `openingId` for the transaction. The descriptions are for the opening details and not needed for canceling and opening.

jodator (changes requested):
Overall nice job on adding the first proposal without designs :)

Some changes are needed though as the extrinsic would not work.

ps.: I also saw that the wording group ID is not used in old proposals, but I'll fix that in:https://github.com/Joystream/pioneer/issues/1828

PR #1836, opened 2021-11-25
📇 Allow the same address for root and controller accounts

### LINKED ISSUE #1773: Allow the same account for root and controller accounts
When creating membership we disallow the same account if chosen for either root or controller. The runtime allows this so we shouldn't prohibit that.

This should work for:
- buy membership
- invite member

DevNikita16:
Closes #1773


DevNikita16:
I have not found any restrictions for the same account in Invite Member modal

jodator:
@DevNikita16 please check code style settings in IDE it changed too much AFAICS.

Issue #1835, opened 2021-11-25
Display remaining proposal types' details in proposal preview


p-sad:
The importance of details to implement follows order in #1802

### Proposals to implement:
- [x] Signal proposal details
- [x] CancelWorkingGroupLeadOpeningProposalDetails
- [x] FillWorkingGroupLeadOpeningProposalDetails
- [x] TerminateWorkingGroupLeadProposalDetails
- [x] SetWorkingGroupLeadRewardProposalDetails
- [x] UpdateWorkingGroupBudgetProposalDetails
- [x] SetMaxValidatorCountProposalDetails
- [x] SetReferralCutProposalDetails
- [x] SetInitialInvitationCountProposalDetails
- [x] SetInitialInvitationBalanceProposalDetails
- [x] SetCouncilBudgetIncrementProposalDetails
- [x] SetCouncilorRewardProposalDetails
- [x] VetoProposalDetails
- [x] SetMembershipPriceProposalDetails

### Proposal to implement later:

Edit: Moved to: #2037

jodator:
All required details were implemented. Other moved to #2037.

PR #1837, opened 2021-11-26
🍺 Display proposal parameters

p-sad:
See #1835 
This adds parameters display to preview pages of:
* UpdateWorkingGroupBudget
* SetMaxValidatorCount
* FillWorkingGroupLeadOpening
* SetWorkingGroupLeadReward
* TerminateWorkingGroupLead
* SetMembershipPrice
* SetCouncilBudgetIncrement

### DIFF
--- a/packages/ui/src/proposals/helpers/getDetailsRenderStructure.ts
+++ b/packages/ui/src/proposals/helpers/getDetailsRenderStructure.ts
@@ -1,5 +1,6 @@
 import { omit } from 'lodash'
 
+import { isDefined } from '@/common/utils'
 import {
   ProposalDetails,
   ProposalDetailsKeys,
@@ -14,6 +15,7 @@ import {
   OpeningDescriptionDetail,
   AmountDetail,
   DestinationsDetail,
+  ProposalType,
 } from '@/proposals/types'
 
 export type RenderType =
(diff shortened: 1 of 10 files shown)


packages/ui/src/proposals/helpers/getDetailsRenderStructure.ts
```diff
@@ -171,14 +173,18 @@ const mapProposalDetail = (key: ProposalDetailsKeys, proposalDetails: ProposalWi
...
+    updateWorkingGroupBudget: ['group', 'amount'],
   }
 
-  if (proposalDetails.type && definedOrders[proposalDetails.type]) {
-    return definedOrders[proposalDetails.type]
+  if (proposalDetails.type) {
+    const order = definedOrders[proposalDetails.type]
+    if (isDefined(order)) {
+      return order
+    }
```
jodator, line 187: I don't see any change in logic here. I think that the result will be the same.

packages/ui/src/proposals/helpers/getDetailsRenderStructure.ts
```diff
@@ -171,14 +173,18 @@ const mapProposalDetail = (key: ProposalDetailsKeys, proposalDetails: ProposalWi
 }
 
 const getDetailsOrder = (proposalDetails: ProposalDetails): ProposalDetailsKeys[] => {
-  const definedOrders: { [key: string]: ProposalDetailsKeys[] } = {
+  const definedOrders: Partial<Record<ProposalType, ProposalDetailsKeys[]>> = {
```
jodator, line 176: :+1:

packages/ui/src/proposals/types/ProposalDetails.ts
```diff
@@ -13,7 +15,7 @@ interface BaseProposalDetails {
   type: undefined
 }
 
-type ProposalDetailsNew<Type extends string, Details> = { type: Type } & Details
+type ProposalDetailsNew<Type extends ProposalType, Details> = { type: Type } & Details
```
jodator, line 18: It looks like the name was temporary :sweat\_smile:

jodator (approved):
Nice. See comments, but those might be for the next PR

jodator (changes requested):
It looks like tests are failing on the vote for proposal step.

packages/ui/src/proposals/helpers/getDetailsRenderStructure.ts
p-sad: This may be just a VSCode TS engine thing, but for some reason the typeguard used on an object member didn't work well and still saw `definedOrders[proposalDetails.type]` as array or undefined. Maybe because a string literal is used to index the object.

jodator (approved):

PR #1845, opened 2021-11-26
🏄‍♂️ Added .env variables to override Olympia-Testnet addresses

### LINKED ISSUE #1841: Add .env variables to override Olympia-Testnet addresses

krupka4soft:
Closes #1841

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -179,3 +179,12 @@ To add any of the above:
   * For `Alice_Stash` and `Bob_Stash` use `//stash` after name, e.g.: `//Bob//stash`
 
 By default, only `Alice`, `Alice_Stash`, `Bob` and `Bob_Stash` accounts has any funds.
+
+## Using custom addresses to connect with node/query node
+To use custom addresses add the `.env` file in `packages/ui` (example: `packages/ui/.env.example`) and set
+
+1. `REACT_APP_OLYMPIA_TESTNET_NODE_SOCKET` default `wss://olympia-dev.joystream.app/rpc`
+2. `REACT_APP_OLYMPIA_TESTNET_QUERY_NODE` default `https://olympia-dev.joystream.app/query/server/graphql`
+3. `REACT_APP_OLYMPIA_TESTNET_QUERY_NODE_SOCKET` default `wss://olympia-dev.joystream.app/query/server/graphql`
+
+Please remember to restart the webpack process after each change.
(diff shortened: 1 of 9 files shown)


krupka4soft (commented):
self cr done

docs/README.md
```diff
@@ -179,3 +179,12 @@ To add any of the above:
   * For `Alice_Stash` and `Bob_Stash` use `//stash` after name, e.g.: `//Bob//stash`
 
 By default, only `Alice`, `Alice_Stash`, `Bob` and `Bob_Stash` accounts has any funds.
+
+## Using custom addresses to connect with node/query node
```
jodator, line 183: :+1:

packages/ui/src/app/config.ts
```diff
@@ -0,0 +1,6 @@
+export const OLYMPIA_TESTNET_NODE_SOCKET =
+  process.env.REACT_APP_OLYMPIA_TESTNET_NODE_SOCKET || 'wss://olympia-dev.joystream.app/rpc'
+export const OLYMPIA_TESTNET_QUERY_NODE =
+  process.env.REACT_APP_OLYMPIA_TESTNET_QUERY_NODE || 'https://olympia-dev.joystream.app/query/server/graphql'
```
jodator, line 4: Could you check if this couldn't be done differently? AFAIR `webpack.DefinePlugin` will inline those: [https://webpack.js.org/plugins/define-plugin/#usage](https://webpack.js.org/plugins/define-plugin/#usage)

This would require adding types for globals https://github.com/Joystream/pioneer/blob/7e8cdde48c329ae0a502e751fbc3d68b34ed02a8/packages/ui/src/types.d.ts#L2-L3

packages/ui/webpack.config.js
```diff
@@ -33,6 +40,7 @@ module.exports = (env, argv) => {
       ],
     }),
     new ForkTsCheckerWebpackPlugin(),
+    new webpack.DefinePlugin(envVariables),
```
jodator, line 43: See here: [https://github.com/Joystream/pioneer/blob/7661cc48f5ee4afd1b200a2113e77121393c486c/packages/ui/webpack.config.js#L23-L26](https://github.com/Joystream/pioneer/blob/7661cc48f5ee4afd1b200a2113e77121393c486c/packages/ui/webpack.config.js#L23-L26)

We already have webpack define plugins as they acted as a feature flag in the past. Shouldn't this be merged together?

jodator (changes requested):
I think that we could merge `webpack.DefinePlugin()` calls into one.

packages/ui/webpack.config.js
krupka4soft: definitely - done :)

jodator (approved):

RafalPloszka (approved):

PR #1852, opened 2021-11-29
🙈 Hide onboarding when have funds

### LINKED ISSUE #1838: The onboarding actions for members with tokens
The onboarding section should be only enabled when the user has no Tokens on all of the accounts:

There's no need to display onboarding when user had plenty of resources.

Additionally, it looks like the "Add membership" button does nothing now as it is tied to onboarding. This button should always invoke "buy membership" modal.

DevNikita16:
Closes #1838


p-sad (approved):
👍

PR #1858, opened 2021-11-29
🎀 Display proposal parameters in proposal preview

p-sad:
See #1835 

This should take care of the remaining proposals in the "Proposals to implement" portion of the issue. It's mostly queries and object casting, though I've added a couple of new display components. Interestingly, MirageJS seems to vehemently believe that the `proposal` field in `VetoProposalDetails` is an inverse association to the veto proposal, which it is not supposed to be. I'm not sure how it works on the testnet, as there are no veto proposals created yet.

Next steps to take:
* Make sure fields display properly in the Vote modal
* Make an issue for Bedeho and the design team to check if the details currently displayed are sufficient


jodator (changes requested):
Great job 🚀  

I found two kinds of improvements: CSS and linking to entities.

###  1. CSS
Opening description looks odd (the description label is shifted down):



Similar in signal (should be right-aligned I think):



...and proposal link:



### 2. Links

- [ ] The "cancel working group opening" should open the opening preview (or link to).
- [x] The "vet proposal" should open the proposal.

p-sad:
I've changed the padding of `StatsButtonInnerWrapper` to the same value as `StatisticItem`; they look similar now, but content of the button is shifted about 2px down, probably because the button component it's wrapped inside of has inherent padding of some sort. I couldn't find the exact value and didn't want to compensate it through the wrapper's padding, as setting it to "stat item's padding minus one" is hackish magic-number-o-logy.

p-sad:
Also, the proposal element in veto already links to a proposal preview, but because it's the same page, `history.push` will not push an address that is same as the history head.

jodator:
> Also, the proposal element in veto already links to a proposal preview, but because it's the same page, `history.push` will not push an address that is same as the history head.

OH, I see this now:



So this should with proper data (not mocked)

jodator:
@p-sad please resolve conflicts, and merge.

jodator (approved):

PR #1870, opened 2021-12-01
👺 Set membership lead quota proposal

stepien4soft:
closes: #1866


packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/SetMembershipLeadInvitationQuota.tsx
```diff
@@ -0,0 +1,66 @@
...
+        </RowGapBlock>
+      </Row>
+      <Row>
+        <RowGapBlock gap={20}>
+          <InputComponent
+            label="Lead Invitation Quota Amount"
+            tight
+            units="JOY"
+            required
+            disabled={isLoading || !group?.leadId}
```
jodator, line 52: We need some info when this field is disabled because of no membership working group lead.

packages/ui/src/proposals/modals/AddNewProposal/machine.ts
```diff
@@ -637,6 +646,15 @@ export const addNewProposalMachine = createMachine<AddNewProposalContext, AddNew
             },
           },
         },
+        setMembershipLeadInvitationQuota: {
+          on: {
+            SET_AMOUNT: {
+              actions: assign({
+                specifics: (context, event) => ({ ...context.specifics, amount: (event as SetAmountEvent).amount }),
```
jodator, line 653:
```suggestion
                specifics: (context, event) => ({ ...context.specifics, amount: event.amount }),
```

Could you check if the `as` is necessary? It should be inferred from the events

jodator (changes requested):
I think that one minor change is needed: inform the user why the proposal can't be created when there's not Membership WG Lead.

jodator (approved):

Issue #1536, opened 2021-09-23
Council schema upgrade tracking issue


jodator:
Because of council schema is not present on the testnet any query that references the unknown entities will fail.

This includes:

*   Any council requests
*   Detailed information about a member on member modal (See https://github.com/Joystream/pioneer/issues/1442#issuecomment-925652155)

p-sad:
We're also probably going to need a `status` field in `Candidate`. It would allow us to tell apart candidates who
* have been elected
* have withdrawn their candidacy
* have failed an election
* are candidates in the current election

jodator:
@DevNikita16 do you remember where you'd like to add this filed?

thesan:
@DevNikita16 `referendumStageRevealingOptionResults` was remove from the schema so it's now replaced by `ElectionRound.endedAtBlock`, `ElectionRound.endedAtTime` and `ElectionRound.endedAtNetwork`. I think this solves the problem.

PR #1872, opened 2021-12-01
☎️ Update query nodes schema

thesan:
It updates the schema based on [this branch](https://github.com/thesan/joystream/commits/council-candidate-status), which add a `Candidadate.status` field (see https://github.com/Joystream/pioneer/issues/1536#issuecomment-932200873).

Also the `Candidadate.status` fixes the `useMemberCandidacyStats` issue mentioned here: https://github.com/Joystream/joystream/issues/2874#issuecomment-982560576 and this one too: https://github.com/Joystream/pioneer/pull/1579#issue-1013296223.

@jodator there are a lot of diffs in `schema.graphql` is it normal or did I miss a step to generate it properly ?


packages/ui/src/memberships/queries/__generated__/members.generated.tsx
```diff
@@ -82,10 +82,10 @@ export type MemberWithDetailsFieldsFragment = {
 }
 
 export type GetMembersQueryVariables = Types.Exact<{
-  where?: Types.Maybe<Types.MembershipWhereInput>
-  orderBy?: Types.Maybe<Array<Types.MembershipOrderByInput> | Types.MembershipOrderByInput>
-  offset?: Types.Maybe<Types.Scalars['Int']>
-  limit?: Types.Maybe<Types.Scalars['Int']>
+  where?: Types.InputMaybe<Types.MembershipWhereInput>
```
thesan, line 85: Also what causes this `Maybe` -> `InputMaybe` in every `queries/__generated__/xxx.generated.tsx`

jodator:
> @jodator there are a lot of diffs in schema.graphql is it normal or did I miss a step to generate it properly ?

Those look like cleanup. I've checked a few of them at random. It looks like that wasn't used anywhere.

For instance: `AmendConstitutionProposalDetailsCreateInput`, `AmendConstitutionProposalDetailsUpdateInput` & `AmendConstitutionProposalDetailsWhereInput` are not used by other types, mostly the `type Query` don't use them. Those look useless.

packages/ui/src/memberships/queries/__generated__/members.generated.tsx
jodator: It might be this: [https://github.com/dotansimha/graphql-code-generator/pull/6815#issue-1022447896](https://github.com/dotansimha/graphql-code-generator/pull/6815#issue-1022447896) It looks like something changed int the GraphQL code generator and there's a new type. The definition in generated types is:

```typescript
export type Maybe<T> = T | null;
export type InputMaybe<T> = Maybe<T>;
```

PR #1994, opened 2021-12-16
🔧 Fix the PastElections page crashing

thesan:
After an election was completed successfully the PastElections page was crashing, because of a `BN::addn` being used to sum the candidates stakes which are strings.


Qcza (approved):

thesan:
@Qcza sorry I requested a review from the wrong Pawel

PR #2027, opened 2021-12-20
🏴󠁧󠁢󠁥󠁮󠁧󠁿 Add i18next config

Qcza:


packages/ui/src/services/i18n/config.ts
```diff
@@ -0,0 +1,32 @@
...
+const resources: Record<Locales, ResourceLanguage> = {
+  en: dictEN,
+  'en-GB': dictEN,
+  'en-US': dictEN,
+  ru: dictRU,
+  'ru-RU': dictRU,
+}
+
+// eslint-disable-next-line import/no-named-as-default-member
+i18next
```
jodator, line 19: Usually I'd say it looks weird, but that's the usage, so :+1:

jodator (approved):
Overall It looks ok 👍

As for PR naming ;) :

*   We use emoji's characters, not the GH equivalents (🏴󠁧󠁢󠁥󠁮󠁧󠁿 vs `:england:` )
*   We don't use the conventional commit message

Fee free to change the above rules

Issue #1942, opened 2021-12-09
[Bounties page] Filters


dmtrjsg:
### Bounty List view -> Filters

<img width="969" alt="Screenshot 2021-12-09 at 16 34 41" src="https://user-images.githubusercontent.com/93924525/145437423-e301930b-59fc-4434-92ef-a2b6881ef186.png">

**Acceptance criteria:**
- [x] Search returns full match on the Name field
- [x] Search matching tech is inherited from already implemented search (e.g. in Proposals). For matching search results in the list view, only display the matching Bounties cards. 
- [ ] Shows empty search page for no results. [Empty page example Figma link](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=5239%3A137902)
- [x] While waiting for response from QN show "_filtering data_" label [see design link](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=5319%3A140055)
- [x] Filters reset component 
- [x]  Sorting and pagination applies
- [x] Filter by name can be applied on Enter keypress
- [ ] Name search input reset - with a small cross in the field > like in Zenhub search
<img width="288" alt="Screenshot 2021-12-16 at 11 34 35" src="https://user-images.githubusercontent.com/93924525/146364601-9d917958-fcfd-4c64-a645-b1670548caab.png">

- [ ] Multiple select of list options depends on ([ticket link> ](https://github.com/Joystream/pioneer/issues/857))

**⛔Out of Scope:**

 - [ ]  Filter by "Tags"[ link to issue:](https://github.com/Joystream/pioneer/issues/1946)
 - [ ] Highlighting of phrase match in the title
<img width="462" alt="Screenshot 2021-12-15 at 17 45 31" src="https://user-images.githubusercontent.com/93924525/146238329-c580a4f0-def0-452a-8fb4-957a266127a5.png">

### Design:
🎨  [Link to Figma](https://www.figma.com/file/jZwIoaxOkTxNFMJNzssBgv/Bounty?node-id=392%3A252407)

PR #2041, opened 2021-12-23
🥥 Fetch bounties from querynode

### LINKED ISSUE #1952: Fetch bounties from the query-node

RafalPloszka:
Closes #1952 

TBD:
- [ ] add test for hook (after schema is updated and hook is its final form - it can change in BountyList)


packages/ui/src/bounty/hooks/useBounties.ts
```diff
@@ -0,0 +1,28 @@
...
+  perPage?: number
+}
+
+export const useBounties = ({ order, perPage = 10 }: UseBountiesProps) => {
+  const orderBy = toQueryOrderByInput<BountyOrderByInput>(order)
+  const { data: dataCount } = useGetBountiesCountQuery()
+  const totalCountPerPage = dataCount?.bountiesConnection.totalCount
+  const { offset, pagination } = usePagination(perPage, totalCountPerPage ?? 0, [order])
+
+  const { loading, data } = useGetBountiesQuery({
```
WRadoslaw, line 19: Shouldn't we utilize `where` variable for extra filtering when it's already declared inside schema and it will be needed for [Bounties page](https://www.figma.com/file/jZwIoaxOkTxNFMJNzssBgv/Bounty?node-id=429%3A230948)?

WRadoslaw (commented):
I'm curious where do we have specified that empty member/oracle equals council?

Qcza (approved):

packages/ui/src/bounty/queries/bounty.graphql
```diff
@@ -0,0 +1,47 @@
...
+    }
+    ... on BountyFundingPerpetual {
+      target
+    }
+  }
+  workPeriod
+  judgingPeriod
+  stage
+  totalFunding
+  bountyentrybounty {
```
Qcza, line 27: This field has been renamed with PR https://github.com/Joystream/pioneer/pull/2043
```suggestion
  entries {
```

Qcza (changes requested):

RafalPloszka:
> I'm curious where do we have specified that empty member/oracle equals council?

@WRadoslaw Theo told me that: "the council can create bounties in which case their creator will be undefined. The same applies to the oracle and the BountyContribution.contributor", but I don't see this as a perfect solution so I left a comment so it can be easily changed if it's handled better in the schema

packages/ui/src/bounty/hooks/useBounties.ts
RafalPloszka: I was going to add this in Filters #1942

Qcza (approved):

PR #2047, opened 2021-12-28
Feature/add bounty modal

WRadoslaw:


RafalPloszka:
Currently, it's not possible to go back to the previous step.
I can see this is according to the design in Figma, however, I think it would be good to reconsider ti as it's not very user friendly now

packages/ui/src/bounty/modals/AddBountyModal/components/FundingDetailsStep.tsx
```diff
@@ -0,0 +1,164 @@
...
+export interface FundingDetailsStepProps extends Omit<FundingPeriodDetailsContext, keyof GeneralParametersContext> {
+  setFundingMaximalRange: (fundingMaximalRange: BN) => void
+  setFundingMinimalRange: (fundingMinimalRange: BN) => void
+  setCherry: (cherry: BN) => void
+  setFundingPeriodLength: (fundingPeriodLength: BN) => void
+  setFundingPeriodType: (fundingPeriodType: string) => void
+  account?: Address
+}
+
+export const FundingDetailsStep = ({
```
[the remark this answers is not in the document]
RafalPloszka: Ok, I figured it out. It turned out that I don't have enough transferable JOY amount, but:
1. Calculation is probably wrong because on my account I have 70,935 JOY as transferable, but when I console.log `transferable` it's  only 673 JOY and it's a maximum value that I can fill so that _Next step_ button is enabled 
2. There should be a kind of error message when the cherry is too high

packages/ui/src/bounty/modals/AddBountyModal/components/WorkingDetailsStep.tsx
```diff
@@ -0,0 +1,202 @@
...
+              filter={whitelistFilter}
+              onChange={onMemberAdd}
+            />
+          </InputComponent>
+          <WhitelistWrapper>
+            {workingPeriodWhitelist?.map((member) => (
+              <MemberWrapper key={member.id}>
+                <MemberInfo member={member} />
+                <CloseButton onClick={removeMemberFromWhitelist(member)} />
+              </MemberWrapper>
```
RafalPloszka, line 110: I'm not sure if this is related to this PR, but when I click on Member avatar, sidebar with info about them is displayed, but AddBountyModal is closed then

packages/ui/src/bounty/modals/AddBountyModal/components/WorkingDetailsStep.tsx
```diff
@@ -0,0 +1,202 @@
...
+      {workingPeriodStakeAllowance && (
+        <Row>
+          <InputComponent id="field-periodStake" units="JOY" inputSize="m" tight>
+            <InputNumber
+              isTokenValue
+              id="field-periodStake"
+              value={workingPeriodStake?.toString()}
+              onChange={(_, value) => setWorkingPeriodStake(new BN(value))}
+            />
+          </InputComponent>
```
RafalPloszka, line 170: There should be some kind of info about minimum stake requirement

packages/ui/src/bounty/modals/AddBountyModal/components/WorkingDetailsStep.tsx
stepien4soft: It's not connected with this PR. It could be improve in future because this changes is time consuming

RafalPloszka (approved):

PR #2048, opened 2021-12-28
🐸 Announcing stage fixes

p-sad:
See #1953 

I think displaying only active candidacies in the revealing stage does the job with both hiding withdrawn candidacies and hiding previous candidates after a missed election. The filtering takes place in the frontend and not in a query, since the current query is simple and elegant enough (and not really in need of optimization yet).

I wasn't able to find a reason why the last point would happen - from what I could gather from the [runtime code](https://github.com/Joystream/joystream/blob/db3885858a7812377a19390968bdbf65221f0270/runtime-modules/council/src/lib.rs#L1476), there is a check for an existing candidacy. Maybe the announcing period number was wrong on-chain?


DevNikita16 (approved):

DevNikita16:
As far as I don't see the behaviour from the last point, I'll close an issue. 
Close #1953 

Nice job 🚀

PR #2059, opened 2022-01-04
🪒 Fix the events usage in the schema

thesan:
This fixes most of the event relationships, which were many to one instead of one to one.
Also it changes the `BountyEntry.works` into an events. Which allows to get blocks out of those and matches the runtime logic better.


thesan:
@WRadoslaw based on the design I think it would make sense to add something like: `BountyEntry.latestWork`. WDYT ?

WRadoslaw (approved):

PR #2094, opened 2022-01-17
⚗️ Enable filtering by multiple roles

### LINKED ISSUE #857: Apply role filters for member search
Handle the "Apply FIlters" in member search.

RafalPloszka:
Closes #857


thesan (approved):
Sorry for the delay. In the end I wasn't able to add workers to the chain. I created an opening with `yarn workspace ui run node-mocks opening:create` but the applications always fail at the 2nd step (`members.StakingAccountConfirmed`) with: `membershipWorkingGroup.InvalidStakingAccountForMember`

That said the component works well and there is no gql error so it should be merged :+1:

PR #2099, opened 2022-01-19
Feature/bounty works tab

WRadoslaw:


packages/ui/src/bounty/components/BountyExpired/BountyExpired.tsx
```diff
@@ -21,14 +22,21 @@ interface Props {
 export const BountyExpired = ({ bounty }: Props) => {
   const { t } = useTranslation('common')
```
stepien4soft, line 23: i think common is a default value so it's not necessary here

packages/ui/src/bounty/components/tabs/WorkTab.tsx
```diff
@@ -46,24 +79,32 @@ export const WorkTab = ({ bountyId }: Props) => {
...
-            />
-          </InputComponent>
-        </div>
-      </FilterContainer>
+      {(!!works?.length || entrantSearch) && (
+        <FilterContainer>
+          {entrantSearch && (
+            <ResetFilter light onClick={() => setEntrantSearch('')}>
+              <CrossIcon />
+              Clear all filters
```
stepien4soft, line 92: please use i18n here

stepien4soft (approved):

PR #2165, opened 2022-01-31
Feature/submit judgment modal

WRadoslaw:


stepien4soft (approved):

thesan (approved):
```ts
new BTreeMap(registry, 'EntryId', 'OracleWorkEntryJudgment', Map)
```
Is the correct oracle judgment type, but the `packages/types` should be update with a `yarn build`

PR #2193, opened 2022-02-08
👇 Point to custom network endpoints

### LINKED ISSUE #2167: Point to customs query node network from the settings
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201742930737719) by [Unito](https://www.unito.io)

thesan:
Closes #2167


mnaamani:
So I'm getting some mixed results, on firefox it works better than chrome (on chrome I still get a CORS error)
For background I did have to add `Access-Control-Allow-Origin *` header from server side which serves the config.json to get it to work for firefox.

On Firefox however is is behaving strangely. It appears to use proper endpoints for query-node but still tries to connect to `ws://127.0.0.1:9944/` as the rpc endpoint

mnaamani:
> (on chrome I still get a CORS error)

I think it might just chrome has cached the response.

mnaamani:
Another observation in chrome - the settings from config in url don't get applied when landing on the site, but I have to refresh the page to for the new settings to get used. It works correctly on firefox

mnaamani:
Awesome, its working well now in both browsers!

mnaamani (approved):

mnaamani:
Only thing left to improve I would have to say is that the only way after configuring the network, is that its only possible to change the settings by updating the network config url again or if you want to use the hardcoded networking setting you would have manually go editing local storage and deleting the network_config key. So if instead of re-using "olympia-testnet" maybe there should be another "custom-network" choice in the settings page. And finally going one step further, make the network_config editable in that custom-network.

PR #2194, opened 2022-02-08
🦋 Update types and protobuff based on `olympia-playground`

### LINKED ISSUE #2195: Update joystream types for Olympia
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201786375635610) by [Unito](https://www.unito.io)

thesan:
Closes #2195


mnaamani:
LGTM
Screen Shot 2022-02-09 at 12 30 13 AM

bwhm (approved):
:+1:

mnaamani (approved):

thesan:
@mnaamani sorry I did not see your comment earlier. I agree we should find a better way. I tried GitPkg but I had some issues building the both packages. Also I don't think it should always point at the latest version of the packages, because just dealing with polkadot dependencies versions can be tricky and could break Pioneer at any time. So I think some scripts could be a good solution.

PR #2196, opened 2022-02-08
🛠️ Working groups central widget

### LINKED ISSUE #2127: Guest View: Working Groups Central Widget
**AC**

- [ ] Working groups widget

**Designs**

LInk to Figma 

Screenshot 2022-01-21 at 11 26 02

RafalPloszka:
Closes #2127

### DIFF
--- a/packages/ui/test/overview/components/WorkingGroupsOverview.test.tsx
+++ b/packages/ui/test/overview/components/WorkingGroupsOverview.test.tsx
@@ -49,7 +49,10 @@ describe('UI: Working groups overview', () => {
     })
 
     it('Displays number of working groups', async () => {
-      expect((await screen.findByText('workingGroups.workingGroups')).previousSibling?.textContent).toBe('4')
+      const mocksGroupsCount = mockWorkingGroups.length
+      expect((await screen.findByText('workingGroups.workingGroups')).previousSibling?.textContent).toBe(
+        mocksGroupsCount.toString()
+      )
     })
 
     it('Displays number of workers', async () => {


packages/ui/test/overview/components/WorkingGroupsOverview.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+      expect((await screen.findByText('workingGroups.workingGroups')).previousSibling?.textContent).toBe('4')
+    })
+
+    it('Displays number of workers', async () => {
+      expect((await screen.findByText('workingGroups.workers')).previousSibling?.textContent).toBe('2')
+    })
+
+    it('Displays total budget', async () => {
+      expect((await screen.findByText('workingGroups.totalBudget')).previousSibling?.firstChild?.textContent).toBe(
+        '14,097'
```
WRadoslaw, line 58: IMO this approach will cause problems in long term. The issue here is that you expect an outcome based on the `seedWorkingGroups` which populate component with json mock, taking it further I think any working group mock generation will result in different budget for each group, so the test will require a change after it. 

`budget: randomFromRange(1000, 5000),` <- this is how budget is determined in mock generation

I would take approach of mocking the hook and steer data manually.

packages/ui/test/overview/components/WorkingGroupsOverview.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    it('Displays openings number', async () => {
+      expect((await screen.findByText('workingGroups.openings')).lastChild?.textContent).toBe('1')
+    })
+
+    it('Displays opening title', async () => {
+      expect(await screen.findByText('forum Working Group regular')).toBeDefined()
+    })
+
+    it('Displays reward', async () => {
+      expect((await screen.findByText('workingGroups.rewardPerBlock')).previousSibling?.textContent).toBe('2,536')
```
WRadoslaw, line 86: Same as above. Opening generation line -> `rewardPerBlock: randomFromRange(1, 5) * 100`

WRadoslaw (changes requested):
I not sure if displaying number of some groups like workers or openings depends on mocks as well, but it's good to check it and if they are, I might be required to refactor them too.

PR #2197, opened 2022-02-09
Update local network settings faucet url to match playground

mnaamani:
Proposing to update the default local-network settings for faucet url to match the playground settings in: https://github.com/Joystream/joystream/blob/eed6f9ae78d721db625cf00da8798e035a17fce5/docker-compose.yml#L259


thesan (approved):
Thanks

Issue #2223, opened 2022-02-12
Olympia - Initial testing bugs


bwhm:
These issues and bugs have been found while testing #2194, and [this playground](https://github.com/Joystream/joystream/actions/runs/1817886021).

I'll keep the major bugs that needs to be resolved in this issue, and will create another one with smaller ones that are not needed for the release in another one at a later time.

### Membership Balances
The `#/members` sidebar balances are showing 0 for all balances. Presumably still using the "old" types.

### Create Lead Opening Proposal
Tested for `Membership Lead` and `Content Lead` only, but presumably the same for all of them.

Regardless of what is set as the "Create Lead Opening" specific values, eg:
- Staking amount (eg. 50k)
- Leaving unstaking period (eg. 100000)
- Reward amount per Block (eg. 42)
The transaction is broadcast with the first two as 0:

![Screen Shot 2022-02-12 at 13 04 27](https://user-images.githubusercontent.com/35237943/153710601-5952f12c-c059-4819-b2f5-06b0e7e9b2fd.png)

Using the same inputs in a transaction created by a script, or the extrinsics tab in [polkadot-js](https://polkadot.js.org) works fine.
(A small sidenote issue: I can't set the "Leaving unstaking period" longer than 100000 blocks. Any number above that gets reset to 100000.)

If approved, the execution of course fails, as these have minimum values on the runtime. See below:

### Proposal Specific Constraints
This is relevant because these values should be displayed to the user in pioneer, alongside input validation. 

Although I haven't tried anywhere near all the proposal types, I don't think any of the above is present for any of the ones I have tried. Basically, users will have no idea what to set here.

### Applying for Openings
Only tested for `Membership Lead`, but but presumably the same for all WG openings:
When you apply for the Lead role, it looks like another types issue. Despite going through (what I assume) is the right flow, ie. setting the correct staking account, when the extrinsic is sent, it wants to use `5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM` as the staking account.
```
subkey inspect-key 5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM
Public Key URI `5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM` is account:
  Network ID/version: substrate
  Public key (hex):   0x0000000000000000000000000000000000000000000000000000000000000000
[...]

PR #2226, opened 2022-02-14
🧩 Add missing data to the member rows

thesan:
Add the balances, slashed and terminated data to the member, and my member rows

See #2223


thesan:
The ["INVITED" count on the `MyMember` component](https://github.com/Joystream/pioneer/blob/0f194334db6ceed57d710dbb4587de1767f54837/packages/ui/src/memberships/components/MemberListItem/MyMember.tsx#L48) is still missing

WRadoslaw (approved):

PR #2283, opened 2022-02-21
🗑️ Delete forum thread

### LINKED ISSUE #2139: Deleting Forum thread created by member
This should work similarly as the design on the moderator section [@figma](https://www.figma.com/file/xP9qdHWamHvV60fm0vvbOz/Forum?node-id=315%3A56824)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201710119872256) by [Unito](https://www.unito.io)

- [ ]  Let's add a bin icon  - link to Figma added in comments
- [ ] Let's re-use the delete post transaction modal
- [ ] Let's try and reuse "global" sign transaction modal component from "bounties module"

RafalPloszka:
Closes #2139


packages/ui/src/common/components/icons/BinIcon.tsx
```diff
@@ -0,0 +1,22 @@
...
+    viewBox="0 0 24 24"
+    preserveAspectRatio="xMidYMid meet"
+    fill="none"
+    color="currentColor"
+    className={className}
+    onClick={onClick}
+  >
+    <path
+      fill-rule="evenodd"
+      clip-rule="evenodd"
```
thesan, line 17: A bunch of icons have this small problems already (and I keep on forgetting too) but the path attributes should be camel case too. Otherwise reacts just adds a warning in the console

packages/ui/test/forum/modals/DeleteThreadModal.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  beforeEach(async () => {
+    stubDefaultBalances(api)
+  })
+
+  it('Requirements passed', async () => {
+    stubTransaction(api, txPath, 10)
+    renderModal()
+    expect(screen.findByText('modals.deleteThread.description')).toBeDefined()
+    expect(screen.findByText('modals.deleteThread.buttonLabel')).toBeDefined()
+  })
```
thesan, line 97: Lately there has been several issues with modals not passing the right parameters to transactions (which typescript doesn't check). From now on it would be good to test this in every new transaction.

Here I think the best would be to add something like:
```ts
const txMock = api.api.tx.forum.deleteThread as unknown as jest.Mock
...
const [userId, categoryId, threadId, hide] = last(txMock.mock.calls)
expect(userId).toBe(...
```
to this `Requirements passed` test. A more complete example here: [packages/ui/test/bounty/modals/SubmitJudgementModal.test.tsx](https://github.com/Joystream/pioneer/blob/8e50fe7cf87a220a6e76203183530c47bd1e9ad3/packages/ui/test/bounty/modals/SubmitJudgementModal.test.tsx#L127)

thesan (changes requested):
It's good but I think we should really test the transaction parameters starting from now

packages/ui/src/common/components/icons/BinIcon.tsx
RafalPloszka: I fixed it for all icons with incorrect fill-rule/clip-rule

PR #2360, opened 2022-02-28
👁 Bugfix/404 response for election tab

### LINKED ISSUE #2319: [USERSNAP] Election page 404 during Normal Council Period
🆕 

- [x] If there is no election (normal council mode), show past election tab...
- [x] Once election is over, also show past election tab.

stepien4soft:
Close #2319


thesan (changes requested):
Fixing the `electionLink` solves a lot of the issue, but the blank page on `/#/election/` and `/#/election/my-votes` once the election ends is also part of it.
It could be handled with something like:
```ts
history.replace({ pathname: ElectionRoutes.pastElections })
```

Also I don't understand why replacing "There are no candidates yet" by "There is no election now", isn't this showing when there is no candidates ?

thesan:
I've added a way to test the `idle` stage for as long as you want.
If you need this run:
1. `docker-compose run --rm build`
2. `yarn workspace ui helpers setChainSpec -s idle -d 1000` (now the chain spec will keep the council idle for 1000 block)
3. `docker-compose up node`

From there you should be able to work with the council `idle` for 100 minutes, when using the `local-mocks` network.

More info in [docs/mocks.md](https://github.com/Joystream/pioneer/blob/5d1dee7808e4b2a665775fb2b8a75cf7af0f0bd1/docs/mocks.md)

packages/ui/src/app/pages/Election/Election.tsx
```diff
@@ -39,14 +40,15 @@ export const Election = () => {
   const { isLoading: isLoadingElection, election } = useCurrentElection()
   const { isLoading: isLoadingElectionStage, stage: electionStage } = useElectionStage()
   const remainingPeriod = useElectionRemainingPeriod(electionStage)
+  const history = useHistory()
   useCandidatePreviewViaUrlParameter()
```
thesan, line 45: I meant something like:
```ts
useEffect(() => {
 if (electionStage === 'inactive') {
    history.replace(ElectionRoutes.pastElections)
  }
}, [electionStage])
```

dmtrjsg:
Merge Olympia branch into this branch @stepien4soft 
@thesan  to test.

packages/ui/test/council/components/ElectionTabs.test.tsx
```diff
@@ -62,7 +62,7 @@ describe('CouncilTabs', () => {
 
     it('New voting period', async () => {
       window.localStorage.setItem(LocalStorageKey, JSON.stringify('inactive'))
-      stubCouncilAndReferendum(api, 'Election', 'Inactive')
+      stubCouncilAndReferendum(api, 'Announcing', 'Inactive')
```
thesan, line 65: @stepien4soft IMO councilStage: 'Election' and electionStage: 'Inactive' still means "inactive stage" (also I don't think it should happen). If this looks right to you please merge the PR :champagne:

PR #2368, opened 2022-03-01
🤯 bugfix/invited count member query added

### LINKED ISSUE #2227: "INVITED" count on `MyMember` is not implemented
The ["INVITED" count on the `MyMember` component](https://github.com/Joystream/pioneer/blob/0f194334db6ceed57d710dbb4587de1767f54837/packages/ui/src/memberships/components/MemberListItem/MyMember.tsx#L48) is still missing

_Originally posted by thesan in https://github.com/Joystream/pioneer/issues/2226#issuecomment-1039246559_

Invited:

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201871804688104) by [Unito](https://www.unito.io)

stepien4soft:
close #2227


packages/ui/src/memberships/queries/members.graphql
```diff
@@ -140,4 +143,7 @@ query GetMemberRowDetails($workerId_in: [ID!]!) {
   terminatedWorkerEventsConnection(where: { worker: { id_in: $workerId_in } }) {
     totalCount
   }
+  memberInvitedEventsConnection(where: { invitingMember: { id_in: $workerId_in } }) {
+    totalCount
+  }
```
thesan, line 148: This invited count is only used on the my member page and I'm not sure of the cost of an extra event type to query, so I think it would be better to add this in a new query/hook which would also query slashed and terminated

thesan (changes requested):

thesan (approved):
Thanks :+1:

PR #2369, opened 2022-03-01
🐕‍🦺 Improve data refetching

### LINKED ISSUE #2277: Refreshing required.
There's a recurring issue with having to refresh the page to get the latest state.
Not clear whether all aspects of this issue is important enough to warrant inclusion in `olympia` release, but it certainly should be rather high up on the post work if not. 

Here are some examples where it causes a pretty bad user experience, and other where it's just annoying:
1. When you, in pioneer:
   - apply to the council
   - apply on an opening
   - create a proposal
   - vote on a proposal (?)
   - vote on the council (?)
   - create a membership
   - post in proposal discussion #2222  (closing originating issue since captured here)
   - (probably) post on the forum
   - ...
You are shown a box as seen below:
announced

However, you can't actually click any of these boxes as that would require a page refresh, which again would remove the button.
If you just click the box away, you also have to refresh to see something has "happened".

2. Even if you are not the person actually performing the action, the same issue applies. Unless you manually refresh your page (changing the tab is not sufficient) you will not know what has happened.

~3. Adding this for to #2270: Pioneer doesn't warn you when you are out of sync. Although it shouldn't manifest itself in any other way than 1. or 2. in most situations, some edge cases causes further issues. During testing, we "managed" to break the QN processor, and the only way to figure it out was to compare state with polkadot-js. Related use case is covered here, so perhaps worth doing together #2269~ 

~I won't try to come up with a permanent "fix", but a temporary one could be to add a warning whenever the `websocket_rpc` blockheight is way out of sync with the `graphql_server_websocket`. This would tell a user a refresh is needed, and could also imply the query node is down. Example:
`warning: the UI is currently `x` blocks behind. Try refreshing your browser window`~

Moved to Dedicated issue >

RafalPloszka:
Closes #2277 

- [x] apply to the council
- [x] apply on an opening
- [x] ~create a proposal~ (already discussed - without council a proposal is automatically rejected)
- [x] vote on a proposal (?)
- [x] vote on the council (?)
- [x] create a membership
- [x] post in proposal discussion
- [x] post on the forum
- [x] withdraw candidacy (added by me)

### DIFF
--- a/packages/ui/src/app/Providers.tsx
+++ b/packages/ui/src/app/Providers.tsx
@@ -9,6 +9,7 @@ import { KeyringContextProvider } from '@/common/providers/keyring/provider'
 import { ModalContextProvider } from '@/common/providers/modal/provider'
 import { NetworkEndpointsProvider } from '@/common/providers/network-endpoints/provider'
 import { OnBoardingProvider } from '@/common/providers/onboarding/provider'
+import { RefetchProvider } from '@/common/providers/refetch/provider'
 import { TransactionContextProvider } from '@/common/providers/transaction/provider'
 import { MembershipContextProvider } from '@/memberships/providers/membership/provider'
 
@@ -32,8 +33,10 @@ export const Providers = ({ children }: Props) => (
                     <RouteActions>
                       <ModalContextProvider>
                         <OnBoardingProvider>
-                          <GlobalStyle />
-                          {children}
+                          <RefetchProvider>
+                            <GlobalStyle />
+                            {children}
+                          </RefetchProvider>
                         </OnBoardingProvider>
                       </ModalContextProvider>
                     </RouteActions>
--- a/packages/ui/src/app/pages/Forum/ForumThread.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumThread.tsx
@@ -22,6 +22,7 @@ import { ThreadTitle } from '@/forum/components/Thread/ThreadTitle'
 import { WatchlistButton } from '@/forum/components/Thread/WatchlistButton'
 import { ForumRoutes } from '@/forum/constant'
 import { useForumThread } from '@/forum/hooks/useForumThread'
+import { useForumThreadPosts } from '@/forum/hooks/useForumThreadPosts'
 import { ForumPost } from '@/forum/types'
 import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
 
@@ -37,6 +38,9 @@ export const ForumThread = () => {
   const newPostRef = useRef<HTMLDivElement>(null)
   const history = useHistory()
   const [replyTo, setReplyTo] = useState<ForumPost | undefined>()
+
+  // TODO: refactor and figure out better solution
+  const { refetch } = useForumThreadPosts(thread ? thread.id : '-1', { post: null, page: null })
 
   useEffect(() => {
     replyTo && newPostRef.current?.scrollIntoView({ behavior: 'smooth', inline: 'end' })
@@ -114,6 +118,7 @@ export const ForumThread = () => {
           removeReply={() => setReplyTo(undefined)}
           getTransaction={getTransaction}
           replyToLink={`${generatePath(ForumRoutes.thread, { id: thread.id })}?post=${replyTo?.id}`}
+          refetch={refetch}
         />
       )}
     </MainPanel>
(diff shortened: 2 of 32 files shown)


thesan (changes requested):
Thanks for the PR it looks like it's working, but in some places the refetch functions are called directly in the rendering code e.g `refetchElection?.()` in `@/council/modals/VoteForCouncil/VoteForCouncilModal.tsx`, these should really be in a `useEffect` (and we should ensure they don't get rerendered multiple times). In other places refetch are called when closing the success modal e.g in `@/forum/modals/PostActionModal/CreatePostModal/CreatePostModal.tsx`, which is fine but if the QN are slow, refetching as soon as the transaction is successful would be slightly better.

I'm thinking we could reduce the complexity and have a more consistent behavior by adding a context provider (similarly to the `ModalContextProvider`). Here's the general idea:

in `@/common/providers/refetch/context.ts`
```ts
type UseRefetch = [(() => Promise<any>) | undefined, () => void]

const RefetchContext = createContext<UseRefetch>([undefined, () => undefined])
```

in `@/common/providers/refetch/provider.tsx`
```ts
const RefetchProvider = ({ children }) => {
  const [refetch, setRefetch] = useState<UseRefetch[0]>()
  return (
    <RefetchContext.Provider value={[undefined, () => undefined]}>
      {children}
    </RefetchContext.Provider>
  )
}
```

in `@/common/hooks/useRefetch.ts`
```ts
type Action = { type: 'set', payload: UseRefetch[0] } | { type: 'do', payload?: boolean }
const useRefetch = (action: Action) => {
  const [refetch, setRefetch] = useContext(RefetchContext)

  useEffect(() => {
    switch (action.type) {
      case 'set':
        return setRefetch(action.payload)
      case 'do':
        if (action.payload ?? true) {
          refetch()
        }
        break
    }
  }, [action.type, action.payload])

  return setRefetch
}
```

And to use it for example:

in `@/app/pages/Proposals/ProposalPreview.tsx`
```ts
...
  const { isLoading, proposal, refetch } = useProposal(id)
  useRefetch({ type: 'set', payload: refetch })
...

```

in `@/forum/modals/PostActionModal/CreatePostModal/CreatePostModal.tsx`
```ts
...
  const [state, send] = useMachine(defaultTransactionModalMachine)
  useRefetchQueries({ type: 'do', payload: state.matches('success') })
...
```

WDYT ?

RafalPloszka:
> Thanks for the PR it looks like it's working, but in some places the refetch functions are called directly in the rendering code e.g `refetchElection?.()` in `@/council/modals/VoteForCouncil/VoteForCouncilModal.tsx`, these should really be in a `useEffect` (and we should ensure they don't get rerendered multiple times). In other places refetch are called when closing the success modal e.g in `@/forum/modals/PostActionModal/CreatePostModal/CreatePostModal.tsx`, which is fine but if the QN are slow, refetching as soon as the transaction is successful would be slightly better.
> 
> I'm thinking we could reduce the complexity and have a more consistent behavior by adding a context provider (similarly to the `ModalContextProvider`). Here's the general idea:
> 
> in `@/common/providers/refetch/context.ts`
> 
> ```ts
> type UseRefetch = [(() => Promise<any>) | undefined, () => void]
> 
> const RefetchContext = createContext<UseRefetch>([undefined, () => undefined])
> ```
> 
> in `@/common/providers/refetch/provider.tsx`
> 
> ```ts
> const RefetchProvider = ({ children }) => {
>   const [refetch, setRefetch] = useState<UseRefetch[0]>()
>   return (
>     <RefetchContext.Provider value={[undefined, () => undefined]}>
>       {children}
>     </RefetchContext.Provider>
>   )
> }
> ```
> 
> in `@/common/hooks/useRefetch.ts`
> 
> ```ts
> type Action = { type: 'set', payload: UseRefetch[0] } | { type: 'do', payload?: boolean }
> const useRefetch = (action: Action) => {
>   const [refetch, setRefetch] = useContext(RefetchContext)
> 
>   useEffect(() => {
>     switch (action.type) {
>       case 'set':
>         return setRefetch(action.payload)
>       case 'do':
>         if (action.payload ?? true) {
>           refetch()
>         }
>         break
>     }
>   }, [action.type, action.payload])
> 
>   return setRefetch
> }
> ```
> 
> And to use it for example:
> 
> in `@/app/pages/Proposals/ProposalPreview.tsx`
> 
> ```ts
> ...
>   const { isLoading, proposal, refetch } = useProposal(id)
>   useRefetch({ type: 'set', payload: refetch })
> ...
> ```
> 
> in `@/forum/modals/PostActionModal/CreatePostModal/CreatePostModal.tsx`
> 
> ```ts
> ...
>   const [state, send] = useMachine(defaultTransactionModalMachine)
>   useRefetchQueries({ type: 'do', payload: state.matches('success') })
> ...
> ```
> 
> WDYT ?

Great idea, thanks :)

thesan (approved):
Looks good :+1:, please just fix the 2 comments before merging

Issue #2483, opened 2022-03-07
Component for Application Questions (WGL Opening proposal)


dmtrjsg:
## Scope

Create a component for the Application Questions (application form) for opening in question. 

- [x] Max characters in the question = 55 char. 
- [x] Toggle for short/ long questions. By default toggle state = short.
- [ ] Short Answer => Text input (like a field form) 70 chars max.
- [ ] Paragraph = Text Area (rtf); no limit on characters in the answer.
- [x] Add more questions button adds a new line with question name mandatory. 
- [x] User can delete question
- [x] New component for toggle which shows as active in both states

⛔  out of scope:

- [ ] Change order of the questions by drag'n'drop.

Referenced in [Figma here ](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1745%3A937)

Preview:

PR #2506, opened 2022-03-09
👀 Feature/Create reusable input list component

stepien4soft:
close: #2483

### DIFF
--- a/packages/ui/src/common/components/EditableInputList/EditableInputList.tsx
+++ b/packages/ui/src/common/components/EditableInputList/EditableInputList.tsx
@@ -12,14 +12,15 @@ import { TextHuge, TextMedium } from '@/common/components/typography'
 export interface EditableInputListProps {
   title?: string
   buttonText: string
+  onChange: (value: boolean) => void
 }
 
 const EditableInputList: React.FC<EditableInputListProps> = React.memo(({ title, buttonText }) => {
-  const [questionFields, addQuestionField] = useState([{ questionField: '', shortValue: true }])
+  const [questionFields, addQuestionField] = useState([{ questionField: '', shortValue: true, inputNumber: 0 }])
 
   const addFields = () => {
     const values = [...questionFields]
-    values.push({ questionField: '', shortValue: true })
+    values.push({ questionField: '', shortValue: true, inputNumber: 1 })
     addQuestionField(values)
   }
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/common/components/forms/ToggleCheckbox.tsx
```diff
@@ -14,19 +14,28 @@ export interface Props {
...
       onChange(setValue)
     }
   }
-
+  const toggleValidation = switchToggle ? switchToggle : false
   return (
     <ToggleContainer groupDisabled={disabled}>
       <ToggleLabel onClick={onClick(true)}>{trueLabel}</ToggleLabel>
-      <ToggleStyledInput isChecked={checked}>
+      <ToggleStyledInput isChecked={checked} switchToggle={toggleValidation}>
```
thesan, line 38: What about something like that:
```ts
  hasNoOffState = false,
}: Props) {
  const onClick = (setValue: boolean) => () => {
    if (disabled !== true) {
      onChange(setValue)
    }
  }
  return (
    <ToggleContainer groupDisabled={disabled}>
      <ToggleLabel onClick={onClick(true)}>{trueLabel}</ToggleLabel>
      <ToggleStyledInput isChecked={checked} hasNoOffState={hasNoOffState}>
```
I'm not sure so convince by `hasNoOffState` but it's more inline with the other flag names in Pioneer, and  using `switchToggle` and the extra `toggleValidation` is kind of confusing

packages/ui/src/common/components/EditableInputList/EditableInputList.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { TransactionButton } from '@/common/components/buttons/TransactionButton'
+import { InputComponent, InputNotification, InputText, ToggleCheckbox } from '@/common/components/forms'
+import { BinIcon } from '@/common/components/icons'
+import { PlusIcon } from '@/common/components/icons/PlusIcon'
+import { TextHuge, TextMedium } from '@/common/components/typography'
+
+export interface EditableInputListProps {
+  title?: string
+  buttonText: string
+}
```
thesan, line 15: The component works well but unless I'm missing something, to make it usable it should be a [controlled components](https://reactjs.org/docs/forms.html#controlled-components) with a `value` and an `onChange` function, similarly to how the `ToggleCheckbox` is implemented

packages/ui/src/common/components/EditableInputList/EditableInputList.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+const ToggleWrapper = styled.div`
+  display: flex;
+  justify-content: end;
+  padding-right: 35px;
+`
+
+const StyledLabel = styled(TextMedium)`
+  font-weight: 900;
+`
```
thesan, line 124: Please add some vertical padding (or whatever works) between the each question, and between the questions and the add button

thesan (changes requested):
Thanks for the PR. In addition to the 3 changes :point_down:, it would be better to keep the numbers from the design:
Screenshot from 2022-03-09 15-36-19
I think it makes it clearer that these fields are grouped together

thesan (approved):

Issue #2395, opened 2022-03-03
Rework Creating Opening via proposal


dmtrjsg:
## Scope

Rework the WGL opening proposal to incorporate the required inputs as per the JSON enclosed.

**Rework create opening flow to include all inputs that are defined in the JSON (see image).**

Below is the step by step description of the fields required to be specified as part of the Proposal **Specific Parameters**

#### Sub-stage 1
stage name - Working Group & Description
- [ ] **Working Group** drop down selector
> display of the working group as label / pill is captured here:
> - #2448
- [ ] **Opening Title** - max length _55 chars_
> display of this field is captured here:
> -  #2501 
- [ ] **Target number of hires.** Do not show this field in GUI, automatically fill in "1" as value for this parameter. (since there's only one WGL)
- [ ] Add **Short description** field to this step
- [ ] Add **Description** field to this step

#### Sub-stage 2
stage name - Duration & Process
- [ ] Opening Title - ⛔  (not in the Metaprotocol), ignore.
- [ ] Add **Application Duration** component #2484 
- [ ] Add **Application Process** Field (aka Application details, Rich Text Editor; ignore preview and text tabs)
[Reference from Figma> ](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1732%3A34248)

####  Sub-stage 3
stage name - Application Form
- [ ] **Application form** #2483 

####  Sub-stage 4
- [ ]  **Staking and reward** - ✅  all good here, no change required

### **Out of scope:**

- [ ] Target applications & hires. Covered here #2364 

### More info

ℹ️  Miro board with [more details on the ticket  ](https://miro.com/app/board/uXjVOHhsIEg=/)

**JSON** 

The example JSON specifies all specific parameters that define _create opening_ proposal. This impacts create proposal journey, opening display and applicant's view, when application to the opening is started. 

![Screenshot 2022-03-03 at 10.45.56.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/4bbe0368-20ab-4c1d-8592-8ac61a164415)

**Text area examples on the applicant side**

<img width="616" alt="Screenshot 2022-03-03 at 10 48 32" src="https://user-images.githubusercontent.com/93924525/156618369-461ba18f-e43c-4495-bb8b-7e0fd82d6014.png">

_______________

[...]

PR #2524, opened 2022-03-11
New form for apply for role staking step

WRadoslaw:


stepien4soft (approved):

thesan (approved):
Sorry about the `it.only` again :facepalm: 

I just realized this has nothing to do with #2395 so I'm merging it now

PR #2530, opened 2022-03-11
🦮 Add missing refetches

### LINKED ISSUE #2277: Refreshing required.

RafalPloszka:
Closes #2277 

Scope: 
- [ ] fix infinite loop
- [ ] Add proposal
- [ ] Add bounty
- [ ] ...

### DIFF
--- a/packages/ui/src/common/hooks/useRefetch.ts
+++ b/packages/ui/src/common/hooks/useRefetch.ts
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useState } from 'react'
+import { useContext, useEffect } from 'react'
 
 import { RefetchContext, UseRefetch } from '../providers/refetch/context'
 
@@ -6,7 +6,6 @@ type Action = { type: 'set'; payload: UseRefetch[0] } | { type: 'do'; payload?: 
 
 export const useRefetch = (action: Action) => {
   const [refetch, setRefetch] = useContext(RefetchContext)
-  const [refetching, setRefetching] = useState<Promise<any> | undefined>()
 
   useEffect(() => {
     switch (action.type) {
@@ -14,11 +13,9 @@ export const useRefetch = (action: Action) => {
         return setRefetch(action.payload)
       case 'do':
         if (action.payload) {
-          setRefetching(refetch?.())
+          refetch?.()
         }
         break
     }
   }, [action.type, action.payload])
-
-  return refetching
 }
--- a/packages/ui/src/proposals/hooks/useProposals.ts
+++ b/packages/ui/src/proposals/hooks/useProposals.ts
@@ -1,6 +1,7 @@
 import { useMemo } from 'react'
 
 import { ProposalWhereInput } from '@/common/api/queries'
+import { RefetchQuery } from '@/common/types/queries'
 import { proposalStatusToTypename } from '@/proposals/model/proposalStatus'
 import { useGetProposalsQuery } from '@/proposals/queries'
 import { asProposal, Proposal, ProposalStatus } from '@/proposals/types'
@@ -17,6 +18,7 @@ export interface UseProposalsProps {
 interface UseProposals {
   isLoading: boolean
   proposals: Proposal[]
+  refetch: RefetchQuery
 }
 
 export const useProposals = ({ status, filters }: UseProposalsProps): UseProposals => {
@@ -54,10 +56,11 @@ export const useProposals = ({ status, filters }: UseProposalsProps): UseProposa
     return { where }
   }, [status, JSON.stringify(filters)])
 
-  const { loading, data } = useGetProposalsQuery({ variables })
+  const { loading, data, refetch } = useGetProposalsQuery({ variables })
 
   return {
     isLoading: loading,
     proposals: data && data.proposals ? data.proposals.map(asProposal) : [],
+    refetch,
   }
 }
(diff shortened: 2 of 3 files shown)


packages/ui/src/proposals/modals/AddNewProposal/machine.ts
```diff
@@ -896,7 +896,12 @@ export const addNewProposalMachine = createMachine<AddNewProposalContext, AddNew
         onDone: [
           {
             target: 'success',
-            actions: assign({ transactionEvents: (context, event) => event.data.events }),
+            actions: assign({
+              proposalId: (_, event) =>
+                parseInt(
+                  getDataFromEvent(event.data.events, 'proposalsCodex', 'ProposalCreated', 1)?.toString() ?? '-1'
+                ),
```
thesan, line 903:
```suggestion
                Number(getDataFromEvent(events, 'proposalsCodex', 'ProposalCreated') ?? -1),
```

packages/ui/src/bounty/modals/AddBountyModal/machine.ts
```diff
@@ -289,12 +289,15 @@ export const addBountyMachine = createMachine<AddBountyContext, AddBountyEvent,
         onDone: [
           {
             target: [AddBountyStates.success],
-            actions: assign({ transactionEvents: (context, event) => event.data.events }),
+            actions: assign({
+              bountyId: (_, event) =>
+                parseInt(getDataFromEvent(event.data.events, 'bounty', 'BountyCreated', 1)?.toString() ?? '-1'),
```
thesan, line 294:
```suggestion
                Number(getDataFromEvent(events, 'bounty', 'BountyCreated') ?? -1),
```

packages/ui/src/bounty/modals/AddBountyModal/AddBountyModal.tsx
```diff
@@ -43,6 +45,11 @@ export const AddBountyModal = () => {
   const { allAccounts } = useMyAccounts()
   const [state, send, service] = useMachine(addBountyMachine)
   const [isValidNext, setValidNext] = useState(false)
+
+  const { refetch: refetchBounties } = useBounties({ status: 'active' })
+  useRefetch({ type: 'set', payload: refetchBounties })
+  useRefetch({ type: 'do', payload: state.matches(AddBountyStates.success) })
```
thesan, line 51: I approved this way of re-fetching in the previous PR (because I didn't remember how some of these modals where called) but IMO it would be better to call the `useRefetch({ type: 'set', payload: refetch })` inside the components where the entities are listed (here it would be `packages/ui/src/app/pages/Bounty/components/BountiesLayout.tsx`). And just call: `useRefetch({ type: 'do', ... ` on the modals. This entities would be refetched with the correct filters/order in just 1 query.

packages/ui/src/common/hooks/useRefetch.ts
```diff
@@ -1,24 +1,21 @@
...
-  const [refetching, setRefetching] = useState<Promise<any> | undefined>()
 
   useEffect(() => {
     switch (action.type) {
       case 'set':
         return setRefetch(action.payload)
       case 'do':
         if (action.payload) {
-          setRefetching(refetch?.())
+          refetch?.()
```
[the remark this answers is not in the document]
RafalPloszka: There was a problem with an infinite loop because of setting state inside useEffect. Probably it could be fixed quite easily, but it was also kind of misleading naming to me (refetch + refetching) so I left it in a simpler way for now

RafalPloszka:
> Sorry about this never ending issue tired_face there's just one last thing I think we should address here: It's complicated to test currently but (unless I'm missing something about how Apollo caches data) it's important that all the `useRefetch({ type: 'set', payload: refetch })` are called in the files where the `use[Entities]` hooks are used to fetch and display the actually data, instead of in the modals code.

Thanks, nice suggestion, I applied it not only in Proposals and Bounties, but everywhere where it was needed, looks better now

thesan (approved):
Looks great, thanks!

PR #2532, opened 2022-03-11
Add auto-conf NetworkType variant to store custom endpoints config

mnaamani:
Made an attempt to improve on the custom endpoint feature.

Introduced a new network type called 'auto-conf' for which we store the custom endpoint settings. And this can only be set at runtime through the url query `network-config`. No network config url is hard-coded in code or taken from env variables, to allow changes to endpoints to be done dynamically otherwise they are persisted in local storage when first loading the site.

env variables `REACT_APP_OLYMPIA_TESTNET_xxx` only configure the the olympia-testnet. If not all values are provided the olympia-testnet option is not displayed in drop-down on settings page. Also made all var names consistent renamed `REACT_APP_MEMBERSHIP_FAUCET_URL` to `REACT_APP_OLYMPIA_TESTNET_MEMBERSHIP_FAUCET_URL`.

Fallback to local network if network config is invalid or manually removed from local storage.

I tried to test it thoroughly with various failure scenarios, human error etc.

With these changes its easier to configure a single instance of pioneer configured at build time for a specific network, and to point at playgrounds (auto-conf), with option of reverting to olympia-testnet from settings page.

### DIFF
--- a/packages/ui/src/common/providers/network-endpoints/provider.tsx
+++ b/packages/ui/src/common/providers/network-endpoints/provider.tsx
@@ -21,7 +21,8 @@ interface Props {
 export const NetworkEndpointsProvider = ({ children }: Props) => {
   const [network, setNetwork] = useNetwork()
   const [endpoints, setEndpoints] = useState<Partial<NetworkEndpoints>>({})
-  const [storedAutoNetworkConfig, storeAutoNetworkConfig] = useLocalStorage<Partial<NetworkEndpoints>>('auto_network_config')
+  const [storedAutoNetworkConfig, storeAutoNetworkConfig] =
+    useLocalStorage<Partial<NetworkEndpoints>>('auto_network_config')
   const [isLoading, setIsLoading] = useState(false)
 
   const updateNetworkConfig = useCallback(
@@ -41,7 +42,9 @@ export const NetworkEndpointsProvider = ({ children }: Props) => {
           throw new Error('fetched config missing endpoints')
         }
 
-        const shouldUpdateConfig = !objectEquals<Partial<NetworkEndpoints>>(newNetworkConfig)(storedAutoNetworkConfig ?? {})
+        const shouldUpdateConfig = !objectEquals<Partial<NetworkEndpoints>>(newNetworkConfig)(
+          storedAutoNetworkConfig ?? {}
+        )
         const shouldUpdateNetwork = network !== 'auto-conf'
 
         if (shouldUpdateConfig) {
@@ -67,7 +70,7 @@ export const NetworkEndpointsProvider = ({ children }: Props) => {
 
   useEffect(() => {
     const endpoints = pickEndpoints(network, storedAutoNetworkConfig ?? {})
-    if(!endpointsAreDefined(endpoints) && network == 'auto-conf') {
+    if (!endpointsAreDefined(endpoints) && network == 'auto-conf') {
       setNetwork('local')
       setEndpoints(localEndpoints)
     } else {
@@ -90,7 +93,7 @@ const endpointsAreDefined = (endpoints: Partial<NetworkEndpoints>): endpoints is
   Object.values(endpoints).length === 4 && Object.values(endpoints).every(isDefined)
 
 const pickEndpoints = <R extends Partial<NetworkEndpoints>>(network: NetworkType, endpoints: R) => {
-  if(network === 'auto-conf') {
+  if (network === 'auto-conf') {
     // Use the stored config in localstorage, fallback on 'local'
     // If config endpoints are partially configured this will produce mixed results, punn intended.
     return {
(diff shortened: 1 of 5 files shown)


mnaamani:
tried to fix formatting with 

`yarn prettier "./{src,test}/**/*.{ts,tsx,html,graphql}" --write`

But too many files are being modified:

```
	modified:   packages/ui/.storybook/preview.jsx
	modified:   packages/ui/README.md
	modified:   packages/ui/dev/node-mocks/commands/council/vote.ts
	modified:   packages/ui/dev/node-mocks/lib/api.ts
	modified:   packages/ui/dev/query-node-mocks/generators/forum/generateCategories.ts
	modified:   packages/ui/dev/query-node-mocks/generators/forum/generateForumMocks.ts
	modified:   packages/ui/dev/query-node-mocks/generators/generateProposals.ts
	modified:   packages/ui/dev/query-node-mocks/generators/generateWorkers.ts
	modified:   packages/ui/src/app/config/network.ts
	modified:   packages/ui/src/common/providers/network-endpoints/provider.tsx
	modified:   packages/ui/test/__mocks__/i18next.js
	modified:   packages/ui/test/__mocks__/react-i18next.js
```

packages/ui/src/common/providers/network-endpoints/provider.tsx
```diff
@@ -38,14 +38,20 @@ export const NetworkEndpointsProvider = ({ children }: Props) => {
           nodeRpcEndpoint: config['websocket_rpc'],
         }
 
-        const shouldUpdateConfig = !objectEquals<Partial<NetworkEndpoints>>(newNetworkConfig)(storedNetworkConfig ?? {})
-        const shouldUpdateNetwork = network !== 'olympia-testnet'
+        if (!endpointsAreDefined(newNetworkConfig)) {
+          throw new Error('fetched config missing endpoints')
+        }
```
thesan, line 43: This will get caught by the `catch` on line 66 and throw the wrong error. But I shouldn't have put this catch statement this far anyway it should just be to line 33 :smile:

packages/ui/src/common/providers/network-endpoints/provider.tsx
```diff
@@ -55,22 +61,18 @@ export const NetworkEndpointsProvider = ({ children }: Props) => {
...
         const errMsg = `Failed to fetch the network configuration from ${configEndpoint}.`
 
-        if (fallbackOnLocalEndpoints) {
-          setEndpoints(localEndpoints)
-          throw new Error(`${errMsg} Falling back on the local endpoints.`)
-        } else {
-          throw new Error(errMsg)
-        }
+        setEndpoints(localEndpoints)
+        throw new Error(`${errMsg} - Falling back on the local endpoints.`)
```
thesan, line 65: This `setEndpoints` was only supposed to happen when `updateNetworkConfig` was called from the `useEffect` at which point the `window.location.reload()` was unnecessary. But when the endpoints are changed after initializing the api and the QN there was sometime some strange and inconsistent behavior between FF and Chrome. So either a reload might be good here or the endpoints could just remain unchanged like in the original `else` branch. WDYT ?

thesan (commented):
Thanks a lot for this PR especially the extra effort on the doc :pray: (I completely forgot about it last time).

This looks like a lot more predictable logic. I just have 2 small comments on the error cases.
And also it would be good to remove `updateNetworkConfig` from the dependencies of the `useEffect` line 79 (it's really details it wouldn't cause any re-render ATM).

packages/ui/src/common/providers/network-endpoints/provider.tsx
mnaamani: Yes I was in-fact relying on that catch block, and had originally changed the throw in it to include the `${err}`.
But your suggestion is better. Will fix.

packages/ui/src/common/providers/network-endpoints/provider.tsx
mnaamani: Makes sense to just keep the current selection. So I will drop calling setEndpoints(localEndpoints) so avoid need to reload page. I'll test changes overall with Chrome as well since I forgot about this subtle issue.

packages/ui/src/common/providers/network-endpoints/provider.tsx
mnaamani: done in [7e017ba](https://github.com/Joystream/pioneer/pull/2532/commits/7e017baf77ef11cac128aa61725bda531b2bfa9b)

packages/ui/src/common/providers/network-endpoints/provider.tsx
mnaamani: done in [7e017ba](https://github.com/Joystream/pioneer/pull/2532/commits/7e017baf77ef11cac128aa61725bda531b2bfa9b)

mnaamani:
Dropped unused dependency `updateNetworkConfig` for `userEffect()` in https://github.com/Joystream/pioneer/pull/2532/commits/bd75fa83eedf065daeb7e487cbe48ce8b517f26b

thesan (approved):
Thanks

PR #2549, opened 2022-03-14
handle edge case where network selection from localStorage is invalid

mnaamani:
Follow up to https://github.com/Joystream/pioneer/pull/2532

Handle edge case where a network choice stored in localStorage is not valid.

1 - For example the choice was `olympia-testnet` but the build of pioneer did not define the OLYMPIA_TESTNET env vars so it is not included in the NetworkType[] of configured networks. In such case the "Loading Endpoints" page would never transition away.

2 - A future version of pioneer changes list of networks.
3 - Someone making manual / incorrect changes to localStorage values 

In these cases, we fallback to `local` network.

### DIFF
--- a/packages/ui/src/common/hooks/useNetwork.ts
+++ b/packages/ui/src/common/hooks/useNetwork.ts
@@ -1,10 +1,19 @@
-import { NetworkType } from '@/app/config'
+import { NetworkType, configuredNetworks } from '@/app/config'
 
 import { useLocalStorage } from './useLocalStorage'
 
 export const useNetwork = () => {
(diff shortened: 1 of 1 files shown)


thesan:
@mnaamani someone from the Pioneer team experienced this, and I created the #2550 PR within 3 min of this one. I solved the problem by just falling back to the local network whenever the endpoints are not fully defined.

mnaamani:
> @mnaamani someone from the Pioneer team experienced this, and I created the #2550 PR within 3 min of this one. I solved the problem by just falling back to the local network whenever the endpoints are not fully defined.

Yes I like your solution, much simpler :+1:

closing..

PR #2551, opened 2022-03-14
Fix on withdraw contribution modal

WRadoslaw:


packages/ui/src/bounty/modals/SubmitWorkModal/SubmitWorkModal.tsx
```diff
@@ -62,7 +63,11 @@ export const SubmitWorkModal = () => {
     }
   })
   const transaction = useMemo(() => {
-    const entry = modalData.bounty.entries?.find((entry) => entry.worker.id === activeMember?.id)
+    const entry = modalData.bounty?.entries?.find((entry) => entry.worker.id === activeMember?.id)
+    if (!entry) {
+      hideModal()
+    }
```
thesan, line 69: Nice catch! But we shouldn't have side effects in `useMemo`, please move the `hideModal()` in a `useEffect` (and I guess `modalData.bounty?.entries?.find((...` in a diferent `useMemo`)

packages/ui/src/bounty/components/BountyPreviewHeader/BountyPreviewHeader.tsx
```diff
@@ -175,17 +196,11 @@ const SuccessfulStageButtons = React.memo(({ bounty, activeMember, t }: BountyHe
...
-  const userEntry = useMemo(() => bounty.entries?.find((entry) => entry.worker.id === activeMember?.id), [bounty])
+const FailedStageButtons = React.memo(({ bounty, t, userEntry, isContributor }: BountyHeaderButtonsProps) => {
   const hasAnnounced = !!userEntry
   const hasSubmitted = hasAnnounced && userEntry.hasSubmitted
   const hasLost = hasSubmitted && !userEntry.winner && !userEntry.rejected && !bounty.isTerminated
-  const isTerminated = !bounty.isTerminated
+  const isTerminated = !bounty.isTerminated && isContributor
 
   if (!hasAnnounced && !isContributor) {
     return null
```
thesan, line 206: This: `isTerminated = !bounty.isTerminated ...` is kind of confusing :raised_eyebrow: 
What about ?
```suggestion
  if (bounty.isTerminated || !hasAnnounced && !isContributor) {
    return null
```

packages/ui/src/bounty/components/BountyPreviewHeader/BountyPreviewHeader.tsx
```diff
@@ -149,24 +177,17 @@ const getReward = (entry: WorkEntry) => {
...
   return (
     <>
       <ButtonGhost size="large">
         <BellIcon /> {t('common:buttons.notifyAboutChanges')}
       </ButtonGhost>
       {winnerConditions && <ClaimRewardButton bountyId={bounty.id} entryId={entryId} reward={reward} />}
-      {(passed || isContributor) && isContributor ? (
+      {(userEntry?.passed || isContributor) && isContributor ? (
         <WithdrawContributionButton bounty={bounty} />
       ) : (
```
thesan, line 192: I don't think `<WithdrawContributionButton />` should show at all during the successful stage

thesan (changes requested):
Thanks for the PR, I just found a few problems most of them related to the previous logic

packages/ui/src/bounty/queries/bounty.graphql
```diff
@@ -160,3 +160,9 @@ query GetBountyContributors(
     ...BountyContributionFields
   }
 }
+
+query GetContributorWithdrawal($id: ID!) {
+  bountyFundingWithdrawalEventByUniqueInput(where: { id: $id }) {
+    id
+  }
+}
```
thesan, line 168: Sorry about the confusion when I suggested to check the withdrawal event I just meant to add:
```gql
withdrawnInEvent { __typename }
```
in the existing:  [`fragment BountyContributionFields`](https://github.com/Joystream/pioneer/blob/78d6521e4550c1b1813e6004f2af6a459d78d075/packages/ui/src/bounty/queries/bounty.graphql#L82)

This way you can add a `isWithdrawy: !!field.asContributor` in [@/bounty/types/casts](https://github.com/Joystream/pioneer/blob/78d6521e4550c1b1813e6004f2af6a459d78d075/packages/ui/src/bounty/types/casts.ts#L83) and there's no need for an extra query, and adding an extra hook to the codebase.

dmtrjsg:
Add cases for cherry and Funding expired.

In case Funding Expires (zero contributions), the cherry needs to be returned to creator upon withdrawing
In case there's one contributor, cherry goes to that user in full (if bounty fails, nothing returned to creator)

dmtrjsg:
When you create bounty, it creates bounty and dispatches fe event, that is just a cherry, but that is treated as contribution.

Cherry < > Bounty contribution

thesan (approved):

Issue #2511, opened 2022-03-09
[USERSNAP] Working Groups: After a lead has been fired the WG list still shows there is a lead


dmtrjsg:
**Taken at**: https://pioneer-2.vercel.app/#/working-groups
**Fields**:
- Houston, we have a problem.: Working Groups: After a lead has been fired the WG list still shows there is a lead
- Give us some details: SSIA - the box should say &#34;no leader&#34; or something similar
- SSIA - the box should say &#34;no leader&#34; or something similar
- Discrod Handle: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/772095ac-e263-40bd-981b-a552e765e3cb-annotated_c19532da-b681-4193-a907-7f540e7f76c0.png?etag=dc6e24f7d786abea125df756589df18a)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/772095ac-e263-40bd-981b-a552e765e3cb-annotated_c19532da-b681-4193-a907-7f540e7f76c0.png?etag=dc6e24f7d786abea125df756589df18a)*

**Browser**: Chrome 99 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1551x1176
**Labels**: bug-non-critical
**[Open #106 in Usersnap](https://app.usersnap.com/l/feedback/2f6cf5b5-aeab-42c5-abac-7d2c21dfe47c?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

bwhm:
Using the query:
```
query {
  workingGroups  {
    id
    name
    leader {
      membershipId
      isActive
      isLead
      membership {
        id
      }
    }
   }
}
```
The example output below shows the difference between a Lead that was terminated, and one that is still active:
```
      {
        "id": "forumWorkingGroup",
        "name": "forumWorkingGroup",
        "leader": {
          "membershipId": "62",
          "isActive": false,
          "isLead": true,
          "membership": {
            "id": "62"
          }
        }
      },
      {
        "id": "membershipWorkingGroup",
        "name": "membershipWorkingGroup",
        "leader": {
          "membershipId": "64",
          "isActive": true,
[...]

PR #2602, opened 2022-03-21
🥸bugfix/lead validation fixed after lead has been fired

stepien4soft:
Close: #2511


thesan (approved):
Looks good :+1:

WRadoslaw:
> @RadoslawWyszynski @stepien4soft I just seeded `isActive` following the worker status (just to keep a consistent UI) and somehow that fixed the tests. So either there is some logic somewhere I don't know about which needs consistency between the worker status and `isActive` or it's just luck (because I didn't have to regenerate the mocks)

Fun fact - it still fails on my local :D

WRadoslaw:
> > Fun fact - it still fails on my local :D
> 
> I've observe my IDE sometime caches part of the build (especially when switching branches a lot). Does it also fail when you run them from a terminal ?

Unfortunately no :c

Issue #2502, opened 2022-03-09
Rename WG names


WRadoslaw:
### Context
----
![image.png](https://images.zenhubusercontent.com/61b33b1e8ddbc86c62bdcd28/d8d85900-f825-43e4-bb94-96dac97cf3a1)

dmtrjsg:
@stepien4soft just for convenience listing all proposal types to cover.. In addition to Proposals:
-  Create WGL opening 
- Fill WGL opening
- Slash Working Group Leader

We need to display updated names in all related WG proposals:
- Cancel WG lead opening
- Decrease WG Lead stake
- Terminate WG lead
- Set WG Lead Reward


- Update WG Budget

PR #2627, opened 2022-03-23
👀 Bugfix/rename wg names

stepien4soft:
Close: #2502


packages/ui/src/app/pages/WorkingGroups/WorkingGroup/WorkingGroup.tsx
```diff
@@ -53,13 +54,14 @@ export function WorkingGroup() {
       )
     }
   }
+  const replaceName = () => nameMapping(group?.name ?? name)
```
WRadoslaw, line 57: If `replaceName` isn't memoized by useMemo/useCallback, then I think u can just use `nameMapping` in places you need it. It will prevent redeclaring `replaceName` on each render.

WRadoslaw (changes requested):

WRadoslaw (approved):

Issue #2639, opened 2022-03-23
Submit judgement serves Submit work first time if process started from guest member


dmtrjsg:
Loom > https://www.loom.com/share/029f107bdbd6452c88daa4928bb7801d

polikosi:
Verified on staging. Test evidences below:

https://user-images.githubusercontent.com/80429191/167632185-c3392a10-ffc3-42d7-95a5-08beb63a7fbe.mp4

PR #2642, opened 2022-03-24
Fix for announce work on judgement modal

WRadoslaw:
follow-up #2639


thesan (changes requested):
It looks good but could you rebase it on `dev` please. It will fix the tests too

Issue #2310, opened 2022-02-23
Fill WG Lead Proposal: Application Membership details card


dmtrjsg:
### Scope

- [ ] Update application card or extend [membership card](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=6%3A863)
- [ ]  Show application form in the modal for selected application ([figma>](https://www.figma.com/file/QSpueQlOQKGbiHig2harPy/Working-groups?node-id=1080%3A4832216))
_______________
#### Originating ticket

**Taken at**: https://pioneer-2.vercel.app/#/proposals/current
**Fields**:
- Discrod Handle: -
- Give us some details: Distinguishing which of the applicants I want to propose to hire requires to crosscheck on another screen.
- Houston, we have a problem.: Choosing which applicants is difficult
- -

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/c4f10877-7945-42d9-b643-4d0f0a11d454-annotated_c43fb9b3-bdc8-475d-a0a0-b1e25688525f.png?etag=737afce0e5417090427734f28adb0ad1)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/c4f10877-7945-42d9-b643-4d0f0a11d454-annotated_c43fb9b3-bdc8-475d-a0a0-b1e25688525f.png?etag=737afce0e5417090427734f28adb0ad1)*

**Browser**: Firefox 97 (macOS Catalina)
**Screen size**: 2560x1440
**Browser size**: 2560x1240
**[Open #14 in Usersnap](https://app.usersnap.com/l/feedback/b0a51f1e-650d-464f-9cad-8ecd6aa6cafd?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

singulart:
<img width="1200" alt="Screenshot 2022-04-14 at 16 24 26" src="https://user-images.githubusercontent.com/7863785/163400242-ca9811a0-4a97-4e47-ac91-076fcb988d2f.png">

PR #2677, opened 2022-03-25
👾bugfix/fix select for fill WG lead

stepien4soft:
Close: #2310


thesan (changes requested):
Apart from the react `unique "key" prop` problem, it looks like the mock generation functions were not changed but all the mocks were regenerated. If it's the case please remove all the mock changes (to prevent unnecessary conflicts) with:
```bash
git checkout olympia packages/ui/src/mocks/data/
```

thesan (approved):

Issue #2651, opened 2022-03-24
[USERSNAP] Blank screen


dmtrjsg:
## Scope

**Desktop:**
- [x] Do not serve blank screen in browsers incompatible with Polkadot JS extensions. Simply put allow users on Safari, Edge etc to browse app in read-me. 

 @stepien4soft pls don't forget to update the copy in the modal on step 1:
 
<img width="530" alt="Screenshot 2022-03-29 at 09 49 20" src="https://user-images.githubusercontent.com/93924525/160572709-53a8eb8d-48ab-4db7-a670-4dcefa13e9e7.png">

Mobile
- [x]  For mobile browsers serve a banner>
Discord URL > https://discord.com/invite/DE9UN3YpRP
Homepage URL >  https://www.joystream.org/

## Designs
[FIGMA> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=8561%3A264038)

__________________

**Taken at**: https://dao.joystream.org/
**Fields**:
- Houston, we have a problem.: Blank screen
- Give us some details: Was told it was supposed to link me to download for pioneer
- Broken link
- Discrod Handle: THELOST+UNFOUNDS

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/f6d4eda6-1388-4d82-b110-107419533e0c-annotated_3429e01b-ad38-4816-8ecf-8919a94e4455.png?etag=59eccf0c96f5ab98fc5e0bb3294fa6bc)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/f6d4eda6-1388-4d82-b110-107419533e0c-annotated_3429e01b-ad38-4816-8ecf-8919a94e4455.png?etag=59eccf0c96f5ab98fc5e0bb3294fa6bc)*

**Browser**: Safari 15 (iOS)
**Screen size**: 428x926
**Browser size**: 428x752
**[Open #126 in Usersnap](https://app.usersnap.com/l/feedback/36eb1b9a-66c1-42ae-8e0c-6191c53ff4d7?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

[...]

PR #2714, opened 2022-03-29
 🙈 Feature/safari issue fixed

stepien4soft:
See: #2651

### DIFF
--- a/packages/ui/src/common/helpers.ts
+++ b/packages/ui/src/common/helpers.ts
@@ -12,8 +12,7 @@ export const cutText = (text: string, length = 100) => (text.length > length ? t
 
 export const isInFuture = (time: string) => {
   const timeAsDate = new Date(time).valueOf()
-  return timeAsDate > Date.now
-  ()
+  return timeAsDate > Date.now()
 }
 
 export const nameMapping = (value: string) => {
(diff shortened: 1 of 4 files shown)


packages/ui/src/common/helpers.ts
```diff
@@ -3,7 +3,7 @@ export const capitalizeFirstLetter = <T extends string>(str: T) =>
 
 export const lowerFirstLetter = (str: string): string => str.charAt(0).toLowerCase() + str.slice(1)
 export const camelCaseToText = (str: string): string => {
-  return capitalizeFirstLetter(str.replace(/(?<=[a-z])([A-Z])|(?<=[A-Z])([A-Z][a-z])/g, ' $1$2'))
+  return capitalizeFirstLetter(str.replace(/([A-Z]+)*([A-Z][a-z])/g, '$1 $2'))
```
thesan, line 6: Wow Safari only crashes because of a regexp problem :open_mouth: 
The new regexp mostly works but it adds an extra space at the beginning of the string (and also ideally we should add tests to these utility functions to make their expected behavior clearer).
IMO since there's already some lodash dependencies in the `ui` workspace the best, would be to remove this function and just use the [lodash `startCase` function](https://lodash.com/docs/4.17.15#startCase) instead. Also `lodash@^4` should be added to the `ui/package.json` (ATM the projects only builds because it's a dependency of `@/joystream/types`)

thesan (approved):

PR #2717, opened 2022-03-29
🤓 Feature/mobile view created

stepien4soft:
Close: #2651


packages/ui/src/app/App.tsx
```diff
@@ -69,3 +70,9 @@ export const App = () => (
     </NotificationsHolder>
   </Providers>
 )
+
+export const Mobile = () => (
+  <Providers>
+    <MobileView />
+  </Providers>
```
thesan, line 77: Can't the `<Providers>` wrapper not be used here ?

thesan (changes requested):
Thanks it looks good but could you also add a storybook for this please

thesan (approved):

PR #2721, opened 2022-03-29
🙃  New logic to check if candidacy lock is recoverable

WRadoslaw:


packages/ui/src/council/types/Candidate.ts
```diff
@@ -17,6 +17,7 @@ export interface ElectionCandidate {
     bannerUri?: string
   }
   status: CandidacyStatus
+  stakingAccount: any
```
thesan, line 20: Shouldn't this be a string ?

packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
```diff
@@ -1,25 +1,33 @@
...
 interface LocksDetailsProps {
   balance: Balances | null
   address: Address
 }
 
 export const LocksDetails = ({ balance, address }: LocksDetailsProps) => {
+  const { election } = useCurrentElection()
+
+  const candidateOnLock = useMemo(
+    () => election?.candidates.find((candidate) => candidate.stakingAccount === address),
```
thesan, line 20:
```suggestion
  const isActiveCandidate = useMemo(
    () => election?.candidates.find((candidate) => candidate.stakingAccount === address)?.status === 'ACTIVE',
```

packages/ui/src/accounts/model/lockTypes.ts
```diff
@@ -57,7 +57,13 @@ const isConflictingWith = (lockTypeA: LockType): ((lockTypeB: LockType) => boole
   }
 }
 
-export const isRecoverable = (type: LockType): boolean => RECOVERABLE.includes(type)
+export const isRecoverable = (type: LockType, isActiveCandidate?: boolean): boolean => {
+  if (isActiveCandidate) {
+    return type === 'Voting'
+  }
+
+  return RECOVERABLE.includes(type)
```
thesan, line 65: It relies on `RECOVERABLE` only having 2 lock types. This would be more future proof and IMO clearer:
```suggestion
  if (!RECOVERABLE.includes(type)) {
    return false
  }
  
  switch (type) {
    case 'Council Candidate':
      return !isActiveCandidate
    
    default:
      return true
  }
```

thesan (changes requested):

packages/ui/src/council/types/Candidate.ts
WRadoslaw: Yup! Good catch

thesan (approved):

PR #2742, opened 2022-03-31
🧊 Fix entry stake value and validation

### LINKED ISSUE #2672: Wrong entrant stake value in the add entry modal
- [x] When the bounty entrant stake is set to e.g 200 the Announce Entry modal will still display 100 (which is currently the min entrant stake const).

- [x]  I'm not sure about the withdraw stake/reward modal.

RafalPloszka:
Closes #2672

### DIFF
--- a/packages/ui/src/bounty/modals/AnnounceWorkEntryModal/AnnounceWorkEntryModal.tsx
+++ b/packages/ui/src/bounty/modals/AnnounceWorkEntryModal/AnnounceWorkEntryModal.tsx
@@ -3,6 +3,7 @@ import BN from 'bn.js'
 import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
+import * as Yup from 'yup'
 
 import { SelectAccount } from '@/accounts/components/SelectAccount'
 import { filterByRequiredStake } from '@/accounts/components/SelectAccount/helpers'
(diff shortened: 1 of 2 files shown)


thesan (changes requested):
The problem is that the entrant stake is set on bounty creation (ATM there is no way to set a higher amount when announcing an entry). So this modal should only display the `bounty.entrantStake` value (so users know how much they're going to be staking as a result of the transaction)

packages/ui/src/bounty/modals/AnnounceWorkEntryModal/AnnounceWorkEntryModal.tsx
```diff
@@ -52,21 +52,17 @@ export const AnnounceWorkEntryModal = () => {
     showModal,
   } = useModal<BountyAnnounceWorkEntryModalCall>()
   const { api, isConnected } = useApi()
-  const minWorkEntrantStake = api?.consts.bounty.minWorkEntrantStake.toNumber() ?? 0
   const { active: activeMember } = useMyMemberships()
   const { allAccounts } = useMyAccounts()
   const balances = useMyBalances()
-  const [amount, setAmount] = useState<string>(String(minWorkEntrantStake))
+  const minWorkEntrantStake = bounty.entrantStake
+  const [amount, setAmount] = useState<string>(minWorkEntrantStake.toString())
```
thesan, line 59:
```suggestion
  const amount = bounty.entrantStake
```
I think this `amount` should also replace `contribution`

RafalPloszka: `Amount` is string and `contribution` is BN (as required for components it is passed to), so this 'conversion' is needed, but we can get rid of useMemo, as it won't change, do you agree?

thesan (approved):

PR #2745, opened 2022-03-31
🤙 Fix on announce work entry tx

### LINKED ISSUE #2739: Bounty Entry: multiple issues
## Context

Originated in related issue: https://github.com/Joystream/joystream/issues/3526

It seems that if a user on submitting entry _manually pastes in an account address_ (instead of selecting from drop down) it allows to proceed to submission even  if an account is already bound to another member. 

This results in a failed transaction:

Screenshot 2022-04-08 at 14 53 29

In one instance it used an address that the user _did not own/have in their wallet._

Reference videos:
* https://drive.google.com/file/d/12tRqKKuSq0unrO0ap8BeU4k0Bz7giyBk/view?usp=sharing
* https://drive.google.com/file/d/18ltYI4zHWPh3kf7KUMrKhrDDlDgrDt1S/view?usp=sharing
* https://drive.google.com/file/d/1pWnD1ohPRrpfzknz4Bf3ZMWqkhAqn580/view?usp=sharing

WRadoslaw:
Closes #2739

### DIFF
--- a/packages/ui/src/bounty/modals/AnnounceWorkEntryModal/AnnounceWorkEntryModal.tsx
+++ b/packages/ui/src/bounty/modals/AnnounceWorkEntryModal/AnnounceWorkEntryModal.tsx
@@ -3,6 +3,7 @@ import BN from 'bn.js'
 import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
+import * as Yup from 'yup'
 
 import { SelectAccount } from '@/accounts/components/SelectAccount'
 import { filterByRequiredStake } from '@/accounts/components/SelectAccount/helpers'
@@ -20,6 +21,7 @@ import { SuccessTransactionModal } from '@/bounty/modals/SuccessTransactionModal
 import { ButtonPrimary } from '@/common/components/buttons'
 import { FailureModal } from '@/common/components/FailureModal'
 import { Input, InputComponent, InputNumber } from '@/common/components/forms'
+import { getErrorMessage, hasError } from '@/common/components/forms/FieldError'
 import {
   Modal,
   ModalFooter,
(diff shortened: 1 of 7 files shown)


packages/ui/src/bounty/modals/AnnounceWorkEntryModal/AnnounceWorkEntryModal.tsx
```diff
@@ -36,14 +38,21 @@ import { WaitModal } from '@/common/components/WaitModal'
...
 import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
 import { BindStakingAccountModal } from '@/memberships/modals/BindStakingAccountModal/BindStakingAccountModal'
 import { SwitchMemberModalCall } from '@/memberships/modals/SwitchMemberModal'
+import { IStakingAccountSchema, StakingAccountSchema } from '@/memberships/model/validation'
 
 const transactionSteps = [{ title: 'Bind staking account' }, { title: 'Announce Work' }]
 
+const schema = Yup.object().shape({
+  amount: Yup.string(),
+  account: StakingAccountSchema,
```
thesan, line 53: Because the goal is to remove as much duplicated logic as possible, if possible I think the `account` should be required (does `StakingAccountSchema.required` works ?) and at the end:
```ts
<ButtonPrimary size="medium" disabled={!isValid} onClick={nextStep}>
```
With `isValid` being the value returned by `useSchema`. This way all the logic linked to `isValidNext` can be removed and YUP takes care of all the validation.

WDYT ?

thesan (changes requested):
It's really nice to have `StakingAccountSchema` be reusable :heart_eyes: 

About the modal, it is only supposed to display the bounty entrant stake not set it, this is what Rafal is fixing in #2742, so IMO it shouldn't be validated either.

Finally do you think we could use something like `(account) => schema.isValidSync(account, { context })` as a filter for the `<SelectAccount />` ?

packages/ui/src/memberships/model/validation.ts
```diff
@@ -18,6 +21,38 @@ export const ReferrerSchema = Yup.object().when('isReferred', (isReferred: boole
...
+}
+
+export const StakingAccountSchema = Yup.object()
+  .test('balance', 'Balance on this account is insufficient', (value, context) => {
+    if (!value) {
+      return true
+    }
+
+    const validationContext = context.options.context as IStakingAccountSchema
+    const extraAmount = validationContext?.requiresBounding ? new BN(200) : BN_ZERO
```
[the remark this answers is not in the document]
WRadoslaw: You are right, I should fetch this amount from API

packages/ui/src/bounty/modals/AnnounceWorkEntryModal/AnnounceWorkEntryModal.tsx
WRadoslaw: Only question is whether we want this issue to just solve the bugs or addinally straight out the logic. I'll be cleaning every bounty modal in separate issue, so I thought I'll leave it out not to overcomplicate PR.

thesan: Got it, I thought you wanted to split the "cleaning" one modal at a time.
But if this PR is just about fixing #2739, lets leave this suggestion and the `<SelectAccount />` filter for another PR.

thesan (approved):

PR #2746, opened 2022-03-31
🤑 Add the stake amount to the cashout sign modal

thesan:
This shows and details the actual amount "cashed out" by bounty winners, which is the reward (already shown before) **+ the amount their entries stake**


WRadoslaw (approved):

thesan (commented):
@RafalPloszka thanks a lot for fixing this :bow:

PR #2767, opened 2022-04-02
MyStakeStat: String Concatenation in getTotalStake

### LINKED ISSUE #2427: [USERSNAP] WG> Openings: "Currently Staking" Widget shows wrong number
**Taken at**: https://pioneer-2.vercel.app/#/working-groups/openings
**Fields**:
- Houston, we have a problem.: Working groups: Currently Staking is wildly incorrect
- Give us some details: Working groups: Currently Staking is wildly incorrect
- Working groups: Currently Staking is wildly incorrect
- Discrod Handle: tomato#9464

SCREENSHOT
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/8956ba1b-9f83-4bd4-a616-03c7d69b1af3-annotated_761ce13f-1c73-4a96-9b27-6b313f75c477.png?etag=3b662bb23cd03200f93618c2d4f3398a)*

**Browser**: Chrome 98 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1718x1306
**[Open #55 in Usersnap](https://app.usersnap.com/l/feedback/55b293f8-7b5c-472c-b653-e9a3501e8739?utm_medium=integration&utm_campaign=integration&utm_content=open)**

oleksanderkorn:
When a user has more than one role, stakes are reduces to a sum, but the stake property is of string type, so it leads to wrong result: 
image
image

Fixes #2427

### DIFF
--- a/packages/ui/src/working-groups/components/MyStakeStat.tsx
+++ b/packages/ui/src/working-groups/components/MyStakeStat.tsx
@@ -4,7 +4,7 @@ import { TokenValueStat } from '@/common/components/statistics'
 import { useMyWorkers } from '@/working-groups/hooks/useMyWorkers'
 import { Worker } from '@/working-groups/types'
 
-const getTotalStake = (workers: Worker[]) => workers.reduce((total, worker) => total + worker.stake, 0)
+const getTotalStake = (workers: Worker[]) => workers.reduce((total, worker) => total + Number(worker.stake), 0)
 
 export const MyStakeStat = () => {
   const { workers } = useMyWorkers()
--- a/packages/ui/src/working-groups/components/MyStakeStat.tsx
+++ b/packages/ui/src/working-groups/components/MyStakeStat.tsx
@@ -1,14 +1,12 @@
 import React from 'react'
 
 import { TokenValueStat } from '@/common/components/statistics'
+import { sumStakes } from '@/common/utils/bn'
 import { useMyWorkers } from '@/working-groups/hooks/useMyWorkers'
-import { Worker } from '@/working-groups/types'
-
-const getTotalStake = (workers: Worker[]) => workers.reduce((total, worker) => total + Number(worker.stake), 0)
 
 export const MyStakeStat = () => {
   const { workers } = useMyWorkers()
-  const totalStake = getTotalStake(workers)
+  const totalStake = sumStakes(workers)
 
   return <TokenValueStat title="Currently staking" value={totalStake} />
 }


oleksanderkorn:
I am not sure if it makes sense to change the type in `WorkerFieldsFragment`,  the question is - do we expect there something other than number, why is it `any` here ?:
image

Or we can parse number here as well, also by checking if it is actually a number, if we expect NaN there:
image

oleksanderkorn:
@jodator @WRadoslaw please take a look :)

oleksanderkorn:
image
Applied fix works as expected.

WRadoslaw:
Hey! Firstly thanks for your input! About the issue I might have two ideas. 

- Question about `any` in `WorkerFieldsFragement` is a good one and basically what caused bug. This type is generated by `@graphgl-codegen` and looks like it just doesn't know what should be `BigInt` type analouge.  @thesan do you know  whether `BigInt` values come only as `string` from QN,? If so we can assigne them in `src/codegen.config.yml` as below and than regenerate all queries.

```
config:
  scalars:
    Date: Date
    BigInt: string
```
- About resolving the bug, I think we should keep this value as `BN`. I fear that in some cases too big number will cause some

thesan:
Hey @oleksanderkorn, thanks a lot for the PR.

@WRadoslaw is correct:

1. On the runtime every amounts of `JOY` are encoded as `u128` so `BN` should be used on Pioneer side. Also because this summing stake pattern is so common there is a helper function for it there: [@/common/utils/bn](https://github.com/Joystream/pioneer/blob/e7f295bca516dba89f3ee9a7c457ca0caa43a07e/packages/ui/src/common/utils/bn.ts).

2. We should map the `BigInt` from the QN schema to `string`s and regenerate the queries (we could handle this in another PR)

oleksanderkorn:
> Hey @oleksanderkorn, thanks a lot for the PR.
> 
> @WRadoslaw is correct:
> 
> 1. On the runtime every amounts of `JOY` are encoded as `u128` so `BN` should be used on Pioneer side. Also because this summing stake pattern is so common there is a helper function for it there: [@/common/utils/bn](https://github.com/Joystream/pioneer/blob/e7f295bca516dba89f3ee9a7c457ca0caa43a07e/packages/ui/src/common/utils/bn.ts).
> 2. We should map the `BigInt` from the QN schema to `string`s and regenerate the queries (we could handle this in another PR)

Thanks, pushed the change to use existing `sumStakes`.

thesan (approved):
Thanks again :bow:

singulart:
Verified on staging. 

Screenshot 2022-04-29 at 00 01 20

PR #2788, opened 2022-04-04
🦎 Fix voting locks recovering

### LINKED ISSUE #2363: Dont allow to recover voting stake / when candidate became councillor (for as long as candidate in the council seat)
## Problem

Right now the CTA is shown, but transaction fails. 

## To-do

- [ ] Hide Recover voting stake for cases, when voting stake was retained as candidate won the election. 

> So user votes  [cant recover]-> (candidate wins)[cant recover] -> council is in normal state [cant recover] -> council transfers to election state [ becomes REcoverable!] 

###  Additional info

-! Should be solved together with -https://github.com/Joystream/pioneer/issues/2543
-! Design enhancement of this point is handled already in a separate issue > 2354 and 2300

~https://polkadot.js.org/apps/?rpc=wss://3.95.67.177.nip.io/ws-rpc#/explorer/query/0xca5e2866c16837c14c477e76d7fd6021e9c48b15e51f74759ccd8fac0f8d5118~ (chain updated since)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201976887648134) by [Unito](https://www.unito.io)

RafalPloszka:
Closes #2363

### DIFF
--- a/packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
+++ b/packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
@@ -27,7 +27,7 @@ export const LocksDetails = ({ balance, address }: LocksDetailsProps) => {
     const candidate = votes?.find((vote) => vote.castBy === address)?.voteFor
     // Disable recovering during election or if candidate won previous election
     // Enable recovering if new council is elected
-    return !!(election || (candidate?.isCouncilMember && !election))
+    return !!candidate && (!election?.isFinished || candidate.isCouncilMember)
   }, [votes, address, election])
 
   if (!balance || !balance.locks.length) {
--- a/packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
+++ b/packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
@@ -7,6 +7,7 @@ import { RowGapBlock } from '@/common/components/page/PageContent'
 import { Label, TextMedium } from '@/common/components/typography'
 import { Address } from '@/common/types'
 import { useCurrentElection } from '@/council/hooks/useCurrentElection'
+import { useMyCastVotes } from '@/council/hooks/useMyCastVotes'
 
 interface LocksDetailsProps {
   balance: Balances | null
@@ -15,19 +16,27 @@ interface LocksDetailsProps {
 
 export const LocksDetails = ({ balance, address }: LocksDetailsProps) => {
   const { election } = useCurrentElection()
+  const { votes } = useMyCastVotes()
 
   const isActiveCandidate = useMemo(
     () => election?.candidates.find((candidate) => candidate.stakingAccount === address)?.status === 'ACTIVE',
     [election, address]
   )
 
+  const isVoteStakeLocked = useMemo(() => {
+    const candidate = votes?.find((vote) => vote.castBy === address)?.voteFor
+    // Disable recovering during election or if candidate won previous election
+    // Enable recovering if new council is elected
+    return !!(election || (candidate?.isCouncilMember && !election))
+  }, [votes, address])
+
   if (!balance || !balance.locks.length) {
     return <TextMedium light>No locks found.</TextMedium>
   }
 
   const allLocks = balance.locks
-  const recoverable = allLocks.filter(({ type }) => isRecoverable(type, isActiveCandidate))
-  const nonRecoverable = allLocks.filter(({ type }) => !isRecoverable(type, isActiveCandidate))
+  const recoverable = allLocks.filter(({ type }) => isRecoverable(type, isActiveCandidate, isVoteStakeLocked))
+  const nonRecoverable = allLocks.filter(({ type }) => !isRecoverable(type, isActiveCandidate, isVoteStakeLocked))
 
   return (
     <>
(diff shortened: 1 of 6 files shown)


packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
```diff
@@ -15,19 +16,27 @@ interface LocksDetailsProps {
 
 export const LocksDetails = ({ balance, address }: LocksDetailsProps) => {
   const { election } = useCurrentElection()
+  const { votes } = useMyCastVotes()
```
thesan, line 19: Sorry this is on me but the `useMyCastVotes` hook is currently very confusing. It won't return anything unless you provide the `cycleId`. So I think this hook won't work by itself.

Replacing the `useCurrentElection` by a hook which queries something like this:
```gql
query LatestElection {
	electionRounds (orderBy: [cycleId_DESC], limit: 1) {
    ...ElectionRoundFields
  }
}
```
Could work because you would still know whether the election is finished (for `isActiveCandidate`) and the `cycleId` would be available so you could query the votes only if the election is finished.

But the production is currently in a weird state it won't really work there ATM

packages/ui/src/accounts/model/lockTypes.ts
```diff
@@ -57,7 +57,7 @@ const isConflictingWith = (lockTypeA: LockType): ((lockTypeB: LockType) => boole
   }
 }
 
-export const isRecoverable = (type: LockType, isActiveCandidate?: boolean): boolean => {
+export const isRecoverable = (type: LockType, isActiveCandidate?: boolean, isVoteStakeLocked?: boolean): boolean => {
```
thesan, line 60: IMO it would be nicer to pass `isActiveCandidate` and `isVoteStakeLocked` in an object, e.g:
```suggestion
export const isRecoverable = (type: LockType, { isActiveCandidate, isVoteStakeLocked }: RecoverStakeConditions): boolean => {
```

thesan (changes requested):
The logic is good the problem is the `useMyCastVotes` behavior and that the prod QN might prevent us from actually testing this

packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
```diff
@@ -6,28 +6,37 @@ import { Balances } from '@/accounts/types'
...
   const isActiveCandidate = useMemo(
     () => election?.candidates.find((candidate) => candidate.stakingAccount === address)?.status === 'ACTIVE',
     [election, address]
   )
 
+  const isVoteStakeLocked = useMemo(() => {
+    const candidate = votes?.find((vote) => vote.castBy === address)?.voteFor
+    // Disable recovering during election or if candidate won previous election
+    // Enable recovering if new council is elected
+    return !!(election || (candidate?.isCouncilMember && !election))
```
thesan, line 30: Here `election` will be defined (unless there something wrong with the QN) but the question is whether:
1. A vote was cast in the latest election: the stake **shouldn't** be locked due to a vote cast several election ago
2. The latest election (where the vote was cast) is still going on, then the stake **should** be locked
3. The vote was cast to a winning candidate, then the stake **should** be locked

```suggestion
    return !!candidate && (!election?.isFinished || candidate.isCouncilMember)
```
(here if `election` is not defined the stake is not locked)

NB: AFAIK `isFinished` exist on the QN but not on the `Election` type so it should be added

thesan (changes requested):
In the end it is only one line but thinking of this flow gave me a headache :exploding_head:
Hopefully I haven't forgot about some other cases again or misinterpreted the handbook :crossed_fingers: 
If you think I'm wrong don't hesitate to tell me we can have a call about it

traumschule:
[Was able to recover on past-votes but not on profile](https://discord.com/channels/811216481340751934/910968995379822633/962689636545732628), created two issues: https://github.com/Joystream/pioneer/issues/2834 https://github.com/Joystream/pioneer/issues/2835

packages/ui/src/council/queries/council.graphql
```diff
@@ -87,6 +87,11 @@ fragment ElectionRoundFields on ElectionRound {
   }
 }
 
+fragment LatestElectionRoundFields on ElectionRound {
+  ...ElectionRoundFields
+  isFinished
+}
```
thesan, line 93: Looks like this fragment is not used

thesan (commented):
Thanks, it looks good, just one more small detail

packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
```diff
@@ -6,28 +6,37 @@ import { Balances } from '@/accounts/types'
...
 
   const isActiveCandidate = useMemo(
     () => election?.candidates.find((candidate) => candidate.stakingAccount === address)?.status === 'ACTIVE',
     [election, address]
   )
 
+  const isVoteStakeLocked = useMemo(() => {
+    const candidate = votes?.find((vote) => vote.castBy === address)?.voteFor
+    // Disable recovering during election or if candidate won previous election
+    // Enable recovering if new council is elected
```
thesan, line 29: Also maybe this comment should be updated

packages/ui/src/council/queries/council.graphql
RafalPloszka: It's used for `LatestElectionRoundFieldsFragment` in `council/types/LatestElection.ts`, but it should be also applied in `query GetLatestElection`, fixed it :)

thesan:
> [Was able to recover on past-votes but not on profile](https://discord.com/channels/811216481340751934/910968995379822633/962689636545732628), created two issues: #2834 #2835

Thank you for the feedback. Sorry I completely missed this message this morning. I can't access the discord message ATM, but did you test this on https://dao-git-fix-block-vote-stake-recover-when-cand-0bac59-joystream.vercel.app/ or just https://dao.joystream.org/ ?

traumschule:
just dao.joystream.org.

Issue #2685, opened 2022-03-26
[USERSNAP] Council "Voter Stake" is incorrect


dmtrjsg:
- [ ] Show cumulative voters stake in this column

**Taken at**: https://dao.joystream.org/#/council
**Fields**:
- Houston, we have a problem.: Council &#34;Voter Stake&#34; is incorrect
- Give us some details: These appear to be the &#34;Announcing stake&#34;, not the &#34;Voters Stake&#34;
- -
- Discord Handle: bwhm

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/26121b3e-3e55-4df8-9290-5dc69cd77b39-annotated_922f6ef1-f4bb-4141-ab32-e6f615c36fb3.png?etag=c4193a68c774a47a714e1d68259aaf4d)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/26121b3e-3e55-4df8-9290-5dc69cd77b39-annotated_922f6ef1-f4bb-4141-ab32-e6f615c36fb3.png?etag=c4193a68c774a47a714e1d68259aaf4d)*

**Browser**: Firefox 98 (macOS Catalina)
**Screen size**: 2560x1440
**Browser size**: 2560x1304
**Labels**: enhancement
**[Open #139 in Usersnap](https://app.usersnap.com/l/feedback/8b9a42d7-f548-4e2e-8184-7fbbe853c34e?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

thesan:
Blocked by Joystream/joystream#3573

ivanturlakov:
Verified on staging.
According to the results of recent council elections, the stakes are displayed correctly.
<img width="1440" alt="Снимок экрана 2022-05-17 в 13 33 47" src="https://user-images.githubusercontent.com/46903215/168797984-fe920563-3d4b-49fd-b72d-ab8e83ec594a.png">
<img width="1440" alt="Снимок экрана 2022-05-17 в 13 34 27" src="https://user-images.githubusercontent.com/46903215/168798006-cb0175f9-a4da-4de3-912b-17f411a1cb51.png">

PR #2791, opened 2022-04-05
🥸 bugfix/incorrect-voter-stake

stepien4soft:
See: #2685


packages/ui/src/app/pages/Council/Council.tsx
```diff
@@ -22,9 +23,10 @@ export const Council = () => {
   const { idlePeriodRemaining, budget, reward } = useCouncilStatistics(council?.electedAt.number)
   const { activities } = useCouncilActivities()
   const { stage: electionStage } = useElectionStage()
-
   const [order, setOrder] = useState<CouncilOrder>({ key: 'member' })
   const councilors = useMemo(() => council?.councilors.slice(0).sort(sortBy(order)) ?? [], [council])
+  // const data = useVoteStake( [council.id], [] )
+  // console.log('data', data)
```
thesan, line 29:
```suggestion
  const { councilors  } = useCouncilorWithDetails(council)
  const sortedCouncilors = useMemo(() => councilors.sort(sortBy(order)), [councilors])
```

thesan (commented):
@stepien4soft I've tested the queries hopefully they'll keep working :crossed_fingers:

traumschule:
LGTM!
- In light of this bug https://github.com/Joystream/joystream/issues/3530#issuecomment-1120361999 it would be very interesting to also see failed candidates somehow - maybe grayed out below or with a heading `Not Elected`?
- Would be great to see stakes for past councils as well (including failed candidates) like here https://dao-git-bugfix-voter-stake-fix-joystream.vercel.app/#/council/past-councils/00000006
- Is it possible to review former candidacies like bullet points, image and program? Basically like Election View during an election but for past terms and with final vote power.

WDYT @dmtrjsg shall i create separate issues?

packages/ui/src/council/queries/council.graphql
```diff
@@ -175,6 +178,23 @@ query GetElectedCouncil {
...
 }
 
+query GetElectionRound {
+  electedCouncils(where: { isResigned_eq: false }, orderBy: [createdAt_DESC], limit: 1) {
+    councilElections {
+      cycleId
+    }
+  }
+}
+
```
thesan, line 188: AFAICS this query is not used

thesan (approved):
Looks good :+1: , almost done !
Please just remove the unused query and regenerate the queries before merging

Issue #2674, opened 2022-03-24
Save state of viewed Overview Deadlines Notif in browser local storage


dmtrjsg:
To-do:

- [ ] Store ids for closed ones in local storage

Bonus:

- [ ] Sort by time across all 4 entity types: proposals; openings; upcoming openings; election

____

**Taken at**: https://dao.joystream.org/#/overview
**Fields**:
- Houston, we have a problem.: Current deadlines
- Give us some details: I think they should be permanently deleted. Otherwise, what&#39;s the purpose of being able to delete them?
- In Overview, I delete the deadlines, refresh the page, and the deleted deadlines reappear
- Discord Handle: adovrn#1094

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/a3ffdb5f-c803-4193-86f5-09921ab50098-annotated_64e37ffa-45cb-41f1-8424-3968b0636f5e.png?etag=94ac97038df12680c237cb18435c2ee3)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/a3ffdb5f-c803-4193-86f5-09921ab50098-annotated_64e37ffa-45cb-41f1-8424-3968b0636f5e.png?etag=94ac97038df12680c237cb18435c2ee3)*

**Browser**: Chrome 99 (Windows 10)
**Screen size**: 1536x864
**Browser size**: 1536x792
**Labels**: bug-non-critical
**[Open #136 in Usersnap](https://app.usersnap.com/l/feedback/eafc46cc-71fc-4f9f-80d5-ce12ca601e6f?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

PR #2813, opened 2022-04-06
👁 Fix/deadlines on storage

stepien4soft:
Close: #2674


packages/ui/src/overview/components/DeadlineList/ElectionListItem.tsx
```diff
@@ -83,7 +84,13 @@ export const ElectionListItem: React.FC<ElectionListItemProps> = React.memo(({ e
     <ElementWrapper>
       <ListItem>
         <TopElementsWrapper>
-          <StyledTriangle deadlineTime={remainingCalculation} /> <StyledClosedButton onClick={setHideElement} />
+          <StyledTriangle deadlineTime={remainingCalculation} />{' '}
+          <StyledClosedButton
+            onClick={() => {
+              setHideElement()
```
thesan, line 90: `useLocalStorage` uses `useState` to update the data in real time (it doesn't just store data and retrieve them when mounted). As a result this shouldn't be needed anymore

thesan (changes requested):

packages/ui/src/overview/components/DeadlineList/DeadlineList.tsx
```diff
@@ -19,14 +21,63 @@ export interface DeadlineListProps {
...
+        ? election?.candidates.filter((candidate) => !tempDeadlines.election.includes(candidate.id)) ?? []
+        : election?.candidates ?? [],
+      openings: tempDeadlines?.openings
+        ? openings.filter((opening) => !tempDeadlines.openings.includes(opening.id))
+        : openings,
+      upcomingOpenings: tempDeadlines?.upcomingOpenings
+        ? upcomingOpenings.filter((upcomingOpening) => !tempDeadlines.upcomingOpenings.includes(upcomingOpening.id))
+        : upcomingOpenings,
+    }
+  }, [storageDeadlines?.proposals, proposals, election, upcomingOpenings, openings])
```
thesan, line 78: It's great that the deadline are all queried in one the hook, but also I think all the filtering/(maybe later sorting) should happen in the same hook too

thesan, line 78: It's great that the deadline are all queried in one the hook, but also I think all the filtering/(maybe later sorting) should happen in the same hook too

thesan (approved):
Looks good thanks :+1:

PR #2836, opened 2022-04-10
👨‍🚀 POC - Form vendor

WRadoslaw:


thesan (changes requested):
It looks good but I would prefer replacing "Controlled" by something else e.g RFH

thesan (approved):
Very impressive PR again :muscle: 
Thanks

Issue #2803, opened 2022-04-06
Implement Husky 


dmtrjsg:
- [ ] Add husky and set up configuration on pre-commit
- [ ] Implement lint-staged library for the fastest code fixes

Issue #2805, opened 2022-04-06
Re-implement Prettier


dmtrjsg:
Together with Husky

- [ ] Create new rules for prettier and linter

PR #2851, opened 2022-04-12
🐕 Feature/husky implementation

stepien4soft:
Close: #2803 #2805


thesan (changes requested):
For me the `yarn lint` script is run entirely which takes ~2min
Screenshot from 2022-04-13 12-29-07
Doesn't it need the command to accept the modified files as arguments ?

Also does it works as expected on macOS ?

Finally it would be great if you could add the `index.html` in there as it's the only file my IDE can't format properly :disappointed:  (if it's not worth spending too long on this though).

thesan (approved):

Issue #2514, opened 2022-03-10
Proposal: Slash WG Lead  needs to show current stake by lead


dmtrjsg:
### Context

Current State:
> currently staked amount is not visible
<img width="1415" alt="Screenshot 2022-03-24 at 15 51 44" src="https://user-images.githubusercontent.com/93924525/159956928-27ccc7ee-da77-40e6-b405-c30b0063d618.png">

### Scope

- [ ] Implement displaying staked amount on creating proposal  (see design links below). Display in the light blue background info box. Copy: "Info: The actual amount staked by this member is {{ XYZ JOY}}."
- [ ] Implement displaying staked amount on proposal details page via a widget (to enable informed voting). Title of widget "Currently Staked by Worker". Content - sum in tJoys

ℹ️ NB: More than at stake can be slashed (and that is OK), so validation does not apply.

### Technical Guidance

This may be helpful in fixing this issue (IF the stake is not available on the worker)

Find the hook that fetches that specific data
Proceed to the schema of the hook and check if you have access to this data on Graph QL
Add this data to entity in council.graphql
Regenerate Queries following [this documentation](https://github.com/Joystream/pioneer/blob/dev/docs/README.md#adding-queries)
Update casting and types
Apply the changes to the component displaying the BP

### Design Notes
[LINK TO FIGMA >](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=235%3A2012)

DESIGN REFERENCE
<img width="1223" alt="Screenshot 2022-03-24 at 15 51 20" src="https://user-images.githubusercontent.com/93924525/159956914-591d88d2-6fb1-4e7a-82b9-49d65077cc5e.png">

⛔  Out of scope:

- [ ] #2460 

_______

**Taken at**: https://pioneer-2.vercel.app/#/proposals/current
**Fields**:
- Houston, we have a problem.: Proposal: slash lead page doesn&#39;t show current lead stake, lets input of higher numbers happen
- Give us some details: Page should show current stake the lead has
- Page should show current stake the lead has
- Discrod Handle: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/5b85d1a5-128e-412c-8b95-b1a72b63d1a3-annotated_cdf09b04-4716-44f0-a63e-b4d0b3f0dde3.png?etag=69ab86b7123b06284856d4a906f2185a)
[...]

PR #2853, opened 2022-04-12
Proposal: Slash WG Lead needs to show current stake by lead

oleksanderkorn:
https://github.com/Joystream/pioneer/issues/2514

image


thesan (approved):
Thank you for the PR. LGTM !

PR #2862, opened 2022-04-15
👾 Hotfix on add new proposal modal long loading

WRadoslaw:


thesan (approved):
LGTM!

But to fix all the modals later I'm thinking of using a custom Rxjs operator like this:

```ts
const defaultWhenTimeout = <T>(fallback: T, timeout: number) => (source: Observable<T>) =>
  source.pipe(
    takeUntil(timer(timeout)),
    defaultIfEmpty(fallback),
    map((data) => merge(of(data), source)),
    switchAll(),
  )
```

Then in `useTransactionFee` the observable becomes:
```ts
transaction?.paymentInfo(address).pipe(defaultWhenTimeout(paymentInfoFallback, 2000))
```
WDYT ?

Issue #2660, opened 2022-03-24
[USERSNAP] Spam deadlines


dmtrjsg:
- [ ] Only show proposals, bounties, openings and election cycles where a _particular_ member has accounts used for staking or signed transaction fees.

For guest users nothing needed :) 
__
⚠️  Limit the scope to just members, and figure out what depends on the staking account to create a follow up issues, once known.
⚠️ This issue depends on https://github.com/Joystream/pioneer/issues/2750
⚠️  Do together with https://github.com/Joystream/pioneer/issues/1544
⚠️ some events may not be available.
________

**Taken at**: https://dao.joystream.org/#/overview
**Fields**:
- Houston, we have a problem.: Spam deadlines
- Give us some details: Maybe there&#39;s a need for personal deadlines here?
- Spam deadlines
- Discrod Handle: adovrn#1094

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/536d0a54-3ba9-422c-9011-d760ebfc1f67-annotated_59d22140-d4d0-4ef6-ae28-9ceb7d1127b0.png?etag=0c8cd0b783b7915918c6bce3680c5154)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/536d0a54-3ba9-422c-9011-d760ebfc1f67-annotated_59d22140-d4d0-4ef6-ae28-9ceb7d1127b0.png?etag=0c8cd0b783b7915918c6bce3680c5154)*

**Browser**: Chrome 99 (Windows 10)
**Screen size**: 1536x864
**Browser size**: 1536x792
**Labels**: bug-non-critical
**[Open #131 in Usersnap](https://app.usersnap.com/l/feedback/902e71b5-7ff1-48e8-979a-71d7ae386f6f?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
https://github.com/Joystream/pioneer/issues/1544

PR #2891, opened 2022-04-22
⏰ Bugfix/fix spam deadline list

stepien4soft:
Closes: #2660


packages/ui/src/overview/components/DeadlineList/ElectionListItem.tsx
```diff
@@ -48,39 +57,52 @@ export const ElectionListItem: React.FC<ElectionListItemProps> = React.memo(({ e
...
+        }
         return (
           <StyledText>
-            {t('deadline.votingPeriod', { electionName: title })}{' '}
-            <StyledLink href={urlAddress + electionId}>{t('deadline.votingLink')}</StyledLink>
+            {t('deadline.votingPeriod')} <StyledLink href={urlAddress}>{t('deadline.votingLink')}</StyledLink>
           </StyledText>
         )
       case 'revealing':
+        if (!allVotesRevealed) {
```
thesan, line 84: Here if all the votes are revealed we should **not** show the revealing deadline
```suggestion
        if (votes.every((vote) => vote.voteFor)) {
```

packages/ui/src/overview/components/DeadlineList/ElectionListItem.tsx
```diff
@@ -22,15 +27,19 @@ import {
...
 
-export const ElectionListItem: React.FC<ElectionListItemProps> = React.memo(({ electionId, title, hideForStorage }) => {
+export const ElectionListItem: React.FC<ElectionListItemProps> = React.memo(({ hideForStorage, member, election }) => {
   const { t } = useTranslation('overview')
   const { stage: electionStage } = useElectionStage()
+  const { allAccounts } = useMyAccounts()
+  const { votesTotal } = useMyCurrentVotesCount(election.cycleId)
+  const canVote = isDefined(votesTotal) && allAccounts.length > votesTotal
   const remainingPeriod = useElectionRemainingPeriod(electionStage)
+  const { votes } = useMyCastVotes(election.cycleId)
```
thesan, line 42: We can skip this hook if we're not on the `revealing` stage and also it would simplify the logic to default to an empty array

```suggestion
  const { votes } = useMyCastVotes(electionStage === 'revealing' ? election.cycleId : undefined) ?? []
```

packages/ui/src/overview/components/DeadlineList/ElectionListItem.tsx
```diff
@@ -48,39 +57,52 @@ export const ElectionListItem: React.FC<ElectionListItemProps> = React.memo(({ e
     }
   }, [])
 
-  const copyMessage = () => {
-    const urlAddress = '#/election?candidate='
+  const getCopyMessage = () => {
```
thesan, line 60: That's my mistake, but it would make more sense to have a `useMemo` here:
```suggestion
  const getCopyMessage = useMemo(() => {
    ...
  }, [electionStage, election.candidates.length, canVote, votes])
```

packages/ui/src/overview/hooks/useDeadlines.ts
```diff
@@ -33,13 +41,15 @@ export const useDeadlines = (params?: UseDeadlinesParams) => {
...
+    const election = rawElection && [asElection(rawElection)]
     const proposals = data?.proposals || []
-    const upcomingOpenings = data?.upcomingWorkingGroupOpenings.map(asUpcomingWorkingGroupOpening) || []
+    const upcomingOpenings = data?.upcomingWorkingGroupOpenings?.map(asUpcomingWorkingGroupOpening) || []
     const openings = data?.workingGroupOpenings || []
     return {
       proposals: proposals.filter((proposal) => !storageDeadlines.proposals.includes(proposal.id)),
-      election: election?.candidates.filter((candidate) => !storageDeadlines.election.includes(candidate.id)) ?? [],
+      elections:
+        election?.filter((election) => !storageDeadlines.elections.includes(`${election.cycleId}:${electionStage}`)) ??
```
thesan, line 51: Because `electionStage` is used here, it would be safer to add it to the dependencies of this `useMemo`

packages/ui/src/overview/hooks/useDeadlines.ts
```diff
@@ -1,26 +1,34 @@
...
 }
 
-export const useDeadlines = (params?: UseDeadlinesParams) => {
-  const options = params?.groupId || {}
-  const { loading, data } = useGetAllDeadLinesQuery(options)
+export const useDeadlines = (params: UseDeadlinesParams) => {
+  const variables = {
+    groupId: params.groupId || {},
+    proposalCreator: params.member.isCouncilMember ? undefined : { id_eq: params.member.id },
+    isLead: false,
```
thesan, line 28: I meant this `isLead: false` as a placeholder but we actually need to add some logic to determine whether the member is the lead of some WG and if so which ones, and finally pass the ids. e.g:
```suggestion
    isLead: groupIds.length > 0,
    group: groupIds.length > 0 ? { id_in: groupIds } : undefined,
```

packages/ui/src/overview/hooks/useDeadlines.ts
```diff
@@ -1,26 +1,34 @@
...
-  election: [],
+  elections: [],
 }
 
-export const useDeadlines = (params?: UseDeadlinesParams) => {
-  const options = params?.groupId || {}
-  const { loading, data } = useGetAllDeadLinesQuery(options)
+export const useDeadlines = (params: UseDeadlinesParams) => {
+  const variables = {
+    groupId: params.groupId || {},
```
thesan, line 26: This is not used by the query
```suggestion
```

packages/ui/src/overview/hooks/useDeadlines.ts
```diff
@@ -33,13 +41,15 @@ export const useDeadlines = (params?: UseDeadlinesParams) => {
 
   const deadlines = useMemo((): Record<DeadlineNamespace, any[]> => {
     const rawElection = data?.electionRounds[0]
-    const election = rawElection && asElection(rawElection)
+    const election = rawElection && [asElection(rawElection)]
```
thesan, line 44: That's a small detail but lets drop the `rawElection` just to make it consistent with the other deadlines
```suggestion
    const election = data?.electionRounds.map(asElection(rawElection)) ?? []
```

thesan (changes requested):
Sorry about the delay to review this one. I added a few CR about the election part but mostly the filtering of the openings/upcoming openings is missing. The deadlines should be all good after that

packages/ui/src/overview/hooks/useDeadlines.ts
```diff
@@ -1,26 +1,42 @@
...
-export const useDeadlines = (params?: UseDeadlinesParams) => {
-  const options = params?.groupId || {}
-  const { loading, data } = useGetAllDeadLinesQuery(options)
+export const useDeadlines = (params: UseDeadlinesParams) => {
+  const { groups } = useWorkingGroups()
+  const { active: activeMembership } = useMyMemberships()
+  const groupIds = useMemo(
+    () => groups.flatMap((group) => (group.leadId === activeMembership?.id ? group.id : [])),
+    [activeMembership, groups.length]
+  )
```
thesan, line 32: Please use the `params.member`, also to avoid to call `useGetAllDeadLinesQuery` multiple times the hook should wait for the groups to be loaded
```suggestion
  const { isLoading: isLoadingWorkingGroup, groups } = useWorkingGroups()
  const groupIds = useMemo(
    () => groups.flatMap((group) => (group.leadId === params.member.id ? group.id : [])),
    [params.member.id, groups.length]
  )
```

packages/ui/src/overview/hooks/useDeadlines.ts
```diff
@@ -1,26 +1,42 @@
...
+  const variables = {
+    proposalCreator: params.member.isCouncilMember ? undefined : { id_eq: params.member.id },
+    isLead: groupIds.length > 0,
+    group: groupIds.length > 0 ? { id_in: groupIds } : undefined,
+  }
+  const { stage: electionStage } = useElectionStage()
+  const { loading, data } = useGetAllDeadLinesQuery({ variables })
```
thesan, line 39:
```suggestion
  const { loading, data } = useGetAllDeadLinesQuery({ variables, skip: isLoadingWorkingGroup })
```

As a result the returned value for `isLoading` should now be:
```ts
  isLoading: isLoadingWorkingGroup || loading
```

packages/ui/src/overview/components/DeadlineList/ElectionListItem.tsx
```diff
@@ -48,39 +57,50 @@ export const ElectionListItem: React.FC<ElectionListItemProps> = React.memo(({ e
...
+        }
         return (
           <StyledText>
-            {t('deadline.votingPeriod', { electionName: title })}{' '}
-            <StyledLink href={urlAddress + electionId}>{t('deadline.votingLink')}</StyledLink>
+            {t('deadline.votingPeriod')} <StyledLink href={urlAddress}>{t('deadline.votingLink')}</StyledLink>
           </StyledText>
         )
       case 'revealing':
+        if (votes?.every((vote) => vote.voteFor)) {
```
thesan, line 83: That's a small detail but there shouldn't be a `?` here because if `votes` was `undefined` (which it can't be) this logic would be wrong
```suggestion
        if (votes.every((vote) => vote.voteFor)) {
```

thesan (changes requested):
Almost done ! just a few optimizations of `useDeadlines` and one more detail on `ElectionListItem`

packages/ui/src/overview/components/DeadlineList/ElectionListItem.tsx
stepien4soft: I think this hook ```const { votes } = useMyCastVotes(electionStage === 'revealing' ? election.cycleId : undefined) ?? []``` could return undefined value

thesan (approved):
I've update the PR directly, please check my changes and if it's all good with you merge it

Issue #2419, opened 2022-03-04
[USERSNAP] Text says "opening to cancel" instead of "select applicant to hire" (Fill WG lead proposal for forum WG)


dmtrjsg:
## Scope

- [ ] update placeholder copy for _opening_ selection to "Choose opening to fill"
- [ ] update application copy to "Choose application"

______________

**Taken at**: https://pioneer-2.vercel.app/#/proposals/current
**Fields**:
- Houston, we have a problem.: Text says &#34;opening to cancel&#34; instead of &#34;select applicant to hire&#34; (Fill WG lead proposal for forum WG)
- Give us some details: Text says &#34;opening to cancel&#34; instead of &#34;select applicant to hire&#34; (Fill WG lead proposal for forum WG)
- Text says &#34;opening to cancel&#34; instead of &#34;select applicant to hire&#34; (Fill WG lead proposal for forum WG)
- Discrod Handle: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/e2e21f13-5b47-4be9-bc9a-03d200f44ded-annotated_7edf4cf0-7869-489c-b9cd-683f325172ff.png?etag=a5fb19e3cb033258f046c37522b35467)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/e2e21f13-5b47-4be9-bc9a-03d200f44ded-annotated_7edf4cf0-7869-489c-b9cd-683f325172ff.png?etag=a5fb19e3cb033258f046c37522b35467)*

**Browser**: Chrome 98 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1718x1306
**[Open #47 in Usersnap](https://app.usersnap.com/l/feedback/6bd8261a-7f10-44cd-ad1d-eb8a8f4d4764?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

singulart:
@dmtrjsg can this be moved to ready for test?

dmtrjsg:
@singulart yep, can be tested on staging. all tickets with label:stabilisation4 are ready for test!

polikosi:
Verified on staging.  Test evidences below: 
[...]

PR #2902, opened 2022-04-24
2419 - Pass placeholder as props for SelectWorkingGroupOpening.

oleksanderkorn:
https://github.com/Joystream/pioneer/issues/2419

image

### DIFF
--- a/packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/WorkingGroupLeadOpening/CancelWorkingGroupLeadOpening.tsx
+++ b/packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/WorkingGroupLeadOpening/CancelWorkingGroupLeadOpening.tsx
@@ -39,7 +39,7 @@ export const CancelWorkingGroupLeadOpening = ({
           >
             <SelectWorkingGroupOpening
               id="opening"
-              placeholder="Choose application"
+              placeholder="Choose opening"
               openingsPositionType="LEADER"
               onChange={(selected) => {
                 setGroupId(selected.groupId)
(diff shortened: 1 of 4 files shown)


oleksanderkorn:
@dmtrjsg i set the placeholder for Cancel Working Group Lead Opening to `Choose opening` instead `Choose application` i think this is correct. Let me know if you disagree.

thesan:
@oleksanderkorn what we meant was that the placeholder of `<SelectWorkingGroupApplication />` should be `Choose application` but the placeholder for `<SelectWorkingGroupOpening />` in `CancelWorkingGroupLeadOpening` should probably be `Choose opening to cancel`

oleksanderkorn:
@thesan  @dmtrjsg 
image

thesan:
Screenshot from 2022-04-25 21-27-50

@oleksanderkorn yes I just realized it's not at all visible on the screenshot of the issue but the `SelectWorkingGroupApplication`'s placeholder is wrong too.

Because we had not realized that the placeholder were hardcoded, this issue was supposed to only be about "Fill WG lead proposal", that's why the proposed copy were "Choose opening to fill" for the opening selector and "Choose application" for the application selector.

oleksanderkorn:
> Because we had not realized that the placeholder were hardcoded, this issue was supposed to only be about "Fill WG lead proposal", that's why the proposed copy were "Choose opening to fill" for the opening selector and "Choose application" for the application selector.

@thesan I have also made the placeholder for SelectWorkingGroupApplication a property and updated it to the correct text:
image

singulart (approved):

thesan (approved):
Thank you for the PR, and for the extra effort to expose the application select placeholder :pray:

Issue #2919, opened 2022-04-26
Proposal Loading Time fix


dmtrjsg:
About this particular **long loading time on the proposal page** it turns out to be due to a particularly slow chain query in the useHasMemberVotedOnProposal hook, which is send for each current proposal and end up overloading the chain connection completely. It happens on the proposal page when a member is picked. In addition reloading the page doesn't cancel the queries properly which causes long Connecting to Joystream node after it.

ivanturlakov:
Verified on staging.

The process of creating the proposal went smoothly.
After selecting an account, there is a slight delay before the next step button becomes active.

https://user-images.githubusercontent.com/46903215/167621509-d4509db0-cf79-449c-903c-71bcc208b0de.mov

PR #2922, opened 2022-04-26
🏎️ Hotfix for performance on checking member proposal votes 

WRadoslaw:
#2919


thesan (changes requested):
That's a way better solution but since the votes are directly available from the `Proposal` schema, could you please add:
```gql
votes {
  voter { id }
}
```
to the `ProposalFields` fragment and just remove `useHasMemberVotedOnProposal` altogether

thesan:
As we spoke about: We need to check that the votes were cast by the current council only (proposals with `CONSTITUTIONALITY > 1` requires more than 1 council to approve them). The good news is `ProposalVotedEvent` has a `votingRound` field. So we can either:
- check if the `vote.votingRound` equals the number of time the proposal has been in the Deciding status (or maybe there's a better way to know the current voting round). 
- or check `vote.inBlock` is greater than the current council elected `inBlock` (this one sounds more hacky though)

Issue #2918, opened 2022-04-26
Do not block app interactions on waiting the transaction to get signed


dmtrjsg:

dmtrjsg:
Radek: Hey, about #2918, it seems like we do not block app interactions neither when app wait for tx to get signer or after it, when we wait for tx execution

@traumschule can you pls point to the specific place where you experienced the app blocked while TX is signed?

traumschule:
not 100% if the app is actually blocked but i experience for some transactions that when i switch to another component the ongoing transaction is somehow lost, does not finish or pioneer is unable to pick up QN events (my interpretation) and i have to reload the whole thing. that's why i usually just leave the tab and do something else - so pioneer is "blocked" in my mind.

PR #2935, opened 2022-04-28
🎲 Improvement on refetching after app interaction

WRadoslaw:
#2918


thesan (changes requested):
As a reminder this PR shouldn't be merged now. Instead of the `useSignAndSendTransaction` changes, the Apollo fetch policies should be fixed

PR #2940, opened 2022-04-29
🔧 Fix local mocks

### LINKED ISSUE #2893: `local-mocks` is broken
When new data get generated (usually for the tests), the `local-mocks` network sometimes breaks.

IMO we should keep this network because it can significantly simplify the development.

However to keep it useful we should:
- [ ] #3022
- [x] Fix the `local-mocks` network
- [x] Don't show the `local-mocks` option in production

RafalPloszka:
#2893


packages/ui/src/common/hooks/useNetwork.ts
```diff
@@ -12,7 +12,7 @@ export const useNetwork = () => {
   const networks = useMemo<NetworkType[]>(
     () => [
       'local',
-      'local-mocks',
+      ...(IS_IN_DEV_ENVIRONMENT ? ['local-mocks' as const] : []),
```
thesan: So this should work:
```suggestion
      ...(IS_DEVELOPMENT ? ['local-mocks' as const] : []),
```

thesan (approved):
Nice job the `local-mocks` is back :rocket:. I just fixed the logic to hide it in production mode. And the test to check that it doesn't break again can be added in a follow-up PR

Issue #2889, opened 2022-04-22
POC: choose lib for loading states


dmtrjsg:
Define the library to handle better loading states Epic

> 

- [ ] Lib handles categories (one field at a time)

dmtrjsg:
implemented on my profile/ accounts

ivanturlakov:
Verified on staging.
The loading state looks good.

https://user-images.githubusercontent.com/46903215/168566732-be0c1650-41a7-46c2-854a-87c0cd61e34f.mov

dmtrjsg:
Token value changes turnes null returnable values into perpetual load.

dmtrjsg:
@singulart no need to upd the reports, this is a nice issue (which is also hard to find) and we will handle it :)

PR #2956, opened 2022-05-02
🙈 POC: Loading state on accounts list

WRadoslaw:
#2889

### DIFF
--- a/packages/ui/src/accounts/components/Accounts.tsx
+++ b/packages/ui/src/accounts/components/Accounts.tsx
@@ -1,12 +1,14 @@
 import React, { ReactNode, useMemo, useState } from 'react'
 import styled from 'styled-components'
 
-import { List, ListItem } from '../../common/components/List'
-import { Loading } from '../../common/components/Loading'
-import { ContentWithTabs } from '../../common/components/page/PageContent'
-import { HeaderText, SortIconDown, SortIconUp } from '../../common/components/SortedListHeaders'
-import { Tabs } from '../../common/components/Tabs'
-import { Colors } from '../../common/constants'
+import { List, ListItem } from '@/common/components/List'
+import { ListItemLoader } from '@/common/components/ListItemLoader'
+import { ColumnGapBlock, ContentWithTabs, RowGapBlock } from '@/common/components/page/PageContent'
+import { Skeleton } from '@/common/components/Skeleton'
+import { HeaderText, SortIconDown, SortIconUp } from '@/common/components/SortedListHeaders'
+import { Tabs } from '@/common/components/Tabs'
+import { Colors } from '@/common/constants'
+
 import { useMyAccounts } from '../hooks/useMyAccounts'
 import { useMyBalances } from '../hooks/useMyBalances'
 import { filterAccounts } from '../model/filterAccounts'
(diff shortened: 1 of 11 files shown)


packages/ui/src/accounts/components/Accounts.tsx
```diff
@@ -29,10 +31,6 @@ export function Accounts() {
     [visibleAccounts, balances, sortBy, isDescending]
   )
 
-  if (!hasAccounts) {
-    return <Loading />
-  }
```
thesan, line 34: IMO it would be an easier pattern to follow for the community if: 
```jsx
<ListItemLoader columnsTemplate="276px repeat(4, 128px) 104px" count={4}>
  ...
</ListItemLoader>
```
Was moved in a separate component called e.g `AccountListLoading` and called here, although you're right `!hasAccounts` should be `isLoading`.

This way the community developer will just replace the `<Loading>` components by `<[Custom]Loading>` components created, and it will be easier for us to review too

packages/ui/src/common/components/ListItemLoader.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const ListItemLoader = ({ children, count = 1, ...styleProps }: ListItemLoaderProps) => {
+  return (
+    <>
+      {Array(count)
+        .fill(null)
+        .map((_, index) => (
+          <Wrapper key={index} {...styleProps}>
+            {children}
+          </Wrapper>
+        ))}
```
thesan, line 20:
```suggestion
      {repeat(
        (index) => (
          <Wrapper key={index} {...styleProps}>
            {children}
          </Wrapper>
        ),
        count
      )}
```

packages/ui/src/common/components/statistics/TokenValueStat.tsx
```diff
@@ -11,12 +12,17 @@ import { StatisticItem, StatisticItemProps } from './StatisticItem'
...
 
 export const TokenValueStat: FC<TokenValueStatProps> = (props) => {
   return (
     <StatisticItem {...props}>
-      <TotalValue value={props.value} textColor={props.textColor} />
+      {!props.isLoading ? (
+        <TotalValue value={props.value} textColor={props.textColor} />
+      ) : (
+        <Skeleton variant="rect" height="32px" width="50%" />
+      )}
```
thesan, line 25: `TotalValue` extends `TokenValue` and `TokenValue` shows `-` if `value` is `null` or `undefined`.

AFAIK this `-` only shows when the `value` is loading. I think it would be better to render the `<Skeleton />` inside `TokenValue` instead of `-`. In this case the `height` and `width` could be set in the parent components style.
E.g:
```ts
export const TotalValue = styled(TokenValue)<{ textColor?: string }>`
  font-size: 20px;
  line-height: 28px;
  color: ${(props) => props.textColor || Colors.Black[900]};

  ${Skeleton} {
    height: 32px;
    width: 50%;
  }
`
```

WDYT ?

thesan (commented):
It looks really good visually :+1: 
I thinks not going with a third party library is a good decision.

In addition to the implementations suggestions, I think it would be good to add a storybook to demonstrate the `Skeleton` component a bit.

thesan (changes requested):
It looks nice but the skeleton doesn't match when transitioning from loading the accounts (from the extension) to loading the individual value (from the chain):
 full
partial
 The rectangles get narrower and taller. IMO the longer rectangle from the first view look better. I think the `${Skeleton}` width and height could be set in the `AccountsWrap` style.

Also the statistics still show `0` while loading but I'm not sure whether this should be fixed in this PR

thesan (approved):
Looks great :raised_hands: !

Issue #2689, opened 2022-03-26
[USERSNAP] Vertical scroller on empty screen


dmtrjsg:
## Context

In the guest user sign up flow, when the Sign up modal in the top of the screen is visible, the vertical scroll appears where it should not (without any polkadot accounts created/visible/ recognised by the browser)

## Scope

Make the onboarding modal top bar function as a sticky header - when user is scrolling the page the panel is affixed to the top of the page and scrollable content goes underneath it.

__________
**Taken at**: https://dao.joystream.org/#/election/past-elections
**Fields**:
- Houston, we have a problem.: Vertical scroller on empty screen
- Give us some details: There is a vertical scroller, which seems to be generated due ot the &#34;Become a member&#34; header, despite there not being any content on the screen.
- Not sure
- Discord Handle: bedeho

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/b6e5ac8c-9757-460f-865f-a4df9f0cafed-annotated_b7e3d549-9f87-4db5-a1a2-f1b84febb109.png?etag=6be0fbbf15211442b42835eaccf46943)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/b6e5ac8c-9757-460f-865f-a4df9f0cafed-annotated_b7e3d549-9f87-4db5-a1a2-f1b84febb109.png?etag=6be0fbbf15211442b42835eaccf46943)*

**Browser**: Firefox 98 (macOS Catalina)
**Screen size**: 5120x1440
**Browser size**: 1765x812
**Labels**: bug-non-critical
**[View screen recording](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/screen_recording/eb8a81ce-d6ea-4d19-a187-909f43902736_screen_recording_0faa9b50-ae03-4b8f-938b-730e189b58f9.webm?etag=1445932010a799e0730890d0d94473c5)**
**[Open #143 in Usersnap](https://app.usersnap.com/l/feedback/9d4e83f1-3eb6-4de0-b6ae-29061993dc70?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

[...]

PR #2958, opened 2022-05-03
Fix Vertical scroller on empty screen

talentdev0807:
#2689
This bug comes from adding `Become a member` header
Originally, main body height is 100%, it is natural to have scroll when we add it on top screen
So, I made `Become a member` header as sticky header

### DIFF
--- a/packages/ui/src/app/components/OnboardingOverlay/OnBoardingOverlay.tsx
+++ b/packages/ui/src/app/components/OnboardingOverlay/OnBoardingOverlay.tsx
@@ -11,7 +11,7 @@ import { StepperStep } from '@/common/components/Stepper'
 import { HorizontalStepper } from '@/common/components/Stepper/HorizontalStepper'
 import { VerticalStaticStepper } from '@/common/components/Stepper/VerticalStaticStepper'
 import { TextHuge, TextSmall } from '@/common/components/typography'
-import { Colors } from '@/common/constants'
+import { Colors, ZIndex } from '@/common/constants'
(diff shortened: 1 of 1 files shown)


singulart:
Fixes https://github.com/Joystream/pioneer/pull/2958

singulart:
Fixes https://github.com/Joystream/pioneer/issues/2689

packages/ui/src/app/components/OnboardingOverlay/OnBoardingOverlay.tsx
```diff
@@ -79,46 +79,57 @@ export const OnBoardingOverlay = () => {
...
+const TopSpace = styled.div`
+  width: 100%;
+  height: 90px;
+`
+
 const MainWrapper = styled.div`
-  position: relative;
+  position: fixed;
+  top: 0;
+  width: 100%;
```
stepien4soft, line 125: Hi @talentdev0807, First of all thank you for this PR. Just one small fix has to be added - for now <ButtonPrimary>Join now </ButtonPrimary> isn’t visible. Two quick solution is to change width for a main wrapper or set up padding on Join now button

stepien4soft (changes requested):
Hi @talentdev0807, First of all thank you for this PR. Just one small fix has to be added - for now <ButtonPrimary>Join now </ButtonPrimary> isn’t visible. Two quick solution is to change width for a main wrapper or set up padding on Join now button

stepien4soft:
Approved! Thanks for that

stepien4soft (approved):

Issue #2884, opened 2022-04-21
[USERSNAP] Deleting a post in a single-post thread doesn't delete the thread and shows "-1" in replies column


dmtrjsg:
## Scope
- [ ] For threads, ON THE VERY FIRST POST (original thread) remove the button/ ability to delete the post. 

ℹ️  FYI delete thread button (bin icon next to its title) is designed for thread deletion.

_____

**Taken at**: https://dao.joystream.org/#/forum/category/12
**Fields**:
- Houston, we have a problem.: Deleting a post in a single-post thread doesn&#39;t delete the thread and shows &#34;-1&#34; in replies column
- Comment: Thread should be deleted when its only post is deleted.

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/3bf5a317-8c08-41bb-8234-4695471201b1-annotated_50ad0588-484a-4907-b1de-4b58cdcdfa8a.png?etag=7b1699f0d2c09b4d198588294379d1ee)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/3bf5a317-8c08-41bb-8234-4695471201b1-annotated_50ad0588-484a-4907-b1de-4b58cdcdfa8a.png?etag=7b1699f0d2c09b4d198588294379d1ee)*

**Browser**: Chrome 100 (macOS Catalina)
**Screen size**: 1440x900
**Browser size**: 1440x722
**[Open #216 in Usersnap](https://app.usersnap.com/l/feedback/8010eb8d-6d90-42c3-8a16-e2655ecce135?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

polikosi:
The button to delete the post ON THE VERY FIRST POST is removed
![Screenshot_14](https://user-images.githubusercontent.com/80429191/168779736-a4c0205d-c3ce-4a31-abd9-6f52e92c708c.png)

But when I delete the threads with one original post, then the "Latest thards" tab displays replies -1

https://user-images.githubusercontent.com/80429191/168781740-3b14704b-d642-47d4-95db-01cc6cc1851f.mp4

singulart:
[...]

PR #2960, opened 2022-05-03
2884: Remove delete post action item from the first post in the thread

oleksanderkorn:
https://github.com/Joystream/pioneer/issues/2884

### DIFF
--- a/packages/ui/src/forum/components/PostList/PostList.tsx
+++ b/packages/ui/src/forum/components/PostList/PostList.tsx
@@ -57,8 +57,9 @@ export const PostList = ({ threadId, isThreadActive, isLoading, replyToPost, isD
   return (
     <Wrapper gap={24}>
       {!isDiscussion && <Pagination pageCount={pageCount} handlePageChange={setPage} page={page} />}
-      {posts.map((post) => (
+      {posts.map((post, index) => (
         <PostListItem
+          index={index}
           key={post.id}
           post={post}
           insertRef={getInsertRef(post.id)}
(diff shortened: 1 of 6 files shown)


packages/ui/src/forum/components/PostList/PostList.tsx
```diff
@@ -57,8 +57,9 @@ export const PostList = ({ threadId, isThreadActive, isLoading, replyToPost, isD
   return (
     <Wrapper gap={24}>
       {!isDiscussion && <Pagination pageCount={pageCount} handlePageChange={setPage} page={page} />}
-      {posts.map((post) => (
+      {posts.map((post, index) => (
         <PostListItem
+          index={index}
```
thesan, line 62: The problem is that since the posts are paginated the index 0 might not be the "original thread" but just the first post on page 2. IMO a better approach would be to pass an optional flag like `isPostOrigin` to the `<PostListItem />`

This could be queried through `ForumPost.origin (...on PostOriginThreadInitial)`:
Screenshot from 2022-05-09 09-56-18

(Although there might be a simpler way)

thesan (changes requested):
Thank you for the PR. I suggested a slightly different approach to account for longer threads

packages/ui/src/forum/components/PostList/PostList.tsx
oleksanderkorn: @thesan  Maybe it is easier to also use the page number together with the post index to make sure this is the very first post?

oleksanderkorn: Page 1:
image
Page 2:
image

PR #2962, opened 2022-05-04
🪢 Show bound account <> membership info

### LINKED ISSUE #2521: Show bound account <> membership info
Screenshot 2022-03-10 at 19 45 47

Screenshot 2022-03-10 at 19 45 50

RafalPloszka:
Closes #2521

### DIFF
--- a/packages/ui/src/memberships/components/MemberProfile/MemberAccounts.tsx
+++ b/packages/ui/src/memberships/components/MemberProfile/MemberAccounts.tsx
@@ -21,6 +21,17 @@ export const MemberAccounts = ({ member }: { member: Member }) => (
         <UnknownAccountInfo address={member.controllerAccount} placeholderName="Controller Account" />
       </AccountRow>
     )}
+
+    {!!member.boundAccounts.length && (
+      <>
+        <SidePaneLabel text="Bound accounts" />
+        {member.boundAccounts.map((account) => (
+          <AccountRow key={account}>
+            <UnknownAccountInfo address={account} placeholderName="Bound Account" />
+          </AccountRow>
+        ))}
+      </>
+    )}
   </AccountsDisplay>
 )
 


packages/ui/src/bounty/queries/bounty.graphql
```diff
@@ -170,3 +170,14 @@ query GetBountyContributors(
...
+query GetLatestBountyByMemberId($memberId: ID) {
+  bounties(where: { creator: { id_eq: $memberId } }, orderBy: [createdAt_DESC], limit: 1) {
+    id
+    createdInEvent {
+      createdAt
+      inBlock
+      network
+    }
+  }
+}
```
thesan, line 183: AFAIK there is no lock when creating a bounty. The bounty lock comes from anouncing an entry so the query should be:
```gql
bountyEntries(where: { stakingAccount_eq: $lockAccount }, orderBy: [createdAt_DESC], limit: 1) ...
```

thesan (changes requested):
Overall the PR looks good. But in addition to the 2 comments, there is some general improvements that would be needed:
1. The locks info are queried using the `membershipId` but whenever possible it would be a lot simpler to query them using the account addresses directly, e.g:
```gql
memberInvitedEvents (where: { controllerAccount_eq: $lockAccount }, orderBy: [createdAt_DESC], limit: 1)
proposals(where: { stakingAccount_eq: $lockAccount }, orderBy: [createdAt_DESC], limit: 1)
bountyEntries(where: { stakingAccount_eq: $lockAccount }, orderBy: [createdAt_DESC], limit: 1)
```
2. There is a lot of boilerplate in the `EntityLockItem` that would make it harder to change/maintain. A solution to this would be to have a `LockItem` component which takes the a block for the lock date and the rest of the dropdown as parameters. E.g in `VoteLockItem` it could be used like this (the default props values could even replace the `DefaultLockItem`):
```tsx
return <LockItem block={eventData}>
  <div>
    <LockRecoveryTime value={faker.date.soon(1).toISOString()} />
  </div>
  <BalanceAmount amount={lock.amount} isRecoverable={isRecoverable} />
  <LocksButtons>
    <LockLinkButton label="Show Council" onClick={goToCouncil} />
    {recoverButton}
  </LocksButtons>
</LockItem>
```
3. Finally its more of a detail but it would be easier to understand the code if there was a folder with only the different locks item types. I.e `LockDate`, `LockLinkbutton`, ... (and also the `LockItem` I suggested in `2.`) would be in a different folder.

RafalPloszka:
@thesan thank you for your feedback, I agree with all your comments and fixed them all 👷 
I made `<LockItem />` even more general than you suggested and it indeed cut off a lot of boilerplate. Apart from mentioned issues I also added some minor improvements and renamings, hope it's better now

packages/ui/src/council/queries/council.graphql
```diff
@@ -379,3 +379,16 @@ query GetPastCouncilStats($startBlock: Int!, $endBlock: Int!) {
     ...CouncilSpendingEventFields
   }
 }
+
+query GetCouncilorElectionEvent($lockAccount: String) {
+  memberships(where: { controllerAccount_eq: $lockAccount }, orderBy: [createdAt_DESC], limit: 1) {
+    councilMembers {
```
thesan, line 385: Unlike with `memberInvitedEvents` the `controllerAccount` might not be the staking account here:
```suggestion
  councilMembers (where: { stakingAccountId_eq: $lockAccount }, orderBy: [createdAt_DESC], limit: 1) {
```

PR #2964, opened 2022-05-04
🎲 Improvement on refetching after app interaction

WRadoslaw:
context #2918


thesan (approved):
Good job finding this Apollo bug :raised_hands:

PR #2967, opened 2022-05-04
Fix/2864-Next button is outside of screen

talentdev0807:
Bug : 'Next' button is outside of the screen when pasting large text
Done : Made Thread Modal scrollable


WRadoslaw (commented):
Hello, firstly thank you for your contribution! I have two suggestions though. First one, I think it would be good to keep changes of another fix in one PR, it looks like you've based this branch on your previous fix, which in some cases might result in reviewer checking same code twice. Second, there is no need to override `ModalBody` styles, since there is  already `ScrolledModalBody` that is utilized throughout the Pioneer, so I think we shouldn't approach it in new way for code uniformity sake.

Issue #2712, opened 2022-03-28
[USERSNAP] Parse markdown in Posts, Applications, Profiles


dmtrjsg:
## Scope

Add markdown interpretation in all fields initially created with CK Editor (Description/ Forum posts etc)

- [ ] Interpret internal identifiers
- [ ] external links
- [ ] headings in text fields
- [ ] Line breaks

⚠️  We do not use markdown in Field form Entries (like titles etc..)
__________
**This should cover all text fields, but specifically:**

- Proposal/ bounty discussion
- Forum post
- ~Widgets with text shown in them~
- Application
- proposals descriptions
- proposal rationales
- wg roles applications
- election candidates 
- bounty descriptions #3218 
- work submitted
- Oracle rationale for judgement

Out of scope:

- Widgets with text shown in them

## Context
____

**Taken at**: https://dao.joystream.org/#/working-groups/openings/operationsWorkingGroupAlpha-1
**Fields**:
- Houston, we have a problem.: Parse markdown in Posts, Applications, Profiles
- Give us some details: There are several places where user can input text but markdown links and headings are usually not interpreted.
- This is a downgrade from previous Pioneer.

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/061695b3-cdae-4e23-b44a-7945a7410038-annotated_c8d11cac-774e-45bc-84c7-e5aa7f351d45.png?etag=b54b9d58e8990b4cae665c375683247a)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/061695b3-cdae-4e23-b44a-7945a7410038-annotated_c8d11cac-774e-45bc-84c7-e5aa7f351d45.png?etag=b54b9d58e8990b4cae665c375683247a)*

**Browser**: Chrome 99 (Linux)
**Screen size**: 1920x1080
**Browser size**: 2390x1216
**Labels**: enhancement
**[Open #159 in Usersnap](https://app.usersnap.com/l/feedback/4d567d1d-21cf-4a08-a6f7-e997f69872e7?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

traumschule:
[...]

PR #3002, opened 2022-05-09
fix/2712-parse markdown forum post

talentdev0807:
#2712


singulart:
@talentdev0807 Please explain the purpose of this change.

Issue #2423, opened 2022-03-04
[USERSNAP] Forum: "Keep editable" is not noticable


dmtrjsg:
## Scope
- [ ] Keep editable must be pre-selected by default

## Context

**Taken at**: https://pioneer-2.vercel.app/#/forum/thread/1
**Fields**:
- Houston, we have a problem.: Forum: &#34;Keep editable&#34; is not noticable
- Give us some details: The decision to keep a post editable or not has a major impact for users, this option should be more obvious or prominent to users somehow.
- The decision to keep a post editable or not has a major impact for users, this option should be more obvious or prominent to users somehow.
- Discrod Handle: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/599bb459-29f6-4ee5-bc70-66ecf67c3757-annotated_9cc0452b-3157-4e69-9feb-91969ca57b98.png?etag=efb851edd94763a747207e9a4ac3d1b1)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/599bb459-29f6-4ee5-bc70-66ecf67c3757-annotated_9cc0452b-3157-4e69-9feb-91969ca57b98.png?etag=efb851edd94763a747207e9a4ac3d1b1)*

**Browser**: Chrome 98 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1718x1306
**[Open #51 in Usersnap](https://app.usersnap.com/l/feedback/d9369adf-1521-4bd5-bfba-a69a1baa8e34?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

talentdev0807:
@talentdev0807

polikosi:
Verified on staging. Test evidence below:

![Screenshot_28](https://user-images.githubusercontent.com/80429191/169647327-0c2e1a57-9d7a-47e1-957a-b3eff762c7fb.png)

PR #3037, opened 2022-05-12
fix/2423 Keep editable check button is not noticeable in Forum

talentdev0807:
#2423


singulart (approved):

singulart (approved):

traumschule:
checked preview: It is not more noticeable but `keep editable` is auto-on.

thesan:
> checked preview: It is not more noticeable but `keep editable` is auto-on.

Yes it was decided that this is the scope of the issue https://github.com/Joystream/pioneer/issues/2423#issue-1159159271

Issue #2849, opened 2022-04-12
[USERSNAP] "Not voted" section in Proposal view won't expand


dmtrjsg:
## Scope

- [ ] Fix expanding the Not Voted section by clicking the chevron in Current Proposals
- [ ] Hide the chevron that expands the field on past proposals

Out of scope:
- [x] #2941

## Tech Guidelines

Fetch council elected at the time when this proposal was / then check filter which council is not in the votes.

<img width="317" alt="Screenshot 2022-04-22 at 12 27 29" src="https://user-images.githubusercontent.com/93924525/164705913-982b2b45-7c8e-471a-b9b1-b67c89de5895.png">

____
**Taken at**: https://dao.joystream.org/#/proposals/preview/84
**Fields**:
- Houston, we have a problem.: &#34;Not voted&#34; section in Proposal view won&#39;t expand
- Comment: Section should expand and show non-voted CMs
- Not sure
- Feedback: isonar#5236

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/084c5403-d6f1-427a-81c4-e8e99ca806d6-annotated_414120bf-c336-4dd0-af96-e73fc7b8885c.png?etag=a0e968dcdd1e69487d6bad2d05d1ad1e)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/084c5403-d6f1-427a-81c4-e8e99ca806d6-annotated_414120bf-c336-4dd0-af96-e73fc7b8885c.png?etag=a0e968dcdd1e69487d6bad2d05d1ad1e)*

**Browser**: Chrome 100 (macOS Catalina)
**Screen size**: 1440x900
**Browser size**: 1600x794
**[View screen recording](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/screen_recording/344a22c3-adb5-4310-a977-98cd03c27905_screen_recording_6a45bade-678d-4eb6-9e3a-3cc1cf65d754.webm?etag=aed556f8c574b09400e8bf13e646c523)**
**[Open #208 in Usersnap](https://app.usersnap.com/l/feedback/31b6530c-3786-4331-83cf-7ba79a495968?utm_medium=integration&utm_campaign=integration&utm_content=open)**

[...]

PR #3044, opened 2022-05-13
🧷 Fix on proposal not voted section

WRadoslaw:
#2849

### DIFF
--- a/packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
+++ b/packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import { useLocation } from 'react-router'
 import { useParams } from 'react-router-dom'
 import styled from 'styled-components'
@@ -18,6 +18,7 @@ import { camelCaseToText } from '@/common/helpers'
 import { useModal } from '@/common/hooks/useModal'
 import { formatBlocksToDuration, formatTokenValue } from '@/common/model/formatters'
 import { getUrl } from '@/common/utils/getUrl'
+import { useElectedCouncil } from '@/council/hooks/useElectedCouncil'
 import { MemberInfo } from '@/memberships/components'
 import { useMyMemberships } from '@/memberships/hooks/useMyMemberships'
 import { ProposalDetails } from '@/proposals/components/ProposalDetails/ProposalDetails'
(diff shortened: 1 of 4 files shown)


packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
```diff
@@ -50,6 +52,21 @@ export const ProposalPreview = () => {
...
+      return
+    }
+
+    const votedMembers = Array.from(votes.map.values())
+      .flat()
+      .map((vote) => vote.voter)
+    const councilMembers = council?.councilors.map((councilor) => councilor.member)
+    return (
+      councilMembers?.filter(
+        (member) => !votedMembers.some((voted) => voted.id === member.id) && member.id !== proposal.proposer.id
```
thesan, line 66: I think the proposers should appear in "not voted", since they can also vote on their own proposals. e.g https://dao.joystream.org/#/proposals/preview/184

thesan (changes requested):

packages/ui/src/proposals/components/VotesPreview/VotesPreview.tsx
```diff
@@ -12,15 +13,16 @@ const { Approve, Reject, Slash, Abstain } = ProposalVoteKind
...
 
-export const VotesPreview = ({ votes }: VotesPreviewProps) => (
+export const VotesPreview = ({ votes, notVoted }: VotesPreviewProps) => (
   <VotesContainer>
     <VotePreview kind={Approve} count={votes.count.approve} votes={votes.map.get(Approve)} />
     <VotePreview kind={Reject} count={votes.count.reject} votes={votes.map.get(Reject)} />
     <VotePreview kind={Slash} count={votes.count.slash} votes={votes.map.get(Slash)} />
     <VotePreview kind={Abstain} count={votes.count.abstain} votes={votes.map.get(Abstain)} />
-    <VotePreview kind="Not Voted" count={votes.count.remain} />
+    <VotePreview kind="Not Voted" count={notVoted?.length ?? 0} votes={notVoted?.map((voter) => ({ voter })) ?? []} />
```
thesan, line 25: One last thing:
```suggestion
    <VotePreview kind="Not Voted" count={notVoted?.length ?? votes.count.remain} votes={notVoted?.map((voter) => ({ voter })) ?? []} />
```
I think defaulting to `votes.count.remain` is a better temporary solution. Because showing `0` is (most of the time) a false information [e.g here](https://dao-git-fix-not-voted-section-joystream.vercel.app/#/proposals/preview/3). While `votes.count.remain` should be correct at least until we address #2941 (i.e the size of the council shouldn't change before that)

thesan (approved):

Issue #2294, opened 2022-02-22
[USERSNAP] tomato#9464


dmtrjsg:
**Taken at**: https://pioneer-2.vercel.app/#/proposals/past
**Fields**:
- Discrod Handle: tomato#9464
- Give us some details: The proposals aren&#39;t current ordered by date/time, this is quite confusing
- Houston, we have a problem.: Past proposals should list in chronological order (most recent first)
- The proposals aren&#39;t current ordered by date/time, this is quite confusing

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/914700de-b095-4984-9bdd-09979d89c2dd-annotated_7bf2af70-2d68-484a-b791-1218cf996b24.png?etag=03e3da160d3e4b8758475f155aced667)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/914700de-b095-4984-9bdd-09979d89c2dd-annotated_7bf2af70-2d68-484a-b791-1218cf996b24.png?etag=03e3da160d3e4b8758475f155aced667)*

**Browser**: Chrome 98 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 2022x1319
**[Open #9 in Usersnap](https://app.usersnap.com/l/feedback/8d1e079a-4fc4-460b-b868-49f61c3c3a26?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

traumschule:
From #2771: Ideally with truncated rationale preview (40..80 chrs) as last column on wider screens, hidden on mobile - or if less effort as third line styled like `Ended at` below the title.

traumschule:
can we please up this. costs me time every time. candidate for comm-dev?

traumschule:
Shortcut `This term` would be great. https://github.com/Joystream/pioneer/issues/2914#issuecomment-1119294707

polikosi:
Current:
✅ Newest on top (default)
![Screenshot_43](https://user-images.githubusercontent.com/80429191/171627710-f9889aa1-63b7-4328-bd6a-562448e2d7da.png)

🚫Oldest on top (I can't sort by date, there are no fields or buttons for this)
[...]

Issue #2771, opened 2022-04-02
Preview rationnale in proposal lists


dmtrjsg:
Display a truncated rationale preview (40..80 chars) as the last column

![Screenshot from 2022-05-20 11-59-59](https://user-images.githubusercontent.com/6571453/169505208-b137358c-f622-4a0f-b20e-d047c62bb8dd.png)

---
**Taken at**: https://dao.joystream.org/#/proposals/past
**Fields**:
- Houston, we have a problem.: Reverse propsal order
- Comment: newest first
- ideally with a an truncated rationale preview (40..80 chrs) as last column for wider screens, hidden on mobile - or if less effort as third line styled like `Ended at` below the title

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/d221c9f2-0ad3-4ce6-b8a2-1f0694056b97-annotated_01d240b4-f1ab-4264-b459-886ef0b74daf.png?etag=a5d7b6e4819da6d9d8bf07ea321174b1)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/d221c9f2-0ad3-4ce6-b8a2-1f0694056b97-annotated_01d240b4-f1ab-4264-b459-886ef0b74daf.png?etag=a5d7b6e4819da6d9d8bf07ea321174b1)*

**Browser**: Chrome 99 (Linux)
**Screen size**: 1920x1080
**Browser size**: 2390x1216
**[Open #180 in Usersnap](https://app.usersnap.com/l/feedback/5f831e02-9b48-4051-891a-423b1cc650fa?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
Duplicate of
- #2294 

Closing

thesan:
I'm repurposing this issue based on the 3rd field suggestion

traumschule:
This may conflict with #2949 and #1794 - can be hidden via CSS.

PR #3060, opened 2022-05-13
2294 Sorting of proposals

oleksanderkorn:
https://github.com/Joystream/pioneer/issues/2294

### DIFF
--- a/packages/ui/src/app/pages/Proposals/PastProposals.tsx
+++ b/packages/ui/src/app/pages/Proposals/PastProposals.tsx
@@ -24,7 +24,7 @@ export const PastProposals = () => {
 
   const { types, stages } = usePastProposals()
 
-  const { order, getSortProps } = useSort<ProposalOrderByInput>('createdAt')
+  const { order, getSortProps } = useSort<ProposalOrderByInput>('statusSetAtTime')
 
   const { isLoading, proposals, pagination } = useProposals({ order: order, status: 'past', filters })
 
--- a/packages/ui/src/app/pages/Proposals/PastProposals.tsx
+++ b/packages/ui/src/app/pages/Proposals/PastProposals.tsx
@@ -1,12 +1,14 @@
 import React, { useRef, useState } from 'react'
 
 import { PageLayout } from '@/app/components/PageLayout'
+import { ProposalOrderByInput } from '@/common/api/queries'
 import { ActivitiesBlock } from '@/common/components/Activities/ActivitiesBlock'
 import { FilterPageHeader } from '@/common/components/forms/FilterBox'
 import { MainPanel } from '@/common/components/page/PageContent'
 import { SearchProcess } from '@/common/components/page/SearchProcess'
 import { SidePanel } from '@/common/components/page/SidePanel'
 import { Pagination } from '@/common/components/Pagination'
+import { useSort } from '@/common/hooks/useSort'
 import { AddProposalButton } from '@/proposals/components/AddProposalButton'
 import { ProposalEmptyFilter, ProposalFilters } from '@/proposals/components/ProposalFilters'
 import { ProposalList } from '@/proposals/components/ProposalList'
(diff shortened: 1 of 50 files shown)


oleksanderkorn:
@thesan please review
Instead of adding the select switch I made the title column sortable by proposal title and added ended at separate column which is also sortable by title.  Let me know what you think.

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
```diff
@@ -24,9 +28,10 @@ export const ProposalList = ({ proposals, isPast }: ProposalListProps) => {
   return (
     <RowGapBlock gap={4}>
       <ProposalsListHeaders $colLayout={ProposalColLayout}>
-        <ProposalListHeader />
+        {getSortProps ? <SortHeader {...getSortProps('title')}>Title</SortHeader> : <ProposalListHeader />}
         <ProposalListHeader>Stage</ProposalListHeader>
         <ProposalListHeader>Proposer</ProposalListHeader>
+        {isPast && getSortProps ? <SortHeader {...getSortProps('createdAt')}>Ended</SortHeader> : null}
```
thesan, line 34:
```suggestion
        {isPast && getSortProps ? <SortHeader {...getSortProps('statusSetAtTime')}>Ended</SortHeader> : null}
```

thesan (changes requested):
Really nice PR again :+1: 
It works well technically. But in the past proposals it makes the design redundant.
IMO the current proposal is great but in the past proposals the `Title` column header should be replaced by the order selector suggested in [#2294](https://github.com/Joystream/pioneer/issues/2294#issue-1146970078). Something like this:
Screenshot from 2022-05-20 11-22-01
But also with: `By Name A-Z` and `By Name Z-A`. WDYT ?

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
oleksanderkorn: Thanks for the feedback, for me it was not very nice to have the dropdown as a column header, not very clean for me, also because we do not have any other places with a similar design. The dropdown we have in bounties but it is not a column header, but a separate UI element. If we can find a proper place to show the dropdown in the past proposals it would be great. In the meanwhile, i wil try the suggested approach and maybe will come up with something else :)

packages/ui/src/app/pages/Proposals/PastProposals.tsx
```diff
@@ -21,7 +23,10 @@ export const PastProposals = () => {
   const [filters, setFilters] = useState(ProposalEmptyFilter)
 
   const { types, stages } = usePastProposals()
-  const { isLoading, proposals, pagination } = useProposals({ status: 'past', filters })
+
+  const { order, getSortProps } = useSort<ProposalOrderByInput>('createdAt')
```
thesan, line 27: Oh! I forgot about this one also:
```suggestion
  const { order, getSortProps } = useSort<ProposalOrderByInput>('statusSetAtTime')
```

packages/ui/src/proposals/components/ProposalList/ProposalList.tsx
thesan: @oleksanderkorn yes I agree it's not an ideal design, I just have a very slight preference for the dropdown over the duplicated end date because I find it nicer on the user to not have redundant information thrown at them, and it leaves some space for the eventual [Preview rationale in proposal lists #2771](https://github.com/Joystream/pioneer/issues/2771). That said I guess removing the "Ended at dd/mm/yy" from the title and just showing the date (without the "Ended at") in the `ENDED` column could work too

packages/ui/src/app/pages/Proposals/PastProposals.tsx
oleksanderkorn: @thesan pushed the requested changes.

thesan (changes requested):
@oleksanderkorn it looks better but please also remove the "Ended at dd/mm/yy" from the title column, and change the `grid-template-columns` from `296px 148px 1fr 1fr` to `296px 148px 156px 1fr`

So it looks like this:
Screenshot from 2022-05-24 17-26-33

oleksanderkorn:
> @oleksanderkorn it looks better but please also remove the "Ended at dd/mm/yy" from the title column, and change the `grid-template-columns` from `296px 148px 1fr 1fr` to `296px 148px 156px 1fr`
> 
> So it looks like this: Screenshot from 2022-05-24 17-26-33

Done :)

Issue #2413, opened 2022-03-03
Bounty Tabs content: Works/ Winners/ Slashed


dmtrjsg:
Displaying Winners and Slashed Workers for bounties:

- [x]  content of winners tab needs to be re-done based on designs [FIGMA>](https://www.figma.com/file/jZwIoaxOkTxNFMJNzssBgv/Bounty?node-id=319%3A145491)
- [x] Second tab needs to be "Slashed" and display slashed entrants and their slashed amount
[Link to Figma ](https://www.figma.com/file/jZwIoaxOkTxNFMJNzssBgv/Bounty?node-id=328%3A157317)
- [x] Show contents of work Tab  [FIGMA> ](https://www.figma.com/file/jZwIoaxOkTxNFMJNzssBgv/Bounty?node-id=289%3A192155)
- [ ] ⚠️ re-test if Terminated bounties display winners/ losers/ slashed correctly.

![Screenshot 2022-03-03 at 19.58.01.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/3ea4af1e-3766-4a70-85cd-122a00a59490)

polikosi:
Winners page displayed right:
![win](https://user-images.githubusercontent.com/80429191/168305370-b6c709e8-7860-41b6-b66b-2e2796760e00.png)

Works Tab displayed right:
![works tab](https://user-images.githubusercontent.com/80429191/168305528-9d398b13-cece-461c-98fe-099e275c3823.png)

Slashed page not right:
1. Slashed amount isn't displayed 
2. Rationale isn't displayed
3. In the sidebar does not display what were slashed
![slashed](https://user-images.githubusercontent.com/80429191/168307201-d86378de-6849-4b67-b804-2512eef36ccc.png)

dmtrjsg:
1. Rationale

Agreed to show rationale re-using this design pattern:

<img width="1138" alt="Screenshot 2022-05-17 at 09 16 05" src="https://user-images.githubusercontent.com/93924525/168763966-80d01c0e-bfdc-48d0-b9fe-7b3a4ff15c16.png">

NB: Rationale is a single text input that applies to all entries for bounty judgement.

2. Slashed amount (this is always = full stake of applicant) we will try and display it
3. Description will not be provided for each candidate/ not part of the runtime design

dmtrjsg:
@WRadoslaw 

Copy:

"
You won the bounty!
[...]

PR #3080, opened 2022-05-17
🦝 Fix on bounty judgement rationale

WRadoslaw:
#2413

### DIFF
--- a/packages/ui/src/bounty/components/tabs/SlashedTab.tsx
+++ b/packages/ui/src/bounty/components/tabs/SlashedTab.tsx
@@ -5,7 +5,6 @@ import { Bounty } from '@/bounty/types/Bounty'
 import { List } from '@/common/components/List'
 import { RowGapBlock } from '@/common/components/page/PageContent'
 import { NotFoundText } from '@/common/components/typography/NotFoundText'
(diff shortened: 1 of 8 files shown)


packages/ui/src/bounty/components/tabs/SlashedTab.tsx
```diff
@@ -20,7 +20,7 @@ export const SlashedTab = ({ bounty }: Props) => {
           {slashedWorkers
             .filter((entry) => entry.rejected)
             .map((entry) => (
-              <BountySlashedListItem entrant={entry.worker} inBlock={randomBlock()} link="xd" />
+              <BountySlashedListItem entrant={entry.worker} stake={entry.stake} inBlock={randomBlock()} link="" />
```
thesan, line 23: The PR looks good but is there a way to replace this `randomBlock()` by the actual block of the judgement ?

stepien4soft (approved):

packages/ui/src/bounty/components/tabs/SlashedTab.tsx
WRadoslaw: Can do. One more thing about list item, should we remove link since winner list item doesn't have it either?

Issue #3054, opened 2022-05-13
CreateThread


stepien4soft:
Refactor the `CreateThread` modal as part of the #2802 epic.

## Dev scope
- [x] Refactor fields validation
- [x] Manage the form state with React Hook Form instead of XState

## QA scope
- [ ] The fields are validated correctly
- [ ] The transaction gets sent with the parameter set in the modal

ivanturlakov:
Tested on staging.

The fields are validated correctly but allow special characters.
The transaction gets sent with the parameter set in the modal.
<img width="1268" alt="Снимок экрана 2022-06-03 в 16 48 59" src="https://user-images.githubusercontent.com/46903215/171868381-119d397c-e9ab-4ce8-b5ad-9729bc430dc9.png">
<img width="1207" alt="Снимок экрана 2022-06-03 в 16 55 09" src="https://user-images.githubusercontent.com/46903215/171868474-48157e1d-be59-402d-82b9-4408ef69a16c.png">

dmtrjsg:
@ivanturlakov we concluded its acceptable for the forum threads.

PR #3108, opened 2022-05-25
✉️ Bugfix/create thread form refactor

stepien4soft:
#3054


packages/ui/src/forum/modals/CreateThreadModal/machine.ts
```diff
@@ -41,8 +41,6 @@ export type CreateThreadEvent =
   | { type: 'PASS'; memberId: string; categoryId: string; controllerAccount: Account }
   | { type: 'NEXT' }
   | { type: 'BACK' }
-  | { type: 'SET_TOPIC'; topic: string }
```
WRadoslaw, line 44: Would be good to delete `topic` and `description` from context in this file keep typing correct on machine usage

WRadoslaw (approved):

Issue #2415, opened 2022-03-03
Top Contributors - shading


dmtrjsg:
Expected result (each card has shading disctinct borders):

![Screenshot 2022-03-03 at 20.10.55.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/cec2b6de-3966-49c6-8843-af31d12f56d0)

Actual result:

![Screenshot 2022-03-03 at 20.10.49.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/bf4ed9f5-230a-4307-ade6-1f0e687ac569)

freakstatic:
Bounty link: https://dao.joystream.org/#/bounty/preview/45

singulart:
@dmtrjsg Please clarify what does it mean `each card has shading disctinct borders`?

ivanturlakov:
Tested on staging.

Cards shadow fixed
<img width="1206" alt="Снимок экрана 2022-06-03 в 18 55 29" src="https://user-images.githubusercontent.com/46903215/171902145-b593b422-eb12-48d5-bc89-87ba8f0c94b3.png">

PR #3112, opened 2022-05-26
Top shading isue fix

cmgoes:
Fixed for top shading issue on bounty #2415
Expected result (each card has shading disctinct borders):


thesan (approved):
Looks good. Thanks you the PR :pray:

Issue #2645, opened 2022-03-24
Need to choose membership after reloading the page


dmtrjsg:
- [x]  Add sign-out  [Link to Figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9470%3A339901)
- [x] Save current member selected in local storage [DEV]

⚠️ "fake" the membership pre-selected while chain is loading

___

**Taken at**: https://dao.joystream.org/#/profile
**Fields**:
- Houston, we have a problem.: need to choose membership after reloading the page
- Give us some details: when i reload the page, i need to choose my membership again. it would be better to make it constant because i reload page many times, it wastes my time:(
- i don&#39;t know
- Discrod Handle: swasilenko#9850

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/2a6ae3c1-be5f-4a78-809f-e9ee81b0f799-annotated_4bbcc3c5-bcf8-45f2-abab-62492aee1edf.png?etag=b9462ec8a1fe3789a2eb19a05d8cdb64)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/2a6ae3c1-be5f-4a78-809f-e9ee81b0f799-annotated_4bbcc3c5-bcf8-45f2-abab-62492aee1edf.png?etag=b9462ec8a1fe3789a2eb19a05d8cdb64)*

**Browser**: Firefox 97 (macOS Catalina)
**Screen size**: 1792x1120
**Browser size**: 1668x982
**Labels**: bug-non-critical
**[Open #124 in Usersnap](https://app.usersnap.com/l/feedback/73adc781-b11d-4d01-8d77-36524ccee8c4?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
@thesan here's the issue, it was in the design sprint column, moving to In-design with a view to turn it around in a few hours. @MattAugustynowicz this is a blocker for a related issue > https://github.com/Joystream/pioneer/pull/3113 which has an open PR already.

dmtrjsg:
@stepien4soft pls add it to https://github.com/Joystream/pioneer/pull/3113

PR #3113, opened 2022-05-26
🩹 Improving UX at page refresh

WRadoslaw:
This PR is proposition to preselect recent active membership at page load. This should improve user experience and slightly decrease fatigue caused by need of refreshing to obtain latest data. This should  be treated as Band-Aid more than any fix.


thesan (changes requested):
I can't find the issue related to this, but I remember it has 2 parts:
1. Stay logged in with the last selected member ✔️
2. Add a logout button ❌

Without the button you can't "logout" without clearing the local storage manually, I don't remember if there is a design for it.

Help @dmtrjsg !

dmtrjsg:
This issue is awaiting treatment for the logout flow > https://github.com/Joystream/pioneer/issues/2645

stepien4soft:
@thesan logout flow was added. Ready for review again 👌

packages/ui/src/memberships/components/SignOutButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { SignOutIcon } from '@/common/components/page/Sidebar/LinksIcons/SignOutIcon'
+
+import { TextSmall } from '../../common/components/typography'
+import { BorderRad, Colors, Transitions } from '../../common/constants'
+
+interface Props {
+  onClick: () => void
+}
+
+export const SignOutButton = ({ onClick }: Props) => {
```
thesan, line 13: Instead of adding a `SignOutButton` with the same style as `AddMembershipButtonSwitch` could you add something like `@/common/components/buttons/DetailsButtons` and a story for it, which would be used like this:
```tsx
<DetailsButtons title="New Member" icon={<MyProfileIcon />} disabled={isTransactionPending} onClick={() => { ... }}>
  Create a New Membership
</DetailsButtons>
{!!active && (
  <DetailsButtons title="Sign Out" icon={<SignOutIcon />} onClick={() => { ... }}>
    Sign out of the active Membership
  </DetailsButtons>
)}
```

packages/ui/src/memberships/components/components.tsx
```diff
@@ -58,6 +59,8 @@ export const MemberPhoto = styled.div<MemberPhotoProps>`
   align-items: center;
   max-width: ${({ small }) => (small ? '24px' : '40px')};
   max-height: ${({ small }) => (small ? '24px' : '40px')};
+  max-width: ${({ big }) => (big ? '80px' : '40px')};
+  max-height: ${({ big }) => (big ? '80px' : '40px')};
```
thesan, line 63:
```suggestion
  ${({ big, small }) => {
    if (big) {
      return css`
        max-width: 80px;
        max-height: 80px;
      `
    } else if (small) {
      return css`
        max-width: 24px;
        max-height: 24px;
      `
    } else {
      return css`
        max-width: 40px;
        max-height: 40px;
      `
    }
  }}
```
This is a lot longer but easier too read IMO

thesan (changes requested):
@stepien4soft sorry for being so late on this review :bow:
It looks good overall, but I have a few suggestions

packages/ui/src/common/components/buttons/DetailsButton.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+  title: 'Common/Buttons/DetailsButton',
+  component: DetailsButton,
+} as Meta
+
+const Template: Story<DetailsButtonProps> = (args) => (
+  <TemplateBlock>
+    <Row>
+      <DetailsButton {...args} />
```
thesan, line 17: To make it the story a bit clearer:
```suggestion
    <BlackBlock>
      <DetailsButton {...args} />
    </BlackBlock>
```
from `../storybookParts/previewStyles`

packages/ui/src/memberships/modals/SwitchMemberModal/SwitchMemberModal.tsx
```diff
@@ -53,12 +56,26 @@ export const SwitchMemberModal = () => {
...
           />
+          {active ? (
+            <DetailsButton
+              icon={<SignOutIcon />}
+              onClick={() => {
+                hideModal()
+                showModal<SignOutModalCall>({ modal: 'SignOut' })
+              }}
+              subtitleText="Sign Out"
+              titleText="Sign out of the active Membership"
```
thesan, line 76: Looks good but `subtitleText` and `titleText` are inverted:
```suggestion
            titleTextNew Member"
            subtitleText="="Create a New Membership"
          />
          {active ? (
            <DetailsButton
              icon={<SignOutIcon />}
              onClick={() => {
                hideModal()
                showModal<SignOutModalCall>({ modal: 'SignOut' })
              }}
              titleText="Sign Out"
              subtitleText="="Sign out of the active Membership"
```

thesan (changes requested):

thesan (approved):
:+1:

PR #3126, opened 2022-05-30
Add proposal ID to proposal card (#2296)

### LINKED ISSUE #2296: [USERSNAP]Add proposal ID to proposal card
## Scope

- [ ] Add proposal ID to the proposal card (list)
- [ ] Make proposal link easy to copy 
- [ ] Dev comment: When user pastes such iD (copied to buffer) to text box (like forum post) the markdown of proposal is identified and shown correctly
__________

**Taken at**: https://pioneer-2.vercel.app/#/proposals/current
**Fields**:
- Discrod Handle: tomato#9464
- Give us some details: the number is an important unique identifier
- Houston, we have a problem.: Current Proposal view should include proposal number
- N/A

SCREENSHOT
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/0392b223-d755-452a-a28f-4f9b10f778ee-annotated_c1176b64-0d17-4e3c-b8be-1153866f4799.png?etag=9ec5fa7c8e98235182741d07bb14bed7)*

**Browser**: Chrome 98 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 2022x1319
**[Open #11 in Usersnap](https://app.usersnap.com/l/feedback/06cb4dae-0490-4fc9-9dc3-36e8e7bc08f1?utm_medium=integration&utm_campaign=integration&utm_content=open)**

freakstatic:
Fixes #2296

### DIFF
--- a/packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
+++ b/packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
@@ -3,6 +3,8 @@ import { generatePath } from 'react-router-dom'
 import styled from 'styled-components'
 
 import { BadgeStatus } from '@/common/components/BadgeStatus'
+import { CopyButtonTemplate } from '@/common/components/buttons'
+import { LinkIcon } from '@/common/components/icons'
 import { TableListItem } from '@/common/components/List'
 import { GhostRouterLink } from '@/common/components/RouterLink'
 import { Tooltip, TooltipDefault } from '@/common/components/Tooltip'
@@ -11,6 +13,7 @@ import { TextSmall } from '@/common/components/typography/Text'
 import { Colors, Overflow } from '@/common/constants'
 import { camelCaseToText } from '@/common/helpers'
 import { toDDMMYY } from '@/common/utils/dates'
+import { getUrl } from '@/common/utils/getUrl'
 import { MemberInfo } from '@/memberships/components'
 import { ProposalColLayout } from '@/proposals/constants'
 import { ProposalsRoutes } from '@/proposals/constants/routes'
@@ -53,6 +56,7 @@ export const ProposalListItem = ({ proposal, isPast, memberId, isCouncilMember }
         return 'Proposal was rejected by council. Rationale can be checked in proposal details, and nothing further can happen.'
     }
   }
+
   return (
     <ProposalItem
       as={GhostRouterLink}
@@ -62,6 +66,7 @@ export const ProposalListItem = ({ proposal, isPast, memberId, isCouncilMember }
     >
       <ToggleableItemInfo>
         <ToggleableItemInfoTop>
+          <Subscription>{`ID: ${proposal.id}`}</Subscription>
           {!isPast && <Subscription>Created at: {toDDMMYY(displayDate)}</Subscription>}
           <BadgeStatus>{camelCaseToText(proposal.type)}</BadgeStatus>
         </ToggleableItemInfoTop>
@@ -74,17 +79,28 @@ export const ProposalListItem = ({ proposal, isPast, memberId, isCouncilMember }
         </Tooltip>
       </StageField>
       <MemberInfo member={proposal.proposer} memberSize="s" showIdOrText />
-      {isPast && (
-        <StageField>
-          <Subscription>{toDDMMYY(displayDate)}</Subscription>
-        </StageField>
-      )}
+      <StageField>{isPast && <Subscription>{toDDMMYY(displayDate)}</Subscription>}</StageField>
       <StageField>
         <ProposalItemVoteDetails proposal={proposal} memberId={memberId} isCouncilMember={isCouncilMember} />
       </StageField>
+      <CopyButton
+        square
+        size="small"
+        textToCopy={getUrl({ route: ProposalsRoutes.preview, params: { id: proposal.id } })}
+        icon={<LinkIcon />}
+      />
     </ProposalItem>
   )
 }
+
+const CopyButton = styled(CopyButtonTemplate)`
+  justify-self: end;
+  visibility: hidden;
+  transition: none;
+  & svg {
+    transition: none !important;
+  }
+`
 
 const ProposalItem = styled(TableListItem)`
   grid-column-gap: 36px;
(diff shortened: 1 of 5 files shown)


packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
```diff
@@ -62,6 +64,10 @@ export const ProposalListItem = ({ proposal, isPast, memberId, isCouncilMember }
     >
       <ToggleableItemInfo>
         <ToggleableItemInfoTop>
+          <ProposalUrlCopyButton
+            altText={`ID: ${proposal.id}`}
+            copyText={getUrl({ route: ProposalsRoutes.preview, params: { id: proposal.id } })}
+          />
```
thesan, line 70: Thank you for the PR. It looks good and works well. But since @MattAugustynowicz provided some designs please use a button with a link icon at the end of the proposal row, which shows on hover only, instead of the `ProposalUrlCopyButton`, see https://github.com/Joystream/pioneer/issues/2296#issuecomment-1075016513:

thesan (changes requested):

packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
freakstatic: Oh didn't saw the button on the designs :facepalm: 
fixed, please check it now!

freakstatic: Oh wait...

thesan (changes requested):
@freakstatic it's not too clear from the design (given the small size of the screen on figma), but it looks to me like the copy buttons are align to the end of the row, (here it would be: `justify-self: end;`). Currently on a bigger screen it looks like this:
Screenshot from 2022-06-02 16-12-25

Also the button is supposed to show on hover only

freakstatic:
> @freakstatic it's not too clear from the design (given the small size of the screen on figma), but it looks to me like the copy buttons are align to the end of the row, (here it would be: `justify-self: end;`). Currently on a bigger screen it looks like this: Screenshot from 2022-06-02 16-12-25
> 
> Also the button is supposed to show on hover only

I have applied the `justify-self: end;` (thanks for the suggestion btw!)
yeah I notice that I forgot about the hover only, should be good now

packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
```diff
@@ -74,18 +82,28 @@ export const ProposalListItem = ({ proposal, isPast, memberId, isCouncilMember }
...
+        />
+      )}
     </ProposalItem>
   )
 }
 
+const CopyButton = styled(CopyButtonTemplate)`
+  ${css`
+    justify-self: end;
+  `}
```
thesan, line 104: Here there's no need for the `css` tag, you can write css directly in the `styled(CopyButtonTemplate)` tag.

More importantly I would really prefer if the hover logic was handled in css:
```suggestion
  justify-self: end;
  visibility: hidden;
```

And inside the `ProposalItem` declaration add:
```css
&:hover ${CopyButton} {
  visibility: visible;
}
```
(in case there is a delay with hover a `transition: none` on the `CopyButton` would take care of it)
This would really reduce the code complexity for this component.

packages/ui/src/proposals/constants/styles.ts
```diff
@@ -2,7 +2,7 @@ import styled from 'styled-components'
 
 import { ListHeaders, ListHeader } from '@/common/components/List/ListHeader'
 
-export const ProposalColLayout = '296px 148px 156px 1fr'
+export const ProposalColLayout = '296px 148px 156px 60px 92px 35px'
```
thesan, line 5: Nice catch on the voting buttons, I had completely forgot about these, but please keep the last column to `1fr` to match the design alignment on bigger screens:
```suggestion
export const ProposalColLayout = '296px 148px 156px 60px 92px 1fr'
```

thesan (changes requested):

freakstatic:
@thesan Thanks very much for patient and input to improve this! 
Please check again now :smile:

packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
```diff
@@ -74,24 +79,38 @@ export const ProposalListItem = ({ proposal, isPast, memberId, isCouncilMember }
...
     </ProposalItem>
   )
 }
 
+const CopyButton = styled(CopyButtonTemplate)`
+  justify-self: end;
+  visibility: hidden;
+  transition: none;
+  & svg {
+    transition: none !important;
```
freakstatic, line 101: @thesan any idea to avoid this `!important`?

thesan: @freakstatic these `transition: all ...` are becoming annoying and I don't even know which properties they should target, so rewriting them properly will be difficult :unamused: .

Thankfully in this case since the button is at the end of the row, replacing `visibility: hidden`/`visibility: visible` by `display: none`/`display: flex` should work without overriding the transitions

thesan:
> @thesan Thanks very much for patient and input to improve this! Please check again now smile

No problem, thank you for your reactivity. Dealing with UI usually requires a lot more back and forth than dealing with pure logic.

packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
freakstatic: yeah that works but the table does a strange effect when dragging

freakstatic: image
image

freakstatic: not sure you can notice in the two images, but the table grows a bit to the right side on mouse hover

thesan (approved):
Sorry for the delay I missed you're reply this morning. I can't really see the problem on the screenshot but I get the idea. In any case I think the `!important` is not too problematic here since it's scoped to the local `CopyButton`

Issue #2220, opened 2022-02-11
Expand drop downs on full panel click 


dmtrjsg:
Currently user needs to click to teh downward pointing chevron to expand Member application in the sliding panel

![Screenshot 2022-02-11 at 17.17.15.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/915cc820-964d-4b0b-90a7-c348d0caba43)

Same applies for  list cards to details page:
- [ ] Openings
- [ ] Proposals
- [ ] Bounties

ytdbuolf:
Will accept this for fixing

ytdbuolf:
Issue has been resolved

ivanturlakov:
Tested on staging

The panel still opens when you click on the chevron

Openings

https://user-images.githubusercontent.com/46903215/174994876-8c19faa9-c0c4-43e8-ade6-51d61d6dd3f4.mov


Proposals

https://user-images.githubusercontent.com/46903215/174994914-50e3ef83-57e8-427a-8974-aff228dd5935.mov


Bounty

https://user-images.githubusercontent.com/46903215/174994921-1b133ae4-25a7-4a33-ba96-cd8ae8c34651.mov

oleksanderkorn:
Hey maybe it was not deployed to staging? 
Because on my branch pr deployment it works fine:
https://dao-git-fork-oleksanderkorn-bugfix-2220-joystream.vercel.app/#/proposals/preview/310

ivanturlakov:
@oleksanderkorn yes, indeed

Retested on https://pioneer-2.vercel.app/

Dropdown expand on full panel click

https://user-images.githubusercontent.com/46903215/175259986-5189a52d-463a-48c6-9137-5f2661fe8e17.mov


https://user-images.githubusercontent.com/46903215/175260008-e221317c-ad4f-4a58-b80c-f6aef18a2a03.mov


https://user-images.githubusercontent.com/46903215/175260023-f5c880b7-b771-43f3-bffe-0ca5bce8559d.mov

PR #3127, opened 2022-05-30
Expand drop down issue fix

ytdbuolf:
"Currently user needs to click to teh downward pointing chevron to expand Member application in the sliding panel"
This has been resolved

### DIFF
--- a/packages/ui/src/common/components/buttons/Toggle.tsx
+++ b/packages/ui/src/common/components/buttons/Toggle.tsx
@@ -2,7 +2,7 @@ import React, { ReactNode } from 'react'
 import styled, { css } from 'styled-components'
 
 import { BorderRad, Colors, Transitions } from '../../constants'
-import { useToggle } from '../../hooks/useToggle'
+// import { useToggle } from '../../hooks/useToggle'
 import { isFunction } from '../../utils'
 import { Arrow, Icon } from '../icons'
 
@@ -12,10 +12,12 @@ interface ToggleableItemProps {
   children: ReactNode | ((isOpen: boolean) => ReactNode)
   absoluteToggle?: boolean
   className?: string
+  isOpen: boolean
+  toggleOpen: () => void
 }
 
-export const ToggleableItem = ({ children, absoluteToggle, className }: ToggleableItemProps) => {
-  const [isOpen, toggleOpen] = useToggle()
+export const ToggleableItem = ({ children, absoluteToggle, className, isOpen, toggleOpen }: ToggleableItemProps) => {
+  // const [isOpen, toggleOpen] = useToggle()
 
   return (
     <Toggle isOpen={isOpen} absoluteToggle={absoluteToggle} className={className}>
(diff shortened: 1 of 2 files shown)


singulart:
Fixes https://github.com/Joystream/pioneer/issues/2220

ytdbuolf:
Fixed issue-#2220

packages/ui/src/common/components/buttons/Toggle.tsx
```diff
@@ -12,10 +12,12 @@ interface ToggleableItemProps {
...
+  toggleOpen: () => void
 }
 
-export const ToggleableItem = ({ children, absoluteToggle, className }: ToggleableItemProps) => {
-  const [isOpen, toggleOpen] = useToggle()
+export const ToggleableItem = ({ children, absoluteToggle, className, isOpen, toggleOpen }: ToggleableItemProps) => {
+  // const [isOpen, toggleOpen] = useToggle()
 
   return (
     <Toggle isOpen={isOpen} absoluteToggle={absoluteToggle} className={className}>
```
thesan, line 23: Thanks for the PR. It works but it introduces some necessary changes. While this:
```suggestion
    <Toggle isOpen={isOpen} absoluteToggle={absoluteToggle} className={className} onClick={toggleOpen}>
```
And removing the `onClick` from `ToggleButton` should work too without changing the component behavior.

Also please try to not leave commented out code in the PRs

thesan (changes requested):

packages/ui/src/common/components/buttons/Toggle.tsx
ytdbuolf: Please check again, it works without changing the component behavior
Removed all comments that had left there before

thesan: By "component behavior" I meant the behavior of `ToggleableItem`, in the current code base `ToggleableItem` manages the state of the toggle (which IMO makes it a bit nicer to use), while in this PR the parent component keeps the state. This leads to the **unnecessary* changes I was talking about (sorry about the typo).

It's not too bad because it seems `ToggleableItem` is currently only used in `MemberRoleToggle`, but please try to keep the changes minimum for smaller this PRs like this one.

singulart:
@thesan Please see the developer comment.

packages/ui/src/common/components/buttons/Toggle.tsx
thesan: @ytdbuolf I'm sorry for the very late reply.

The icon you are referring to is the arrow, right ? In this case, yes both the panel and icon should be clickable.

But the way I see it, since the icon is within the `onClick` should be on the panel (i.e `<Toggle />`) and not on the icon (i.e `<ToggleButton><Arrow/><ToggleButton>`). Is this what are proposing in your latest comment ? If so, is there a drawback to this solution that I am missing ?

thesan:
Done in #2220

PR #3129, opened 2022-05-30
Strip backslashes in MarkdownPreview to make it work

oleksanderkorn:
https://github.com/Joystream/pioneer/issues/2712
image

### DIFF
--- a/packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
+++ b/packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
@@ -62,12 +62,12 @@ export const MarkdownPreview = ({ markdown, append, ...styleProps }: MarkdownPre
     }
   }, [markdown, append])
 
-  const stripSlashes = (text: string) => text.replace(/\\(.)/gm, '$1')
+  const stripBackslashes = (text: string) => text.replace(/\\(.)/gm, '$1')
   return (
     <div className="markdown-preview">
       <MarkdownPreviewStyles {...styleProps} />
-      <ReactMarkdown rehypePlugins={rehypePlugins} components={components} rawSourcePos skipHtml unwrapDisallowed>
-        {stripSlashes(markdown)}
+      <ReactMarkdown rehypePlugins={rehypePlugins} components={components} rawSourcePos>
+        {stripBackslashes(markdown)}
       </ReactMarkdown>
       {appendAfter && <p>{append}</p>}
     </div>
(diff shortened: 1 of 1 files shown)


oleksanderkorn:
@thesan please review :)

oleksanderkorn:
@thesan please review :)

packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
```diff
@@ -62,11 +62,12 @@ export const MarkdownPreview = ({ markdown, append, ...styleProps }: MarkdownPre
...
 
+  const stripSlashes = (text: string) => text.replace(/\\(.)/gm, '$1')
   return (
     <div className="markdown-preview">
       <MarkdownPreviewStyles {...styleProps} />
-      <ReactMarkdown rehypePlugins={rehypePlugins} components={components} rawSourcePos>
-        {markdown}
+      <ReactMarkdown rehypePlugins={rehypePlugins} components={components} rawSourcePos skipHtml unwrapDisallowed>
+        {stripSlashes(markdown)}
       </ReactMarkdown>
```
traumschule, line 71: nice work!
wonder if tables in https://pioneer-2-git-fork-oleksanderkorn-bugfix-2712-joystream.vercel.app/#/forum/thread/57 work with [gfm](https://blog.logrocket.com/how-to-safely-render-markdown-using-react-markdown/), also some of the `## headings`
```
import gfm from 'remark-gfm'
<ReactMarkdown remarkPlugins={[gfm]}
```

oleksanderkorn: Seems to be not working locally, also tried here with enabled gfm, but no luck: https://remarkjs.github.io/react-markdown/
image

thesan (changes requested):
Really nice catch :raised_hands: 

But it would be great to find where is the escaping coming from (in Pioneer, in the mapping, or in the query node). This way it could maybe be fixed at the source. If not we should at least make sure it happens everywhere Pioneer is displaying markdown.

Also even as a temporary solution we should not escape all `\`, this regex should fix this:
```js
/(^|[^\\])\\([^\\])/g
```

thesan (approved):
Thanks for checking where the escaping happens. In this case fixing this in `MarkdownPreview` does look like the best solution.

About the regex, you're right, `/\\(.)/` works perfectly, I'm really sorry I did test this regex just not very well :facepalm: . Just about the flags I think you meant to use [s](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll#description) instead of [m](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/multiline#description) but it doesn't really affect this use case.

Once again great PR :raised_hands:

Issue #3051, opened 2022-05-13
InviteMember form refactor


stepien4soft:
Refactor the `InviteMember` modal as part of the #2802 epic.

## Dev scope
- [x] Refactor fields validation
- [x] Manage the form state with React Hook Form instead of XState

## QA scope
- [ ] The fields are validated correctly
- [ ] The transaction gets sent with the parameter set in the modal

ivanturlakov:
In this case, I entered the already taken handle that was successfully validated and I was able to initiate the transaction
<img width="1434" alt="Снимок экрана 2022-05-26 в 12 32 41" src="https://user-images.githubusercontent.com/46903215/170462285-7d86e7e5-a230-4d34-8c33-980242498836.png">

ivanturlakov:
Root account field validation.

https://user-images.githubusercontent.com/46903215/170666415-0af4b8fa-89c2-4472-b615-63e67e9e4a7e.mov

dmtrjsg:
Root account validation - check for regex
Membership Handle - check for duplicates from QN

ivanturlakov:
Tested on staging.

Root account validation fixed
Membership Handle validation fixed

https://user-images.githubusercontent.com/46903215/171623096-42894a5c-fa0e-4fcb-8aad-8e5f8d14b180.mov

PR #3145, opened 2022-06-01
🦚 Fix on invite member modal 

WRadoslaw:
#3051


thesan (approved):
Really good catch ! The `formHandleMap` was not even debounced :scream:. It might have caused some serious perturbation with rpc connections.

Issue #2431, opened 2022-03-04
[USERSNAP] Forum: Top threads button doesn't work


dmtrjsg:
**Taken at**: https://pioneer-2.vercel.app/#/forum/latest-threads
**Fields**:
- Houston, we have a problem.: Forum: Top threads button doesn&#39;t work
- Give us some details: Forum: Top threads button doesn&#39;t work
- Forum: Top threads button doesn&#39;t work
- Discrod Handle: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/92b107e7-14d7-4c11-bdee-8c320f2103e1-annotated_97e95d7b-e6f4-4a32-a7d0-ecf198136a3c.png?etag=bc143ed16fb1e42876133d3a11766d3b)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/92b107e7-14d7-4c11-bdee-8c320f2103e1-annotated_97e95d7b-e6f4-4a32-a7d0-ecf198136a3c.png?etag=bc143ed16fb1e42876133d3a11766d3b)*

**Browser**: Chrome 98 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1718x1306
**[Open #59 in Usersnap](https://app.usersnap.com/l/feedback/618c7305-fef6-4405-a695-6d484b754c8f?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

talentdev0807:
@talentdev0807

secretgir1:
Hey guys @mochet @dmtrjsg @bwhm Description here is unclear.

Let's discuss requirements regarding this ticket.
What is you expectation? 

As I see button is clickable, but here is another problem. Page appears the same as a Categories Tab.
We should understand what user have to see on this page. Do we have UI for this page?

As Forum lead I expect to see list of the most popular threads. For example list of the top 10-15 threads. 

Any ideas on that? What do you expect to see on this page?

talentdev0807:
- We don't have page, router and query for this
- You are right, we can get top 10-15 threads
I thinks 5 threads is good

dmtrjsg:
[...]

PR #3173, opened 2022-06-03
Fix/Top threads tab is not clickable on forum

talentdev0807:
#2431 
- Add new TopThreads Component
-  Add new query to fetch the top threads
-  Add new topThreads router


thesan (approved):
Great PR ! Thanks !

Issue #3159, opened 2022-06-02
Map QN `BigInt` into `string`


thesan:
## Context
Currently the QN transmit `BigInt` fields as `string` code generated from the schema maps them to `any`. As a result typescript fails to detect some typing errors, which leads to issues like #3156.

## Scope
- [x] Map `BigInt` to `string` in the [`codegen.config.yml`](https://github.com/Joystream/pioneer/blob/d7bddcb98d79fc7a61c24b1a44a266d794541de5/packages/ui/codegen.config.yml)
- [x] Fix potential remaining type errors

PR #3205, opened 2022-06-08
🪛 Bugfix/big int into string

stepien4soft:
#3159


thesan (changes requested):
Nice job!
Thanks for not converting anything to numbers, now with `1 JOY` being a 10 digits integer: `900 720 JOY` is already above the `Number.MAX_SAFE_INTEGER` (which is lower than the runtime upgrade proposal stake for example), so we have to be extra careful.

thesan (approved):
LGTM!

Issue #3143, opened 2022-05-31
Stake not marked as recovered (awaiting Runtime upgrade)


dmtrjsg:
**Taken at**: https://dao.joystream.org/#/election/past-votes
**Fields**:
- Houston, we have a problem.: Stake not marked as recovered
- Comment: My steak is not marked as restored. I have already restored it and used the funds again. Sometimes I work from two devices.

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/c7ed1209-fd82-4075-9bd3-fe77b5270c7b-annotated_6bfe6442-4b6c-4c2f-a088-c7f9c274f96b.png?etag=32ced7063064a4c8a5d9afa4da85b505)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/c7ed1209-fd82-4075-9bd3-fe77b5270c7b-annotated_6bfe6442-4b6c-4c2f-a088-c7f9c274f96b.png?etag=32ced7063064a4c8a5d9afa4da85b505)*

**Browser**: Chrome 102 (Windows 10)
**Screen size**: 1920x1200
**Browser size**: 1920x1057
**[View screen recording](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/screen_recording/de06fdcc-0a99-4831-a437-76c9dc1996dd_screen_recording_4c90a23e-e01c-40d3-aa4f-128281c9b169.webm?etag=23594c12160a47502d6f0f7d7719af60)**
**[Open #268 in Usersnap](https://app.usersnap.com/l/feedback/d25e90ad-8481-4387-9871-b2ae8ee6b512?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

mochet:
Another user reported the same problem on Discord.

The previous votes section contains several votes which no longer exist within `chain state` which meant the user thought that they had 2m tJOY locked--in truth the amount was only 10k tjoy
![image](https://user-images.githubusercontent.com/50830161/171466046-f8d8e55c-971c-4857-8a4a-1152cb1b33d0.png)

[...]

PR #3206, opened 2022-06-08
🔄 Fix recovering Vote stake

RafalPloszka:
#3143

### DIFF
--- a/packages/ui/src/accounts/components/AccountItem/components/LockItem.tsx
+++ b/packages/ui/src/accounts/components/AccountItem/components/LockItem.tsx
@@ -80,7 +80,7 @@ export const LockItem = ({
         <BalanceAmount amount={lock.amount} isRecoverable={isRecoverable} />
         {memberInfo && (
           <div>
-            <DetailLabel>Bound to:</DetailLabel>
+            <DetailLabel>{lock.type === 'Voting' ? 'Voted for' : 'Bound to:'}</DetailLabel>
             <MemberInfo member={memberInfo} onlyTop />
           </div>
         )}
--- a/packages/ui/src/accounts/hooks/useIsVoteStakeLocked.ts
+++ b/packages/ui/src/accounts/hooks/useIsVoteStakeLocked.ts
@@ -6,5 +6,5 @@ export const useIsVoteStakeLocked = (
 ) => {
   // Lock stake if the vote was cast: in current election or to winning candidate
   // Enable stake recovery if election is finished
-  return isLatestElection && !!candidate && (!isElectionFinished || candidate.isCouncilMember)
+  return isLatestElection && (!isElectionFinished || !!candidate?.isCouncilMember)
 }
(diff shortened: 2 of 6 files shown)


packages/ui/src/accounts/components/AccountItem/components/LockItem.tsx
```diff
@@ -80,7 +80,7 @@ export const LockItem = ({
         <BalanceAmount amount={lock.amount} isRecoverable={isRecoverable} />
         {memberInfo && (
           <div>
-            <DetailLabel>Bound to:</DetailLabel>
+            <DetailLabel>{lock.type === 'Voting' ? 'Voted for' : 'Bound to:'}</DetailLabel>
             <MemberInfo member={memberInfo} onlyTop />
```
thesan, line 84: @RafalPloszka this might be correct I can't check rn but are you sure the candidate membership will show here and not the membership of the user who voted ?

RafalPloszka: In VoteLockItem.tsx there is:
`const voteFor = vote?.voteFor?.member`
`const voteForMember = voteFor && asMember(voteFor)`
and `voteForMember` is passed to LockItem `memberInfo` props

So I'm pretty sure, however, this info is not available during Voting Period and if a vote wasn't unsealed. But then `<MemberInfo>` is just not displayed

packages/ui/src/accounts/hooks/useIsVoteStakeLocked.ts
```diff
@@ -6,5 +6,5 @@ export const useIsVoteStakeLocked = (
 ) => {
   // Lock stake if the vote was cast: in current election or to winning candidate
   // Enable stake recovery if election is finished
-  return isLatestElection && !!candidate && (!isElectionFinished || candidate.isCouncilMember)
+  return isLatestElection && (!isElectionFinished || !!candidate?.isCouncilMember)
```
RafalPloszka, line 9: Before this change, all votes in _Voting period_ and sealed votes in _Reveal period_ have undefined `voteFor` field so useVoteStakeLocked returned false even though stake shouldn't be recoverable

thesan: Nice catch! Plus the logic is a lot more readable this way.

But I think for it to work fully in [@/accounts/components/AccountItem/components/LocksDetails.tsx:31](https://github.com/Joystream/pioneer/blob/4231911c6abf17d34a54a969cf5b9a23b9da0337/packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx#L31) the line should be changed to:
```ts
const isVoteStakeLocked = useIsVoteStakeLocked(candidate, {
  isLatestElection: !!candidate,
  isElectionFinished: election?.isFinished,
})
```
Also I just realized the comment is wrong in `useIsVoteStakeLocked`, AFAICS it should be:
```ts
// Always allow stake recovery for votes not cast in the latest election
// also allow when both the election is over and the voted for candidate lost
```

RafalPloszka: Thanks for the comment in `useIsVoteStakeLocked`, it's much better now

thesan (approved):
:raised_hands:

Issue #3211, opened 2022-06-09
Spike: Images unfurling


dmtrjsg:
As a user, when I include a link to an image into my forum post, I expect that link to be presented as an actual image (unfurled) when I save my post, so that other users experience added context and continuity of conversation.

DOD:

- [ ]  Find the library/ approach to implementing the above and advise on design constraints based on tech we choose

⚠️ at current we would not be able to support pasting a screenshot/ image

ivanturlakov:
Tested on staging

**png, jpg, jpeg, gif**

<img width="1419" alt="Снимок экрана 2022-06-27 в 12 11 24" src="https://user-images.githubusercontent.com/46903215/175915610-a7655ac3-e650-4802-a214-7205c095a12a.png">
<img width="1431" alt="Снимок экрана 2022-06-27 в 12 58 46" src="https://user-images.githubusercontent.com/46903215/175915776-b4e289c4-0a47-4664-901d-4f2ac9c84a5e.png">



**svg**

<img width="1432" alt="Снимок экрана 2022-06-27 в 13 03 58" src="https://user-images.githubusercontent.com/46903215/175917282-3df90fc6-5992-4a56-8213-8ace2bb4fe56.png">
<img width="1426" alt="Снимок экрана 2022-06-27 в 13 06 22" src="https://user-images.githubusercontent.com/46903215/175917290-6cc3debc-85b6-4868-bd43-c0024bab8149.png">

PR #3232, opened 2022-06-14
🖼️ Inserting images with url paste on editor

WRadoslaw:
Since we parse editor data with Markdown plugin we lose img data in the process ([here is more](https://daringfireball.net/projects/markdown/syntax#img)) so image resizing isn't possible from MarkdownPreview prespective. Styling on both side on img had to be done to keep feature in bound.

#3211


thesan (approved):
@WRadoslaw the `markdown-editor` build failed because of `@ckeditor/ckeditor5-image/src/autoImage` instead of `@ckeditor/ckeditor5-image/src/autoimage`. When the build fails quietly (with `yarn build`): the `packages/markdown-editor/dist/ckeditor.js` you compiled is used, so everything still works. I think Windows doesn't mind the case mismatch so you were able to compile successfully.

The PR looks good to me, please have a quick look to check if everything still works as intended after my changes before merging it.

thesan:
Also is there a quick way to support svg and/or image urls with a query string ?
e.g:
- https://avatars.dicebear.com/api/bottts/yODEyOTQzNTA4NzYxNzQyNw==.svg
- and: https://picsum.photos/seed/yNDIzODExMTk3MDcwMzMwMg==/800/600.jpg/?blur

If not I think this is good enough for now.

PR #3242, opened 2022-06-15
👛 POC on selecting wallet

WRadoslaw:


WRadoslaw:
Closing as different approach will be taken

PR #3252, opened 2022-06-16
#2220 Put onClick={toggleOpen} to the whole Toggle component.

oleksanderkorn:
Fixing: https://github.com/Joystream/pioneer/issues/2220 with the approach proposed by @thesan


thesan (changes requested):
Turns out this was a bad idea, sorry about that :bow: :bow: :bow: . Now the opened panel is clickable and closes the toggle when clicked (which is not the behavior what we want).

Please either go with @ytdbuolf solution (and just remove the commented brackets on line `109-110`), or add a `header: ReactElement` prop to the `Toggle` component (in which case `children` is what gets toggled), or whatever other solution you prefer !

singulart (approved):

thesan (approved):
Works great! Thanks!

Issue #1794, opened 2021-11-19
Pioneer Responsive Design


bedeho:
# Context 

The product was designed with no consideration of responsiveness, and while the developers have done some work to fill in some gaps, but largely the product does not work for example on mobile devices. 

![Screenshot 2021-11-19 at 10 47 44](https://user-images.githubusercontent.com/437292/142602010-caa62fb1-287d-47ad-9575-82287c111a5b.png)

We thought the the primary device for users would be the desktop, for example because the mobile browsers generally do not support extensions. However, it's also clear that some users will end up viewing Pioneer from a mobile device, for example because they are just linked to some screen which has some information, and currently the product would entirely break in many such cases.

⚠️ Focus: > small desktops/ not full screen/ find all places that can be tuned for removal of overflowing content
🆕 : Mobile user agents. 

# Scope
Suggest to split the scope into 2 parts

## Part 1 - MVP -> Focus 🎯 

Does not include connecting wallet on mobile phone and signing transactions. Only viewing of the pages is concerned here. 

Specifically let's go through all of the modules and suggest the Responsive Updates for the existing features:

- My Profile
- Working Groups
- Proposals
- Council
- Election
- Forum
- Members
- Settings

Break Points in scope: 
xxs - 320px
xs - 425px
sm - 768px
md - 1024px
lg - 1440px

## Part 2 - Out of scope for now ⛔ (let's do it later)

Out of scope breakpoints:
xl - 1920px
xxl - 2560px

Update the design of the transactional flows
 **Forum**
 - Post to forum
 **Proposals**
 - Create proposal
 - Vote on proposal
 **Council/ Elections**
 - Apply to council
 - Vote for council candidate
 - Reveal vote
 **Working groups**
 - Create opening
 - Apply to opening
  
  # Design Support
  
Latest Figma assets for Pioneer

[Wireframes >](https://www.figma.com/files/project/25387558/Wireframes?fuid=1039560990590227876)
[Designs >](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=0-1&t=ig1SfT4xyELPGJl7-0) 
[Storybook >](https://pioneer-2-storybook-git-carthage-joystream.vercel.app/?path=/story/common-tokenvalue--default&args=value:000+000000000)

[...]

PR #3262, opened 2022-06-20
hotfix/join modal style fixed

### LINKED ISSUE #3259: New users ob - Signer CTA disabled on screen zoom < 100%

stepien4soft:
#3259


thesan (approved):
I don't understand the screen resolution of the screenshot in the issue, I could only reproduce this with a window of 615px high or less, but it doesn't look like it's the case there.

IMO most modal should not have a `max-height` and they should be scrolled over the screen instead on being scrollable by default (maybe we could address this in https://github.com/joystream/pioneer/issues/1794).

In the meantime this solution good quick fix :+1:

PR #3307, opened 2022-06-28
⌚ Fix locks recovery time

### LINKED ISSUE #3109: Add LockRecoveryTime logic
Add logic to LockRecoveryTime component for locks:

- [x] BountyLockItem
- [x] CouncilCandidateLockItem
- [x] CouncilorLockItem
- [x] ProposalLockItem
- [x] WorkingGroupLockItem

Out of scope: 

- [x] BoundAccountLockItem, InvitationLockItem - no way to know when lock will be recoverable
- [x] VoteLockItem - recovery time can vary depending on whether candidate will win

RafalPloszka:
Closes #3109

### DIFF
--- a/packages/ui/src/accounts/components/AccountItem/components/LockDate.tsx
+++ b/packages/ui/src/accounts/components/AccountItem/components/LockDate.tsx
@@ -13,5 +13,5 @@ export const LockDate = React.memo(({ createdInEvent }: LockDateProps) => {
     return <TextMedium value>Unknown</TextMedium>
   }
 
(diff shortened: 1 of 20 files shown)


stepien4soft:
@thesan @WRadoslaw @RafalPloszka Just for consideration: In our project we already have a date-fns library installed. Maybe instead of using a custom hook we can use this. Here is a example (instead of Date.parse(time) - Date.now() we can use somethig like formatDostanceToNow()

stepien4soft (approved):

Issue #3203, opened 2022-06-08
Remove network name from the blockheight widget


dmtrjsg:
## Scope
* [ ] Remove Network name from this widget  from all places in the app
* [ ] Currently widget used on Forum

![Screenshot 2022-06-08 at 16.12.48.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/81218492-9063-4ad6-a0f9-f0d0b7448daa)

## Technical Guidelines
A good start would be to delete `NetworkType` from [@/common/types/Block.](https://github.com/Joystream/pioneer/blob/4253b73a6e7493cf90c7b030c23561941649ea58/packages/ui/src/common/types/Block.ts) and removing `network` from all the `queries/xxx.graphql` files, and then removing the network fields by following all the typescript errors.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202444165016487) by [Unito](https://www.unito.io)

dmtrjsg:
@thesan can you hide this behind a flag and create a ticket to show it once we are on mainnet? I think we should not release it with "Mainnet" before it actually happens as users will get confused ..

polikosi:
Verified on https://dao-5i7qbhq9l-joystream.vercel.app/#/profile.
Network name is hidden.

Member profiles:
![Screenshot_145](https://user-images.githubusercontent.com/80429191/178809229-08cc0a94-f2c9-44b1-a1c4-6f6d874297c0.png)


Forum thread header and replies:
![Screenshot_150](https://user-images.githubusercontent.com/80429191/178809090-4814ccd7-c927-4300-b752-3c617a8495f5.png)

Application:
![Screenshot_147](https://user-images.githubusercontent.com/80429191/178808774-5cc5c922-f41f-4a2e-ab1d-eda7e55b56c9.png)

Openings:
![Screenshot_148](https://user-images.githubusercontent.com/80429191/178808641-fdd33e0c-443a-4a8e-ac8a-6018edf10ba5.png)

Bounty:
![Screenshot_149](https://user-images.githubusercontent.com/80429191/178808560-21dbb147-d8a6-4c49-88a3-6860ae7affd8.png)

polikosi:
tested on staging. Network name is hidden.

PR #3340, opened 2022-07-09
🏴 Hide network name on block information

Palllke2015:
Remove network name from the blockheight widget #3203

### DIFF
--- a/packages/ui/src/common/components/BlockTime/BlockInfo.tsx
+++ b/packages/ui/src/common/components/BlockTime/BlockInfo.tsx
@@ -13,13 +13,15 @@ export interface BlockInfoProp {
   inline?: boolean
 }
 
+export const NETWORK_VISIBILITY = false
+
 export const BlockInfo = ({ block, lessInfo, inline }: BlockInfoProp) => (
   <BlockInfoContainer lessInfo={lessInfo} inline={inline}>
     <BlockIcon />
     <span>
       {formatTokenValue(block.number)} {lessInfo && 'block'}
     </span>
-    {!lessInfo && <BlockNetworkInfo>on {block.network} network</BlockNetworkInfo>}
+    {!lessInfo && NETWORK_VISIBILITY && <BlockNetworkInfo>on {block.network} network</BlockNetworkInfo>}
   </BlockInfoContainer>
 )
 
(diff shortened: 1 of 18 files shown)


WRadoslaw (changes requested):
Hello! Firstly thank you for your PR! I have 3 suggestions I'd like you to consider.

1. Let's keep changes from another PR strictly in that PR . Exepction from it would be situation where that changes are required to resolve issue that is taken on in second PR.
2. Please remeber to reqenerate queries when changes were made in any `.graphql` file. This is the script to do it -> `yarn run queries:generate`. 
3. Your solution is making network info kinda toggable (only from developer perspective, but still) and from the other side you are removing property of network from beeing fetched. Basically if you want to remove network from query, you have to go through each casting function and interface that posses that prop and make it optional, preferably remove it and resolve errors that will be caused by that action (that shouln't be many). 

Form the other side we can just remove the line with block string inside `BlockInfo` and do not bother about queries and interfaces, I'd consider it when there is possiblity in the future of utilizing `network` information in any other places. But that question should be addressed to @dmtrjsg.

dmtrjsg:
Discussed pros and cons of
 removing it from the component and keep fetching it in the meantime vs changing the queries to not fetch the network name.
 
Decision made to go with faster option so only change of component required.

packages/ui/src/common/components/BlockTime/BlockInfo.tsx
```diff
@@ -13,13 +13,15 @@ export interface BlockInfoProp {
...
+export const NETWORK_VISIBILITY = false
+
 export const BlockInfo = ({ block, lessInfo, inline }: BlockInfoProp) => (
   <BlockInfoContainer lessInfo={lessInfo} inline={inline}>
     <BlockIcon />
     <span>
       {formatTokenValue(block.number)} {lessInfo && 'block'}
     </span>
-    {!lessInfo && <BlockNetworkInfo>on {block.network} network</BlockNetworkInfo>}
+    {!lessInfo && NETWORK_VISIBILITY && <BlockNetworkInfo>on {block.network} network</BlockNetworkInfo>}
```
WRadoslaw, line 24:
```suggestion
    {/* <BlockNetworkInfo>on {block.network} network</BlockNetworkInfo> */}
```

Plus remove `NETWORK_VISIBILTY` and styled component

WRadoslaw (approved):

Issue #2888, opened 2022-04-22
Additional columns for Members list


dmtrjsg:
## Context

please add columns to Members table for

who referred if any member
time when membership was created (bedeho: Give us some details: I would like to know at what point in time a given membership was created on the All Members table. ID is a proxy for order, but I would like to know time.)

![Screenshot 2022-04-22 at 12.22.12.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/421d0831-1f1d-44e3-a722-ef094603a640)

## Scope

1️⃣  Redesign as per figma: [👉 Figma Link here](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9825%3A387305)

As we do not have a lot of space to accommodate all the information, I propose to relocate the member ID  directly below the membership handle (as Dmitry rightly pointed out during the meeting, no one will sort the list by ID number). 

It is also not necessary to show council members as a separate field, since there are only a few of them for 4k+ members. So I think it would be a good idea to move this indicator as a check icon just after the user handle.

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/c8e4d113-50a6-4ec7-8f52-1c392debe46f)

2️⃣  Lock icons cosmetic fix

<img width="366" alt="Screenshot 2022-07-06 at 08 47 30" src="https://user-images.githubusercontent.com/93924525/177498363-b36c89aa-d248-497e-84b6-0cac6c9623d1.png">

MattAugustynowicz:
[👉 Figma Link here](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9825%3A387305)

As we do not have a lot of space to accommodate all the information, I propose to relocate the member ID  directly below the membership handle (as Dmitry rightly pointed out during the meeting, no one will sort the list by ID number). 

It is also not necessary to show council members as a separate field, since there are only a few of them for 4k+ members. So I think it would be a good idea to move this indicator as a check icon just after the user handle.

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/c8e4d113-50a6-4ec7-8f52-1c392debe46f
)

freakstatic:
@singulart please assign

singulart:
> @freakstatic Could we also tackle the misplacement of the lock icons in this ticket?
> 
[...]

PR #3351, opened 2022-07-13
add additional column

Palllke2015:
Additional columns for Members list #2888


packages/ui/src/memberships/components/MemberReferrer/MemberReferre.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react'
```
WRadoslaw, line 1: Typo in file name

WRadoslaw (changes requested):
Hey! Nice work and thanks for PR! From my side: 

1. Since throughout project we keep styled components under the function components let's keep doing it for the sake of uniformity. If the components you created are used in more than one place and you really think that they deserve seperate file, as I remeber this kind of files are just named `components` instead of `Fields`.

2. I think istead of creating new fragment, query, cast and interface just to fetch one more field is kinda overkill. I suspect that you approach was based on not overfetching that graphql offers, but on the other adding so much code to use it one place when it differ by only one parameter isn't desired. IMHO we should just add `invitedBy` to the `MemberWithDetailsFields` and remove `MemberAllWithDetailsFields` and its corresponding interface `MemberWithReferrer`. The overfetch won't be so big, since not many users use invite function and if they do numbers are pretty small.

WRadoslaw (changes requested):
Unfortunately it's still not it. If I mislead you somehow, sorry. The main goal is to not touch almost any casting function. The only change in graphql file was to add
```
invitedBy {
    ...MembersField
}
```
to the `MemberWithDetailsFields`. I see that i forgot to mention that in fact you will need to add new query to fetch all users (same as `GetMembers`, but for `MemberWithDetailsFields` fragment. Then it's only the matter of creating new hook for it and adding invitedBy to the interface and cast (in cast u can use `invitedBy: fields.invitedBy.map(asMember)` if I'm correct) and it the end just use `MemberWithDetails` interface throught your solution.

Palllke2015:
> Unfortunately it's still not it. If I mislead you somehow, sorry. The main goal is to not touch almost any casting function. The only change in graphql file was to add
> 
> ```
> invitedBy {
>     ...MembersField
> }
> ```
> 
> to the `MemberWithDetailsFields`. I see that i forgot to mention that in fact you will need to add new query to fetch all users (same as `GetMembers`, but for `MemberWithDetailsFields` fragment. Then it's only the matter of creating new hook for it and adding invitedBy to the interface and cast (in cast u can use `invitedBy: fields.invitedBy.map(asMember)` if I'm correct) and it the end just use `MemberWithDetails` interface throught your solution.

It's ok. I will update my PR.

PR #3355, opened 2022-07-14
fix: Candiate bullet point in the voting modal

gyroflaw:


packages/ui/src/council/modals/VoteForCouncil/components/CandidacyReview.tsx
```diff
@@ -33,9 +33,9 @@ export const CandidacyReview = memo(({ candidate, minStake }: CandidacyReviewPro
       <h4>{candidate.info.title}</h4>
 
       <CandidacyPointList>
-        <CandidacyPoint>The More Important the Work, the More Important the Rest</CandidacyPoint>
-        <CandidacyPoint>How to build a loyal community online and offline</CandidacyPoint>
-        <CandidacyPoint>Helping a local business reinvent itself</CandidacyPoint>
+        {candidate.info.bulletPoints.map((bulletPoint, index) => (
+          <CandidacyPoint key={index}>{bulletPoint}</CandidacyPoint>
```
freakstatic, line 37: `key={index}` is probably not needed here

freakstatic (changes requested):
@gyroflaw thanks for the PR!
I added a comment please check

packages/ui/src/council/modals/VoteForCouncil/components/CandidacyReview.tsx
gyroflaw: yeah, you are right,
But in my opinion react requires key in the map elements to identify map elements.
In this case, we don`t use index for key, and use something related to bullet points.
Since graphql query returns only string[] for bp array, so I used index for key.
If you think the key is not needed, I can remove it.

freakstatic: yeah please do since is not used right now

Issue #2311, opened 2022-02-23
Caching and Polling Implementation For Pioneer


dmtrjsg:
## Context
The overall objective is to serve the most up to date state of the platform to users, while ensuring the app works fast and smooth. 

The objective of this task is to rollout a suitable caching and polling approach for the variety of pages, which do not involve a transaction flow.

* [x] Council Page
* [x] Bounty Preview
* [x] Openings Current and Upcoming
* [x] Current Candidates list
* [x] Proposal Page + discussions (preview)
* [x] Current Proposals list
* [x] Forum categories

Out of scope ⛔ :
* [x] ~Current Bounties list~ [deprecated]
* [ ] #3352
* [ ] ~Overview~

## Scope
* [ ] Create a hook to enable polling "useRefetchQueries" (already used in Bounty.tsx, packages/ui/src/app/pages/Council/Council.tsx and Election..)
* [x] Define the polling frequency for each -> Done here: https://github.com/Joystream/pioneer/pull/3221
* [ ] Implement caching and polling for all checkboxes that are in scope for this ticket

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202512541954625) by [Unito](https://www.unito.io)

dmtrjsg:
@ QA: Polling is set to 6 seconds (block time) - just check on each page that it is not flicking / trying to load every 6 seconds (like it is now on prod with council).

polikosi:
Tested on staging. 
✅Council Page
✅Bounty Preview
🚫Openings Current - requests aren't sent
✅Openings Upcoming
✅Current Candidates list
🚫Proposal page + discussions (preview) - requests aren't sent
✅Current Proposals list
✅Forum categories

singulart:
@dmtrjsg Failed in QA, please see outcome

dmtrjsg:
We've triaged it with @WRadoslaw and identified the issue:

1. Proposals page was not done in this PR, so that was skipped (now refetches every second block, so 12 seconds on staging). We will also remove the refetch from past proposals discussion page.  On Carthage site, this worked fine. So once we merge Carthage this would be okay, if not we will re-raise it. 
2. Openings current - just checked on staging, and worked like a charm for us :) 

Based on these comments marking this as QA-passed for this release. Thank you for raising @singulart

PR #3359, opened 2022-07-15
🐻‍❄️ Polling on chosen pages

WRadoslaw:
#2311

### DIFF
--- a/packages/ui/src/common/hooks/useRefetchQueries.ts
+++ b/packages/ui/src/common/hooks/useRefetchQueries.ts
@@ -25,7 +25,15 @@ export const useRefetchQueries = (
     if (couldRefetchNext.current && when) {
       if (interval) {
         const handler = setInterval(() => {
-          apolloClient.refetchQueries({ include })
+          apolloClient.refetchQueries({
+            include,
+            onQueryUpdated(_, { complete }) {
+              if (complete) {
+                isRefetched.current = true
+              }
+              return complete
+            },
+          })
           isRefetched.current = true
         }, interval)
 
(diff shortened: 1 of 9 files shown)


packages/ui/src/common/hooks/useRefetchQueries.ts
```diff
@@ -25,15 +26,19 @@ export const useRefetchQueries = (
       if (interval) {
         const handler = setInterval(() => {
           apolloClient.refetchQueries({ include })
+          isRefetched.current = true
```
thesan, line 29: `isRefetched` is a good idea for list of entities where the default is `[]` instead of `undefined` but if it starts refetching before the query even completed the UI might show something like "There are no current proposals yet" when it's not the case. To solve this we could:
1. Return `undefined` when `data` is `undefined` but it means changing a bunch of hooks (also I just discovered a [`networkStatus` prop](https://www.apollographql.com/docs/react/data/queries#inspecting-loading-states) that could be used but it might not be necessary) on top of that something would have to be passed to `useRefetched` `when` to not refetch not completed query.
2. I also just found out that `refetchQueries` has a [`onQueryUpdated` option](https://www.apollographql.com/docs/react/data/refetching/#refetching-selectively). The way I understand it the second param can be used refetch only completed query. But even better in the 3rd example the query is updated accordingly to it's cache policy with: `return observableQuery.reobserve()` I think it would make the `isLoading` not switch back to `true` (I'm not sure though...)

WRadoslaw: Fix applied in new commit

PR #3363, opened 2022-07-18
:neckbeard: New columns for members list

WRadoslaw:
Follow-up: #2888 

UI side of work: @Palllke2015


WRadoslaw:
@singulart hey, #2888 was decided to be handled internally with this PR

singulart:
@dmtrjsg Interesting case here. Which side should be given SPs in this case - community team or core team? :)

dmtrjsg:
@singulart I'd say in this case community as bulk of the work was carried out and Radek's fix can be used as showcase for the right queries and part of the learning curve.

PR #3372, opened 2022-07-20
🆕 New graphql schema for carthage

WRadoslaw:


thesan (approved):
Nice! I was afraid there would be a lot of renamed stuff

Issue #3394, opened 2022-07-27
Carthage: loading wallet modal on local storage


dmtrjsg:
Progress modal not disappearing on carthage

PR #3395, opened 2022-07-27
♾️ Fix for infinite waiting modal

WRadoslaw:
#3394


thesan (changes requested):
I think you should also update the dependencies (since `status`, `balance`, `transactionFee` will remain empty while the modal should actually show):
```ts
}, [status, !balance, !transactionFee, !transaction, !signer])
```

thesan (approved):
:+1:

Issue #2959, opened 2022-05-03
Claim vested $JOY


bedeho:
## Context
On mainnet, $JOY will be encumbered by a vesting schedule, and giving users a simple way to claim such $JOY, and seeing what the vesting vs. vested balance for individual accounts.
:information_source:  Here's technical details https://polkadot.js.org/docs/substrate/extrinsics/#vesting

## Design

[LINK TO FIGMA > ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=10535%3A402088)

## Scope

Individual Locks on account level:

- [ ] Staring block
- [ ] Ending block
- [ ] Original Vesting
- [ ] Unlocked per block
- [ ] Locked
- [ ] Claimable

Widgets for compound info on all locks across accounts:

- [ ] Total Vesting
- [ ] Total Locked
- [ ] Total Claimable
- [ ] Claim CTA
- [ ] Ability to claim full amount  only (user cannot specify how much to claim from single account/ Claiming is possible from one account at a time)

☝️  @ QA

<img width="339" alt="Screenshot 2022-08-10 at 15 50 33" src="https://user-images.githubusercontent.com/93924525/183934274-d881102f-8016-4124-b6f4-379403a48f75.png">

Test with multiple accounts for single memebrship and multiple locks per single account

bedeho:
> Add a new tab "Vested Tokens"

This description does not spell it out entirely, but what entity does a row here correspond to? its not tokens or a token, its an account, vesting occurs per account. The natural place of such information is the account table, why a new table?

bedeho:
I don't think I understand this, feel free to supply a low-fi sketch or something.

dmtrjsg:
Issue description re-written after confirmed with Bedeho.

bedeho:
1. I think so yes, but even if you could claim them to some other account, this can be done in a second transfer later, so lets not overcomplicate the design with this variation even if possible.
2. What do you mean by unilateral, do you mean identical? if so then no, and there will be future stakeholders that may be subject to new vesting schedules.
3. This question was a  bit unclear, but on mainnet launch, each account will have exactly one vesting allocation. In the future, its unclear whether we will do more vesting allocations, but they would almost certainly be on fresh accounts, so the vast majority of accounts with vesting will have 1 allocation.

dmtrjsg:
@Adam based on the above:
[...]

PR #3460, opened 2022-08-04
🗜️ Introduction of vesting

WRadoslaw:
#2959


packages/ui/src/accounts/modals/ClaimVestingModal/ClaimVestingModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+    if (balances) {
+      Object.entries(balances).forEach(([key, value]) => {
+        if (value.isVesting) {
+          setSelectedAccount(allAccounts.find((a) => a.address === key))
+        }
+      })
+    }
+  }, [!balances])
+
+  const transaction = useMemo(() => api?.tx.vesting.vest(), [])
```
thesan, line 44: This would work:
```suggestion
  const { transaction } = useTransactionFee(selectedAccount?.address, () => api?.tx.vesting.vest())
```

Although ideally in the future we should probably refactor `useSignAndSendTransaction` to accept a callback and update the `TransactionFeesContext`

thesan (changes requested):
Great job on delivering this one so fast again :bow: 

I added a few CR but more importantly please add stories for `VestingListItem`, `ClaimVestingModal`, and `SelectVestingAccount`. It would be really useful to help visualizing this workflow.

For now I am not sure if anything needs tests but please add some if/where you think they would be useful.

packages/ui/src/accounts/modals/ClaimVestingModal/ClaimVestingModal.tsx
WRadoslaw: If we take transaction out of `useMemo` it will cause loop cause `useSignAndSendTransaction` will receive new object on every re-render. 

Maybe we can stick to this? WDYT
```
  const transaction = useMemo(() => api?.tx.vesting.vest(), [])

  useTransactionFee(selectedAccount?.address, () => transaction, [transaction])

  const { isReady, sign } = useSignAndSendTransaction({
    transaction,
    signer: selectedAccount?.address ?? '',
    service: service as any,
  })
  ```

File: packages/ui/src/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem.stories.tsx [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+import { Observable } from 'rxjs'
+
+import { VestingLockListItem } from '@/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem'
+import { ApiContext } from '@/common/providers/api/context'
+
+import { stubApi } from '../../../../../../test/_mocks/transactions'
+
+export default {
+  title: 'Accounts/VestingLockListItem',
+  component: VestingLockListItem,
```
thesan, line 14: But this one is the one that can be used to understand the vesting workflow. So I re-wrote this story with parameters
It also made me realized there's probably a bunch of subscriptions to `rpc.chain.subscribeNewHeads` used only one time all around the app, which impact the performances through polkadot.js and webworker interaction and a lot of useless re-render. We should improve `useObservable` to deal with this (but in another PR ofc).

```ts
import { BN_ZERO } from '@polkadot/util'
import { Meta, Story } from '@storybook/react'
import BN from 'bn.js'
import { clamp } from 'lodash'
import React, { useMemo } from 'react'

import { VestingLockListItem } from '@/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem'
import { JOY_DECIMAL_PLACES } from '@/common/constants'
import { ApiContext } from '@/common/providers/api/context'

import { stubApi, stubQuery } from '../../../../../../test/_mocks/transactions'

const blockControl = { control: { type: 'range', min: 0, max: 1000 } }

export default {
  title: 'Accounts/VestingLockListItem',
  component: VestingLockListItem,
  argTypes: {
    currentBlock: blockControl,
    startingBlock: blockControl,
  },
} as Meta

const defaultProps = {
  currentBlock: 50,
  locked: 1000,
  perBlock: 1,
  startingBlock: 0,
}

export const Default: Story<typeof defaultProps> = ({ currentBlock, locked, perBlock, startingBlock }) => {
  const api = useMemo(() => {
    const api = stubApi()
    stubQuery(api, 'rpc.chain.subscribeNewHeads', { number: { toBn: () => BN_ZERO } })
    return api
  }, [])

  const endBlock = useMemo(() => locked / perBlock + startingBlock, [locked, perBlock, startingBlock])

  const vested = useMemo(() => clamp((currentBlock - startingBlock) * perBlock, 0, locked), [currentBlock, endBlock])

  return (
    <ApiContext.Provider value={api}>
      <VestingLockListItem
        startingBlock={new BN(startingBlock)}
        endBlock={new BN(endBlock)}
        perBlock={new BN(perBlock * 10 ** JOY_DECIMAL_PLACES)}
        locked={new BN(locked * 10 ** JOY_DECIMAL_PLACES)}
        vested={new BN(vested * 10 ** JOY_DECIMAL_PLACES)}
      />
    </ApiContext.Provider>
  )
}

Default.args = defaultProps

```

thesan (changes requested):
Last CRs

packages/ui/src/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem.stories.tsx
WRadoslaw: Sure thing!

packages/ui/src/accounts/components/AccountItem/AccountItem.tsx
```diff
@@ -38,7 +37,10 @@ export const AccountItem = ({ account }: AccountItemDataProps) => {
           <TokenValue value={balance?.locked} isLoading={!isDefined(balance?.locked)} />
           <AccountLocks locks={balance?.locks} />
         </ValueAndLocks>
-        <TokenValue value={balance?.recoverable} isLoading={!isDefined(balance?.recoverable)} />
+        <TokenValue
+          value={balance?.recoverable?.add(balance?.vestedClaimable ?? BN_ZERO)}
+          isLoading={!isDefined(balance?.recoverable)}
+        />
```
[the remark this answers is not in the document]
WRadoslaw: To remove these null checks we would have to change a logic a little bit, since RN w/o `?` we will get runtime errors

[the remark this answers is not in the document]
thesan: No need, my bad

packages/ui/test/common/model/formatters.test.ts
```diff
@@ -70,7 +70,7 @@ describe('formatters', () => {
     const dateString = '1983-10-01T06:42:00.155Z'
 
     it('Default format', () => {
-      expect(formatDateString(dateString)).toMatch(/^[0-9]{2}\/[0-9]{2}\/[0-9]{4}, [0-9]{2}:[0-9]{2} (AM|PM) [A-Z]+/)
+      expect(formatDateString(dateString)).toMatch(/^[0-9]{2}\/[0-9]{2}\/[0-9]{4}, [0-9]{2}:[0-9]{2}(AM|PM) [A-Z]+/)
```
thesan, line 73: That is not a great test :sweat_smile:

thesan (approved):
Once again great job :bow: :bow: :bow: :bow:

Issue #3418, opened 2022-07-29
[USERSNAP] Carthage: Proposal created with wrong name


dmtrjsg:
⚠️ pls remember to create a ticket with the long term solution. 

**Taken at**: https://dao.joystream.org/#/working-groups
**Fields**:

* Houston, we have a problem.: Carthage: Proposal created with wrong name
* Comment: https://dao-git-carthage-joystream.vercel.app/#/proposals/preview/3?network-config=https://54.172.154.45.nip.io/config.json
* Feedback: dmtrjsg

**Browser**: Chrome 103 (macOS Catalina)
**Screen size**: 1512x982
**Browser size**: 1453x895
**[Open #307 in Usersnap](https://app.usersnap.com/l/feedback/8dfad3b3-2981-4218-b5f5-84d8d4d9a183?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202691676627123) by [Unito](https://www.unito.io)

dmtrjsg:
in the mappings theres a sanitising string function which could be changed to check if its a hexadecimal and convert to a properly UTF8 string.

dmtrjsg:
Temp solution> Decision made to disable special characters in proposal titles. Add validation "Special characters not allowed".

Long term solution >

dmtrjsg:
Agreed to add:  !@£$%^&*()# to the allowed characters. :"><"'; in the change request.

PR #3476, opened 2022-08-08
😲 Extra validation on proposal title

WRadoslaw:
Temporary solution for #3418


packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
```diff
@@ -151,7 +151,10 @@ export const schemaFactory = (titleMaxLength: number, rationaleMaxLength: number
       stakingAccount: StakingAccountSchema.required('Field is required'),
     }),
     proposalDetails: Yup.object().shape({
-      title: Yup.string().required('Field is required').max(titleMaxLength, 'Title exceeds maximum length'),
+      title: Yup.string()
+        .required('Field is required')
+        .max(titleMaxLength, 'Title exceeds maximum length')
+        .matches(/^[a-zA-z 1-9]*$/, 'Title includes forbidden characters'),
```
thesan, line 157: Don't we want at least most of ASCII ? (or maybe even a bit more)
```suggestion
        .matches(/^[\x0A-\x7E]*$/, 'Title includes forbidden characters'),
```

thesan (changes requested):

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
WRadoslaw: As on the call, it will be updated with most ASCII chars - `\x20-\x7E`

WRadoslaw, line 157:
```suggestion
        .matches(/^[\x20-\x7E]*$/, 'Title includes forbidden characters'),
```

thesan (approved):

Issue #3487, opened 2022-08-11
[USERSNAP] I can't create proposal with a new account


dmtrjsg:
## Scope

- [x] Update the modal to display correct validation message (already fixed by Radek)
- [ ] Change the validation of sufficient balance to cover TX fees everywhere to consider transferable balance + invitation lock. Cases where to test first (Proposal, Forum Posting, Voting for Candidate) 
- [x]  invitation lock funds can be used for staking, as this is a non-rivalrous lock. ✅  confirmed, we should verify it.

+ ⚠️ Check that account has sufficient balances for both TX fees and minimum staking amount. 

## Context 
**Taken at**: https://dao-git-carthage-joystream.vercel.app/#/proposals/current
**Fields**:

* Houston, we have a problem.: I can&#39;t create proposal with a new account
* Comment: I can create proposal after I click on &#34;create proposal&#34; with a new account

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/b940af75-cc6a-4c46-adfb-ed8574fa851b-annotated_aeadfd1a-d74c-4b05-a93d-0cea308ce87a.png?etag=f230aab1e2f347726d855b3f00dadbda)
_[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/b940af75-cc6a-4c46-adfb-ed8574fa851b-annotated_aeadfd1a-d74c-4b05-a93d-0cea308ce87a.png?etag=f230aab1e2f347726d855b3f00dadbda)_

**Browser**: Firefox 103 (Windows 10)
**Screen size**: 1536x864
**Browser size**: 1536x739
**[Open #328 in Usersnap](https://app.usersnap.com/l/feedback/869a38e8-6f9f-459e-a030-f054c5fdfbe2?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202783425457211) by [Unito](https://www.unito.io)

PR #3506, opened 2022-08-16
📨 Fix on invitation lock logic

WRadoslaw:
#3487


packages/ui/src/accounts/modals/MoveFoundsModal/MoveFundsModal.tsx
```diff
@@ -34,39 +34,56 @@ const actionNameMapper = (lock?: LockType) => {
 export const MoveFundsModal = () => {
   const {
     hideModal,
-    modalData: { requiredStake, lock },
+    modalData: { requiredStake, lock, isFeeOriented },
```
thesan, line 37: @WRadoslaw is the `isFeeOriented` flag used anywhere ?

WRadoslaw: It's not, but it's should be. I'll cover usages in next commit.

thesan (approved):
LGTM ! Nice job :raised_hands:

Issue #3453, opened 2022-08-02
[USERSNAP] Funding proposal transaction fails


dmtrjsg:
## Scope

QA:

- [ ] ⚠️ this ticket needs to be retested on the newest playground:

>https://54.172.204.195.nip.io/network/config.json](https://dao-git-carthage-joystream.vercel.app/#/settings?network-config=https://54.172.204.195.nip.io/network/config.json)

## Context

**Taken at**: https://dao-git-carthage-joystream.vercel.app/#/proposals/current
**Fields**:

* Houston, we have a problem.: Founding proposal transaction fails
* Comment: Expecting created proposal
* Link to polkadot block: https://polkadot.js.org/apps/?rpc=wss://54.172.154.45.nip.io/ws-rpc#/explorer/query/0x8e857868d196d8592a0cb4515b94e2eeafae5f4b31344099cf0efdb1d6a555c1
* Feedback: Ivan T#1098

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/532992a2-1d96-4d27-bd4d-a46a60d1f1c5-annotated_4d4c1e27-8d43-41f8-8b21-1788ac6ef0cd.png?etag=f621caee27ee0441a82779931123690b)
_[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/532992a2-1d96-4d27-bd4d-a46a60d1f1c5-annotated_4d4c1e27-8d43-41f8-8b21-1788ac6ef0cd.png?etag=f621caee27ee0441a82779931123690b)_

**Browser**: Chrome 103 (macOS Catalina)
**Screen size**: 1440x900
**Browser size**: 1440x531
**[Open #320 in Usersnap](https://app.usersnap.com/l/feedback/a7c29ae2-2594-40d7-a204-5368076c7492?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202705820155406) by [Unito](https://www.unito.io)

ivanturlakov:
[...]

PR #3513, opened 2022-08-17
🙈 Ignoring validation for undefined constants

WRadoslaw:
resolves last comment from #3453


thesan (approved):
Really nice solution :raised_hands:

It's actually lot less change on each validation than I thought. So if you prefer you can remove the `?? 0` in the other `maxMixed` and `lessThanMixed` validations (defaulting to `0` is usually ok for minimum values). But it's fine for me either way.

WRadoslaw:
Thanks! 🥳 

I don't see any more `?? 0` fallbacks on these validators. The only fallbacks left are ones that utilize `yup` default `min` `max` methods and they do not accept `undefined` values.

Issue #3464, opened 2022-08-05
Fix deleted sub-category display in Pioneer


dmtrjsg:
## Context
Forum smoke testing results" https://joystream.notion.site/Report-Smoke-testing-Forum-4e55316df35c4d94b0f2380dadf20b46

⚠️ pls point the PR to Carthage branch

## Scope
* [ ] Fix erroneous display of deleted subcategory in Forum

![Screenshot 2022-08-05 at 13.41.16.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/e3034fd5-201a-4d3c-896d-4a5c3d1d303f)-

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202727563990196) by [Unito](https://www.unito.io)

singulart:
SP estimate is missing

dmtrjsg:
@singulart added

ivanturlakov:
✅ Tested on https://dao-git-carthage-joystream.vercel.app/#/settings?network-config=https://54.172.154.45.nip.io/network/config.json

Category "Test subcategory" has been created in "test".

<img width="1183" alt="Снимок экрана 2022-08-29 в 15 37 26" src="https://user-images.githubusercontent.com/46903215/187203745-6cfd35db-8f12-4dde-b56b-067145854b60.png">
<img width="1180" alt="Снимок экрана 2022-08-29 в 15 38 16" src="https://user-images.githubusercontent.com/46903215/187203755-caa84900-5380-42c1-abe7-ccbd6176bf88.png">


Category "Test subcategory" has been deleted in "test".

<img width="1183" alt="Снимок экрана 2022-08-29 в 15 39 56" src="https://user-images.githubusercontent.com/46903215/187203759-029f8f8a-1928-4628-a0ee-c4ef09dc98fb.png">
<img width="1181" alt="Снимок экрана 2022-08-29 в 15 40 11" src="https://user-images.githubusercontent.com/46903215/187203768-8daf55af-40ef-4ffa-bee9-6d80fb7d7c7a.png">

PR #3528, opened 2022-08-18
👻 Fix on forum phantom subcategories

WRadoslaw:
- [x] #3464

### DIFF
--- a/packages/ui/src/forum/types/ForumCategory.ts
+++ b/packages/ui/src/forum/types/ForumCategory.ts
@@ -70,13 +70,17 @@ export const asForumCategory = (fields: ForumCategoryFields): ForumCategory => (
   ...asBaseForumCategory(fields),
   moderators: fields.moderators?.map(({ id, membership }) => ({ id, handle: membership.handle })) ?? [],
   subcategories:
-    fields.forumcategoryparent?.map((fields) => ({ ...asSubCategory(fields), status: fields.status.__typename })) ?? [],
+    fields.forumcategoryparent
+      ?.filter((fields) => fields.status.__typename !== 'CategoryStatusRemoved')
+      .map((fields) => ({ ...asSubCategory(fields), status: fields.status.__typename })) ?? [],
 })
 
 export const asArchivedForumCategory = (fields: ArchivedForumCategoryFields): ForumCategory => ({
   ...asBaseArchivedForumCategory(fields),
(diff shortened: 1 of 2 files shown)


packages/ui/src/forum/types/ForumCategory.ts
```diff
@@ -70,13 +70,17 @@ export const asForumCategory = (fields: ForumCategoryFields): ForumCategory => (
...
+      ?.filter((fields) => fields.status.__typename !== 'CategoryStatusRemoved')
+      .map((fields) => ({ ...asSubCategory(fields), status: fields.status.__typename })) ?? [],
 })
 
 export const asArchivedForumCategory = (fields: ArchivedForumCategoryFields): ForumCategory => ({
   ...asBaseArchivedForumCategory(fields),
   subcategories:
-    fields.forumcategoryparent?.map((fields) => ({ ...asSubCategory(fields), status: fields.status.__typename })) ?? [],
+    fields.forumcategoryparent
+      ?.filter((subcategory) => subcategory.status.__typename !== 'CategoryStatusRemoved')
```
thesan, line 82: `CategoryStatusRemoved` is supposed to propagate down to subcategories. So when a categories gets archive, it's subcategories should be archived too.

E.g on `query.joystream.org`:
image

So here we don't want to filter the `CategoryStatusRemoved`

packages/ui/src/forum/types/ForumCategoryWithDetails.ts
```diff
@@ -8,5 +8,8 @@ export interface ForumCategoryWithDetails extends Omit<ForumCategory, 'subcatego
 
 export const asCategoryWithDetails = (fields: ForumCategoryWithDetailsFieldsFragment): ForumCategoryWithDetails => ({
   ...asBaseForumCategory(fields),
-  subcategories: fields.forumcategoryparent?.map(asForumCategory) ?? [],
+  subcategories:
+    fields.forumcategoryparent
+      ?.filter((subcategory) => subcategory.status.__typename !== 'CategoryStatusRemoved')
```
thesan, line 13: But here we should probably decide whether to filter `CategoryStatusRemoved` subcategories based on `fields.status.__typename`

thesan (changes requested):

thesan (approved):

Issue #3536, opened 2022-08-22
Cannot create proposal with Vesting lock


dmtrjsg:
## Context
This is the final overview for Vesting and Invitation locks supported tx: https://github.com/Joystream/pioneer/issues/3535
Bug identified here> https://github.com/Joystream/pioneer/issues/3495#issuecomment-1219891897

## Scope
* [ ] Fix last modal on Pioneer to validate based on total funds - rivalrous locks. 

Comment: 

We seem to have fixed it for invitation locks, but remains to be done for vesting locks

⚠️ Comment from Theo:
> Yes it looks like a bug for all locks because it is the bind staking account transaction which is a non-rivalrous lock

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202838539805224) by [Unito](https://www.unito.io)

PR #3538, opened 2022-08-22
🧛 Fix for fee balance validation across multiple modals

WRadoslaw:
- #3536

### DIFF
--- a/packages/ui/src/common/hooks/useSignAndSendTransaction.ts
+++ b/packages/ui/src/common/hooks/useSignAndSendTransaction.ts
@@ -4,6 +4,9 @@ import { Hash } from '@polkadot/types/interfaces/types'
 import { useCallback, useEffect, useState } from 'react'
 import { ActorRef } from 'xstate'
 
+import { useBalance } from '@/accounts/hooks/useBalance'
+import { getFeeSpendableBalance } from '@/common/providers/transactionFees/provider'
+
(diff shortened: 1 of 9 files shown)


packages/ui/src/common/hooks/useSignAndSendTransaction.ts
```diff
@@ -50,5 +54,6 @@ export const useSignAndSendTransaction = ({ transaction, signer, service, skipQu
     paymentInfo,
     sign,
     isReady,
+    noFeeFunds: Boolean(balance && paymentInfo && getFeeSpendableBalance(balance).lt(paymentInfo.partialFee)),
```
thesan, line 57: The logic is really nice! Again a PR which fix a bug while simplifying the code :raised_hands: 

I'm just confused by the `noFeeFunds` name. `hasFunds` was lot clearer IMO. Or maybe we could even use `canAfford` to be consistent with whats returned by `useTransactionFee` elsewhere.

WDYT ?

thesan (changes requested):

packages/ui/src/common/hooks/useSignAndSendTransaction.ts
WRadoslaw: Thanks! Sure, sounds better and seems more straightforward 😃

thesan (approved):

Issue #3153, opened 2022-06-02
Forum Category Page


dmtrjsg:
## Context
Forum UX facelift upd

## Design Preview

[Link to Figma> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9364%3A339383)

![](https://user-images.githubusercontent.com/93924525/171611749-395475b8-274a-4e5b-a9ff-3a2339a66df3.png)

## Scope
:one: **Popular threads**
![](https://user-images.githubusercontent.com/93924525/171612242-f27c7b99-d2e5-41b1-ae01-09f9976bb592.png)

* [ ] Popular threads - Top 10 threads sorted by number of thread replies
* [ ] Carousel and card are same as Latest threads
* [ ] Fetch / show number based on screen width

:two: **Subcategories**
![](https://user-images.githubusercontent.com/93924525/171612180-ef116e33-9ce3-4bd5-adec-5a4a08a921b6.png)

* [ ] On hover show sub-categories
* [ ] On lowest level on hover keep showing description
* [ ] Categories should wrap on smaller screens

out of scope: ⛔ 
* [ ] ~Display 4 sub-categories in a grid and button "show all"~

:three: **Threads**

![](https://user-images.githubusercontent.com/93924525/171612195-e519e6e3-cb8b-4828-b13f-e080ab0a03f1.png)

* [ ] Table 
* [ ] Total counter on top as pill
* [ ] Truncated thread name
* [ ] Replies counter
* [ ] Last replies
* [ ] Author

Basically no change from production just copy pasta :) 

:four: 

* [ ] Search bar
* [ ] Moderators avas

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202464660764776) by [Unito](https://www.unito.io)

Ken-tech-max:
Hi, Isonar, Can you explain the issues one by one?
I want to ensure the issues and jump to development.

gyroflaw:
hey I am gyroflaw#5905

traumschule:
this was inactive for a bit, unassigning for now because #3029 is in `mainnet` scope. #3562 was reverted, @thesan how should we go from here?

dmtrjsg:
@traumschule we've decided to finish the forum in the team internally, so pls feel free to ignore all Forum tasks for communty dev for the time being.

dmtrjsg:
was done in a diff PR but worth testing based on issue description

[...]

Issue #3157, opened 2022-06-02
Forum Landing Page - Main Tab - categories


dmtrjsg:
## Context
UX facelift of Forum Module

## Designs 

[Link to Figma >](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=8966%3A319910)

This page > https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9364%3A333164

## Scope
**Categories Section on Main Forum Tab**

![](https://user-images.githubusercontent.com/93924525/171575443-6a0a2770-8beb-4859-9688-86bf39c85565.png)

AC: 

* [ ] Header - Categories
* [ ] Order of categories is hard-coded with fixed positioning
* [ ] Top 8 categories for the build:

1. Working Groups
. ~Bounties~  :warning:  Hide Bounties from Forum
1. Marketing
1. Development
1. Validators 
1. Discussions
1. Help & Support

* [ ] ~Show all button expands the rest of the categories:~

1. Content Creators
1. Creatives
..
1. Governance
1. Grants and Funding
1. Education
1. Atlas Feedback
1. Lounge

* [ ] New categories get added to the bottom of the list

Out of scope ⛔ : 
- [ ] #3364

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202464581204491) by [Unito](https://www.unito.io)

secretgir1:
Hi @dmtrjsg 

Could you please provide link for this page in figma?

dmtrjsg:
@secretgir1 sure, [here's the link>](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9364%3A333164)

Will also review other tickets in this epic and do the same if needed.

secretgir1:
> @secretgir1 sure, [here's the link>](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9364%3A333164)
> 
> Will also review other tickets in this epic and do the same if needed.

It would be nice if you provide design of see all button, the rest is understandable, thanks.

dmtrjsg:
@secretgir1 thanks for the question! 

I've isolated this AC to a separate issue and added to the backlog for design team. This will be done in a few weeks time as we are busy on smth else on Atlas rn. 

[...]

Issue #3151, opened 2022-06-02
Forum: Landing Page - Main Tab - Latest Threads


dmtrjsg:
## Context
UX facelift of Forum Module

## Designs Preview

[LInk to Figma> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9364%3A333164)

Thread Card
![](https://user-images.githubusercontent.com/93924525/171571255-9eed3d5d-d6b8-4ece-acd9-73d52d6c0219.png)

Latest Threads

![](https://user-images.githubusercontent.com/93924525/171567397-0032aefd-4979-40ba-b188-ba30d285c603.png)

## Scope
:one:   **Latest Threads Carousel.**

AC: 

* [ ] Newest on the left
* [ ] Show all threads in the carousel 
* [ ] Load first X depending on page width 
* [ ] Chevron click loads a full set of new cards to take full screen width
* [ ] Loading with skeleton ([refer to loading states design](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=5011%3A133032))

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202464660556485) by [Unito](https://www.unito.io)

Funnydev1031:
Hello. I confirmed my task.

singulart:
Assigned to @Funnydev1031

ivanturlakov:
✅ Tested on https://dao-git-staging-joystream.vercel.app/#/settings?network-config=https://34.230.5.182.nip.io/network/config.json



https://user-images.githubusercontent.com/46903215/204791555-d5b1b1a4-4bdc-4e98-8bfa-eceeb957cbbc.mov


⚠️ @dmtrjsg `text-overflow: ellipsis;` could be applied for long thread titles(as on member handle), if we have a time

<img width="1197" alt="Снимок экрана 2022-11-30 в 14 09 16" src="https://user-images.githubusercontent.com/46903215/204793153-635e9362-48d9-4d93-b395-2c8ee3683465.png">

Issue #3029, opened 2022-05-12
Forum UX Upd


dmtrjsg:
## Context
Forum during MVP dev phase was built around splitting the overall experience to tabulated pages and an overview landing page. The info architecture proved to be suboptimal and caused a number of change requests from the community members and JSG operations team.

Namely:

* #2429 
* #2426 
* #2951 
* #2952 
* #2437 
* #2428 
* #2438 
* #2734 
* #2785 

This ticket is aimed on surfacing all underlying use cases and optimising the forum UX experience for New users and Experienced users, including Forum WG Lead.

## Design
[Link to Figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=8966%3A319910)

## Scope
No qa required:
* [x] #3183
* [x] #3184
* [x] #3651
* [x] #3165

To-test:
* [x] #3151 
* [x] #3152
* [x] #3153
* [x] #3154
* [x] #3157 
* [x] #3155
* [x] #3162
* [x] #3224 
* [x] #3352
* [x] #2440 
* [x] #3163

Out of scope ⛔ :
* [x] #3187

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203142073969108) by [Unito](https://www.unito.io)

MattAugustynowicz:
Based on tests with several users, familiarization with the activities of other services having forums, our discussions, and knowledge, we come up with a proposal of forum solutions.   Let's start with what the user sees at the very beginning.

### The main page of the forum
[👉 Figma link here](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9364%3A333164)

<img width="1095" alt="image" src="https://user-images.githubusercontent.com/95480823/170501418-de2fbb03-ef2c-4f83-bbe3-33f36e388ee0.png">

[...]

PR #3562, opened 2022-08-22
🏛️ Implement new forum category page design

gyroflaw:
- #3153


gyroflaw:
> Hi @thesan 
For about this
SubCategoryListItem should be renamed and replace CategoryListitem
SubForumCategoryList should be renamed and replace ForumCategoryList
PopularThread should be renamed and replace ThreadItem
PopularThreadList should be renamed and replace ThreadBrowser
I agree with you about changes in PopularThread and PopularThreadList but for SubCategoryListItem and SubForumCategoryList, I think they need to exist since
image
This is ForumCategoryList
And
image
This is SubCategoryListItem 
So in my opinion they should be exist for different view of form category

singulart:
@thesan Some changes were committed. PR is ready to be reviewed once more. Thank you!

thesan:
@gyroflaw this is the category lists on the current forum landing page:
> image

But the new forum landing page is [@figma](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9364%3A333164):
image

The good news is AFAICS replacing `SubCategoryListItem` and `SubForumCategoryList` should also address:
- #3157

FYI here's the Forum redesign epic #3029.

Also @singulart can the builders group somehow access the designs on figma ?

gyroflaw:
Hi @thesan 
Thanks for your kind reply.
But I think this PR fixes this issue: https://github.com/Joystream/pioneer/issues/3153
This issue, describes updating on Forum Category page and the new landing page design would be raised at this issue: https://github.com/Joystream/pioneer/issues/3029
So I suggest this PR can be merged with current solution and for the landing page I think we can create another PR according #3029 issue.
What do you think?
@thesan  @singulart

thesan:
@gyroflaw you are correct that we should generally avoid addressing multiple issues in a single PR. But what I meant here was simply that the code of `CategoryListitem` and `ForumCategoryList` should be replaced by what is currently in `SubCategoryListItem` and `SubForumCategoryList`.

The reason is that we won't need the current version of `CategoryListitem` and `ForumCategoryList` anywhere on the app. So simply updating the existing components (and their stories) is the better option because it will keep the code simpler.

packages/ui/src/app/pages/Forum/ForumCategory.tsx
```diff
@@ -113,4 +128,8 @@ export const ForumCategory = () => {
 
 const ModeratorsContainer = styled(Label)`
   align-items: center;
+  flex-direction: column;
+`
+const Flex = styled(PageHeaderWrapper)`
+  flex-direction: column;
```
thesan, line 134: @gyroflaw here I meant that you should remove this `Flex` component and just add `flex-direction: column;` to the `PageHeaderWrapper` component directly.

gyroflaw:
> @gyroflaw you are correct that we should generally avoid addressing multiple issues in a single PR. But what I meant here was simply that the code of `CategoryListitem` and `ForumCategoryList` should be replaced by what is currently in `SubCategoryListItem` and `SubForumCategoryList`.
> 
> The reason is that we won't need the current version of `CategoryListitem` and `ForumCategoryList` anywhere on the app. So simply updating the existing components (and their stories) is the better option because it will keep the code simpler.

Okay, I will fix the codes and push soon.

packages/ui/src/app/pages/Forum/ForumCategory.tsx
gyroflaw: @thesan 
I think this might exist since `PageHeaderWrapper` is used in other components.
If we add flex-direction to the component directly, I think another used component's design would change.

gyroflaw:
@thesan @singulart Please review the changes

packages/ui/src/app/pages/Forum/ForumCategory.tsx
thesan: Yes `PageHeaderWrapper` is used in all pages but from looking through the app it looks to me like changing `flex-direction` to `column` shouldn't break anything.

gyroflaw: Okay, I will do. @thesan

gyroflaw:
Hi @singulart  @thesan 
Could you let me know why CI test failed?
I think the error is not related to current fix.

thesan:
@gyroflaw don't worry about it, it like some tests just took too long to run

packages/ui/src/forum/components/category/CategoryListItem.tsx
```diff
@@ -93,9 +75,12 @@ const CategoryListItemTitle = styled.h5`
 
 export const CategoryListItemStyles = styled(TableListItem)`
```
thesan, line 76: This component shouldn't extends `TableListItem` anymore because it's no longer the same kind of component (also it creates a slight misalignment of the first item). Instead maybe some style from `ThreadItemWrapper` could be factored out and reused here

thesan (changes requested):
@gyroflaw thank you for replacing the `SubCategoryListXXX` components and fixing the other changes requested.

I added a few more CRs, but please also replace:
- `PopularThread` by `ThreadItem`
- and `PopularThreadList` by `ThreadBrowser`

Finally there are a lot of unused references in these components now. To get a full list you can `cd` in `packages/ui` and run:
```sh
$(yarn bin eslint) "src/{forum/components,app/pages/Forum}/**/*.{ts,tsx}"
```
Please remove them on your last commit.

gyroflaw:
Thanks for your kind explanation. @thesan

gyroflaw:
@thesan , pls review the changes

packages/ui/src/forum/components/threads/ThreadBrowser.tsx
```diff
@@ -25,88 +17,23 @@ export interface ThreadBrowserProps {
...
-  totalCount,
-  isLoading,
-  currentPage,
-  setCurrentPage,
-  emptyText,
-}: ThreadBrowserProps) => {
-  const onPrevClick = () => setCurrentPage(currentPage - 1)
-  const onNextClick = () => setCurrentPage(currentPage + 1)
-
+export const ThreadBrowser = ({ label, threads, isLoading, emptyText }: ThreadBrowserProps) => {
```
thesan, line 20: thread browser
The new thread browser should also support pagination, therefore the properties: `pageCount`,` totalCount`, `currentPage`, and `setCurrentPage` are still needed.

Please try to reuse some of the previous `ThreadBrowser` logic to implement the new pagination.

packages/ui/src/forum/components/threads/ThreadItem.tsx
```diff
@@ -85,42 +73,30 @@ export const EmptyThreadItem = ({ text }: { text: string }) => (
...
   transition: ${Transitions.all};
 `
 
-const ThreadItemTime = styled(TextInlineExtraSmall)`
+export const ThreadItemTime = styled(TextInlineExtraSmall)`
   ${Overflow.FullDots};
 `
 
-const ThreadItemText = styled(TextMedium)`
+export const ThreadItemText = styled(TextMedium)`
```
thesan, line 93: `ThreadItemHeader`, `ThreadItemTitle`, `ThreadItemTime`, and `ThreadItemText` don't need to be exported anymore.

thesan (changes requested):
In addition to the following change requests please don't forget to address [this one](https://github.com/Joystream/pioneer/pull/3562/files#r958347168) too

gyroflaw:
Hi @thesan , thanks for your kind explanation, I think implementing pagination would take some time. I will push changes in a day. Thanks again.

gyroflaw:
Hi @thesan , I pushed the changes, please review.

packages/ui/src/forum/components/threads/ThreadItem.tsx
```diff
@@ -109,18 +97,6 @@ const ThreadItemText = styled(TextMedium)`
   overflow: hidden;
```
thesan, line 97: A lot more work is needed on the thread browser and it's items, but it should probably be done on follow up PRs. That said the items should at least make cap the post text to 4 lines. One way to do this would be to add:
```ts
DotsNLines: (n: number) => css`
  display: -webkit-box;
  -webkit-line-clamp: ${n};
  -webkit-box-orient: vertical;
  -moz-box-orient: vertical;
  overflow: hidden;
`,
```
to `Overflow` into `@/common/constants/styles`.

So this would become:
```suggestion
  ${Overflow.DotsNLines(4)};
```

thesan (changes requested):
Thank you for the changes I think this PR almost ready. What remains is:

1. A not yet addressed CR from an earlier review:
   > This component shouldn't extends TableListItem anymore because it's no longer the same kind of component (also it creates a slight misalignment of the first item). Instead maybe some style from ThreadItemWrapper could be factored out and reused here
2. Also the one about removing `category/LatestPost` and `category/PopularThread`
3. This review CR about keeping thread item text under 4 lines
4. Removing some of the storybook controls which are no longer in use. e.g "moderatorsCount", "latestPostText", etc... on the [CategoryListItem story](https://pioneer-2-storybook-git-fork-gyroflaw-fix-foru-8f3a22-joystream.vercel.app/?path=/story/forum-categories-categorylistitem--default) and some others on the [ThreadItem story](https://pioneer-2-storybook-git-fork-gyroflaw-fix-foru-8f3a22-joystream.vercel.app/?path=/story/forum-threads-threaditem--default)

singulart:
Re-requested review for https://github.com/gyroflaw/pioneer/commit/ffd1c9f8d24d8c4fc8c2c25148f53d139899e51b

packages/ui/src/forum/components/threads/ThreadItem.tsx
```diff
@@ -156,3 +130,25 @@ export const ThreadItemWrapper = styled.a<{ $halfSize?: boolean }>`
     }
   }
 `
+
+const StyledThreadItem = styled.a`
```
thesan, line 134: Also the goal of extracting `ThreadItemWrapper` is to have less code. So the common styles shouldn't be duplicated.

This looks looks good, so I think something this:
```ts
export const CardItem = styled.a`
  display: flex;
  flex-direction: column;
  gap: 16px;
  border: 1px solid ${Colors.Black[100]};
  border-radius: ${BorderRad.s};
  background-color: ${Colors.White};
  padding: 24px;
  &:hover,
  &:focus,
  &:focus-within {
    h5 {
      color: ${Colors.Blue[500]};
    }
  }
`
```
Should be extracted to a new file, e.g: `@/forum/components/CardItem`. And imported in CategoryListItem and ThreadItem, and if needed extended in there.

thesan (changes requested):

packages/ui/src/forum/components/threads/ThreadItem.tsx
```diff
@@ -147,7 +119,17 @@ export const ThreadItemWrapper = styled.a<{ $halfSize?: boolean }>`
   ${ThreadItemText} {
     -webkit-line-clamp: ${({ $halfSize }) => ($halfSize ? '3' : '14')};
   }
+`
 
+const StyledThreadItem = styled.a`
```
thesan, line 124: Again please use `CardItem` instead of this one if possible to have only one component here

thesan (changes requested):
About `ThreadItem.stories.tsx` the post text should show inside the component.

If you are having some issues running storybook locally (with `yarn storybook`) don't pay attention to the warnings it should start after a while.

thesan (approved):
@gyroflaw thank you for implementing all of the changes on this PR. Although a few improvement are needed the PR looks good for now.

Improvement for follow up PR:
- The thread slider should show part of the next tread and slide when the arrows are clicked: #3151
- The category items should display their sub category on hover: #3157

Issue #3533, opened 2022-08-19
Modals to disappear yet still block the modal context when they return `null`


dmtrjsg:
## Context
Was raised in epic
- #3522 

## Scope
- SSIA

PR #3567, opened 2022-08-24
🗑️ Fix for nullish modals

WRadoslaw:
- #3533


thesan (approved):
Great work :raised_hands: 
Using the states meta to store the common modals messages is really clever and it saved more code than I thought this refactor would

PR #3574, opened 2022-08-25
🧵 Fixed datetime is not displayed for Forum thread

### LINKED ISSUE #3347: Creation date and time is not displayed for a Forum Thread
Hello guyz:

I just noticed this possible improvement or addition in Joystream Pioneer.
If you create a new Forum Thread and open it, the date and time of creation of the thread is not shown.
Whereas if you reply on that thread, the date and time of reply thread is shown in the right side.

Please see picture on link below for easy illustration...
https://drive.google.com/file/d/1m9l8nQgA-pHohckhKUT8eqKM9Te-UxlG/view?usp=sharing

I suggest Joystream Pioneer be improved by adding creation date and time of a forum thread.
Possibly on upper right side of the thread same with reply.

Thank you

mkbeefcake:
Fixes #3347

### DIFF
--- a/packages/ui/src/common/types/Block.ts
+++ b/packages/ui/src/common/types/Block.ts
@@ -1,7 +1,7 @@
 import * as Types from '@/common/api/queries'
 import { isNumber } from '@/common/utils'
 
-export type NetworkType = 'BABYLON' | 'ALEXANDRIA' | 'ROME' | 'GIZA' | 'OLYMPIA'
+export type NetworkType = 'BABYLON' | 'ALEXANDRIA' | 'ROME' | 'GIZA' | 'OLYMPIA' | ''
 
 export interface Block {
   number: number
--- a/packages/ui/src/common/types/Block.ts
+++ b/packages/ui/src/common/types/Block.ts
@@ -1,7 +1,7 @@
 import * as Types from '@/common/api/queries'
 import { isNumber } from '@/common/utils'
 
-export type NetworkType = 'BABYLON' | 'ALEXANDRIA' | 'ROME' | 'GIZA' | 'OLYMPIA' | ''
+export type NetworkType = 'BABYLON' | 'ALEXANDRIA' | 'ROME' | 'GIZA' | 'OLYMPIA'
 
 export interface Block {
   number: number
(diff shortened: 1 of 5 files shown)


freakstatic:
@mkbeefcake thanks for the PR! :+1:

thesan (commented):
Thank you for the PR. This is a clever solution unfortunately the design requires to show the block number (plus we are planning to link these to the polkadot block explorer in the future). Therefore showing block 0 is not acceptable. However this information is available for example in the result of this graphqQl query:
```gql
query {
  forumPosts (where: { thread: { id_eq: "470" } }) {
    forumthreadinitialPost {
      createdInEvent {
        inBlock
      }
    }
    postaddedeventpost {
      inBlock
    }
    text
  }
}
```
Which correspond to this thread: https://dao-git-fork-mkbeefcake-fix-datetime-nodisp-fo-5cfa44-joystream.vercel.app/#/forum/thread/470

So by changing the fragment in `packages/ui/src/forum/queries/forum.graphql`

thesan (changes requested):
:point_up:

mkbeefcake:
Thanks for your detailed information. I will check around and fix this issue.

mkbeefcake:
issue_1
Hello thesan

I have one thing to discuss with you, I tried to check useForumThreadPosts() in packages/ui/src/forum/hooks/useFormThreadPosts.ts file and added the console.log() function for fetched response (postsResults.data?.forumPosts) via GraphQL.  Here as you can see in the screenshot, it returns "postaddeventpost' field as empty for the first POST item. and the this causes that problem.

so I can guess we have to change the code for adding 'postaddeventpost' field at the first post creation time in thread. also hope to know how to get the block number about this 'empty' field..

thesan:
Hello @mkbeefcake yes you are right that the current query node behaviour is not ideal, `postaddeventpost` shouldn't be empty on the thread post. However this can be worked around with the query I suggested yesterday. Here's the result from the playground:
thread-block

This way both the thread and its replies can be fetched in one query.

mkbeefcake:
Hello thesan

I updated the graphql and changed the source code. please review it again

thesan (approved):
Perfect ! Thank you for your work :pray:

Issue #3299, opened 2022-06-27
Storage of Avatar Images internally (as opposed to asking for URL) 


dmtrjsg:
## Context
Ensure less friction and better experience for all users, since avatar image via URL come with many problems. Atlas team is building a simple back end service

* https://github.com/Joystream/atlas/issues/2885

And corresponding adjustments to the onboarding flow 

* https://github.com/Joystream/atlas/issues/2884

## Scope
* [ ] Update the membership creation and editing flows in favour of uploaded avatars
* [ ] Update the modal to Edit Membership
* [ ] Existing [file uploader](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9707%3A378598)

## Info
![](https://user-images.githubusercontent.com/93924525/185081953-f9a7688f-cc88-471f-85b5-fbf0c704cb2b.png)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203143384618923) by [Unito](https://www.unito.io)

toiletgranny:
Alright, @dmtrjsg, designs for both the Create and Edit Membership dialogs can be found here: https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=10799%3A567889

Notes: 
1. Naturally, I started off trying to re-use existing patterns and components in Pioneer, but the only component for uploading files that I have found ("Dropzone") felt excessive when put in the context of (a) uploading a single image file to be used as an avatar, and (b) inside of a dialog and not a full-page experience.
2. Because of the above, instead of using a "Dropzone" component, I went for a classic [HTML file input type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file) (styled). 
3. To an engineer implementing this: make sure not to miss the error & loading states designs.

dmtrjsg:
@toiletgranny Agreed! 


I guess preview is also not needed here, I bet you thought about it already.. Perhaps its safe to remove the drop zone variant by this time as Choose file is perfect :)

dmtrjsg:
@toiletgranny can we pls add small preview in a circle pls for the avatar

toiletgranny:
If we can spend a bit more time on this, then sure! I updated the designs, and here they are — https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=10804%3A577502

dmtrjsg:
[...]

PR #3587, opened 2022-08-31
🧷 Member avatar service

WRadoslaw:
- #3299


packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -288,10 +304,30 @@ export const BuyMembershipForm = ({
...
+        setButtonText('Uploading avatar...')
+        const data = await uploadAvatarImage(fields.avatarUri).then((res) => res.json())
+        onSubmit({ ...fields, avatarUri: `${process.env.REACT_APP_AVATAR_UPLOAD_URL}/${data.fileName}` })
+      } else {
+        onSubmit(fields)
+      }
+    } catch (e) {
+      onSubmit(fields)
+    }
+  }
```
thesan, line 320: Again this logic is shared in a lot of places. WDYT of using a hook like:
```js
const useUploadAvatarAndSubmit = (onSubmit) => {
  const form = useFormContext()
  const [isUploading, setIsUploading] = useState(false)
  const uploadAvatarAndSubmit = useCallback(() => {
    setIsUploading(true)
    const avatarUri = await uploadAvatarImage(fields.avatarUri)
    onSubmit({ ...fields, avatarUri })
    setIsUploading(false)
  })
  return { isUploading, uploadAvatarAndSubmit }
}
```
With `uploadAvatarImage` changed too:
```ts
if (!avatarUri && avatarUri instanceof File) return avatarUri ?? ''
try {
  const body = new FormData()
  body.append('file', image, image.name)
  const data = fetch(process.env.REACT_APP_AVATAR_UPLOAD_URL ?? '', {
    method: 'POST',
    body,
  }).then((res) => res.json())
  return `${process.env.REACT_APP_AVATAR_UPLOAD_URL}/${data.fileName}`
} catch (err) {
  console.err('Error while uploading the avatar:', err)
  return ''
}
```

thesan (changes requested):

thesan (changes requested):

thesan (approved):
:raised_hands:

Issue #3569, opened 2022-08-25
Add loading spinner underneath account selector


dmtrjsg:
## Context

Unknown status of the staking account takes forever to get validated for accounts which are not bound to memberships:

![Screenshot 2022-08-25 at 09.53.54.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/c4f35941-be0a-46ac-9986-a598b21a939c)

## Scope
- [ ] Add a small loading spinner underneath the selector  with copy "Please wait, validation in progress."
- [ ] Remove the red highlight from the account component while validating
- [ ] Query the staking account status in the background

## Designs
Adapt the designs from  this ticket to this use case
#3508 

[Link to Figma for component >](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=10881%3A421078)

ivanturlakov:
Tested on https://dao-git-carthage-joystream.vercel.app/#/proposals

With a stable Internet connection, the validation process is very fast and the spinner does not show up. In order to see it, I disconnected from the network at a certain point. If the test doesn't fit, please let me know how to do it.


https://user-images.githubusercontent.com/46903215/191809563-7df1be24-5840-4352-9d63-9fb69ebe77ab.mov

dmtrjsg:
For the future this may help to test similar cases (like loading states for example which are in the pipeline)

<img width="553" alt="Screenshot 2022-09-22 at 20 46 21" src="https://user-images.githubusercontent.com/93924525/191816296-6b639ca7-edfe-4812-8fff-5adfd7e9c97c.png">

<img width="622" alt="Screenshot 2022-09-22 at 20 46 15" src="https://user-images.githubusercontent.com/93924525/191816302-c7f2d99a-f7b6-414a-b0aa-a62aa670fed4.png">

PR #3589, opened 2022-09-01
🚧 Improvement on staking account status UX

WRadoslaw:
- #3569


thesan (changes requested):
This one looks nice but it would be event better if it returned:
```js
<Loading text="Validating staking account" />
```

thesan (approved):
:+1:

Issue #3578, opened 2022-08-26
Test Invitation Lock coverage of Misc fees Pioneer


dmtrjsg:
## Context

2nd round of QA, Env details are in the Epic pls see 3577>
- https://github.com/Joystream/pioneer/issues/3577

Testing Invitation Lock Fees Coverage

## Scope
1️⃣ Ensure that 100% Invitation lock funds received from Faucet can only cover the following transactions:

###  ✅  Transactions Supported  Pioneer

   - Forum thread deposit (`forum::create_thread`)
   - Forum post deposit (`forum::create_thread`, `forum::add_post`)
   - **Staking**
    - Council candidate stake (`council::announce_candidacy`)
    - Staking account candidate stake (`membership::add_staking_account_candidate`)
    - Proposal stake (`proposals_codex::create_proposal`)
    - Referendum vote stake (`referendum::vote`)
    - Opening creation stake (`working_group::add_opening`)
    - Application/Worker stake (`working_group::apply_on_opening`, `working_group::increase_stake`)
    - Validator / nominator stake
    
2️⃣ Ensure that 100% Invitation lock funds received from Faucet cannot cover the following transactions:

###  ⛔ Not supported  Pioneer 
- **Transfers** (`balances::transfer`, `balances::transfer_keep_alive()`, `balances::transfer_all()` )
- **Budget funding** (`council::fund_council_budget`, `working_group::fund_working_group_budget`)
- **Membership fee** (`membership::buy_membership`, `membership::gift_membership`)

singulart:
Assigned to Ivan T#1098

ivanturlakov:
Tested on https://dao-git-carthage-joystream.vercel.app/#/settings?network-config=https://54.172.204.195.nip.io/network/config.json

Unexpected behavior is marked with an ❗️

### Transactions Supported Pioneer

✅ Forum thread deposit (forum::create_thread)
**Create category via extrinsic**
https://dao-git-carthage-joystream.vercel.app/#/forum/category/1
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2F54.172.204.195.nip.io%2Fws-rpc#/extrinsics/decode/0x170100504372656174652043617465676f72792054657374804372656174652043617465676f72792054657374204465736372697074696f6e
❗️ **Create thread via extrinsic (Only via extrinsics possible, UI throws - Insufficient Funds)**
https://dao-git-carthage-joystream.vercel.app/#/forum/thread/1
[...]

PR #3597, opened 2022-09-07
🦺 Improvement for forum actions with invitation lock

WRadoslaw:
- #3578


thesan (approved):
Looks great. Thank you for the quick PR

Issue #3523, opened 2022-08-18
Refactor Query subscriptions


thesan:
## Context
Since chain interactions are proxies through a web worker `useObservable` create extra unnecessary subscriptions because it passes an observable created on component rendering instead of a callback. Also some of this subscription should close after getting the first data (so an alternative hook should probably be used), instead it creates unnecessary re-renders.

## Scope
- Pass a callback to `useObservable`.
- Create another hook which relies on [`firstValueFrom`](https://rxjs.dev/api/index/function/firstValueFrom)

PR #3600, opened 2022-09-08
💳 Subscriptions refactor

WRadoslaw:
- #3523


packages/ui/src/common/hooks/useFirstObservableValue.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  const observable = useMemo(observableFactory, deps)
+
+  useEffect(() => {
+    if (observable) {
+      firstValueFrom(observable).then((value) => setData(value))
+    }
+  }, [observable])
+
+  return useMemo(() => data, [JSON.stringify(data)])
```
thesan, line 15: Since only one value will be emited, I don't think the `useMemo` is useful here:
```suggestion
  return data
```

packages/ui/src/accounts/hooks/useBalance.ts
```diff
@@ -11,16 +9,15 @@ import { Balances } from '../types'
...
 
 export const useBalance = (address: Address = ''): Balances | null => {
-  const { api, connectionState } = useApi()
+  const { api } = useApi()
 
   const allMyBalances = useMyBalances()
   const myBalances = allMyBalances?.[address]
 
-  const balancesObs = useMemo(
+  const balances = useObservable(
```
thesan, line 17: I think this one could use `useFirstObservableValue` or am I missing something ?

thesan (changes requested):
Nice work! I really like the `api?.isConnected` it makes things more readable :+1:

packages/ui/src/accounts/hooks/useBalance.ts
WRadoslaw: Yeah, you are right. I forgot that balances from wallet addresses are coming from BalancesContext which is using `useObservable`, therefor it will get every balance update (if someone transfers us some money). So in here there is no reason to observe it constantly as well, nice catch! :)

thesan (approved):
:raised_hands:

Issue #2823, opened 2022-04-08
Add link to Polkadot apps


bedeho:
⚠️ Validating if we need two separate links to validators / nominators.

### Scope
Add link  to 

* [x] onboarding flow 
* [x] settings
* [x] link on block numbers to specific number preview (linked #1993 )

Suggested Copy: "Some operations for nomination and validation are only supported via Polkadot.js App. Check Joystream test network explorer."

:information_source:  link to Joystream explorer will be pointing to : 

https://polkadot.js.org/apps/?rpc=wss://rpc.joystream.org:9944#/explorer
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frpc.joystream.org%3A9944#/explorer

### Context
- - -
We should have a link to a version on of Polkadot apps, which allows people to do all the validation and nomination stuff, as well as a block explorer, as requested here
https://discord.com/channels/811216481340751934/813361923172335648/961888072734244874

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202816092179171) by [Unito](https://www.unito.io)

MattAugustynowicz:
I propose that we could add our infoboxes components with the relevant information and a Polkadot icon

1️⃣ Onboarding Flow
[👉 Figma link here](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=5441%3A168962)

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/33006ec6-0527-461b-88c8-b6e98b007eb9)

2️⃣ Settings
[👉 Figma link here](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=8187%3A231780)

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/bcb79602-7fc5-4eda-be3b-52c023d5a8bf)

dmtrjsg:
Nice work! Only comment is for the overlay in the first screen to take the full width of the page, aligning the copy centrally.

dmtrjsg:
Agreed to extend the guest view explainer to the bottom and find an organic place for the info box

MattAugustynowicz:
According to the discussion and after having checked a few variants, let's stretch the box to its full width, while the texts are aligned to the left. I am also in favor of making the dark background stretch to the full height of the window :)

[...]

PR #3602, opened 2022-09-09
🗺️ Displaying link to polkadot explorer

WRadoslaw:
- #2823


packages/ui/src/app/components/SideBar.tsx
```diff
@@ -99,6 +104,17 @@ export const SideBar = () => {
...
+            <NavigationLinksItem>
+              <NavigationLink
+                icon={<PolkadotSymbol />}
+                onClick={() =>
+                  window.open(`https://polkadot.js.org/apps/?rpc=${endpoints.nodeRpcEndpoint}#/explorer`, '_blank')
+                }
+              >
+                Explorer
+                <LinkSymbol color={Colors.Black[500]} />
+              </NavigationLink>
```
thesan, line 116: It would be nicer to have use a hyperlink which shares the `NavigationLink` styles so users don't need to allow a popup

thesan (changes requested):

thesan (approved):
:raised_hands:

Issue #2948, opened 2022-04-30
Placeholder Screens for Accounts and Memberships for New users


dmtrjsg:
## Context

If wallet is not authorised, or  accounts are not created/ hidden or connected accounts have no membership for Joystream..

Accounts Page is blank and

Memberships Page has Lorem ipsum outdated message

<img width="1252" alt="Screenshot 2022-08-17 at 18 36 56" src="https://user-images.githubusercontent.com/93924525/185205695-38b021e0-b73e-4d5e-ab58-6830fce68ffd.png">

## Scope
- [ ] hide My profile side nav menu item if Account is not connected with the App
- [ ] Update the lorem ipsum copy on Memberhsip page when none of the connected accounts have associated memberships

"Membership is required for participation in the DAO activities and beneficial for accumulating the stats and reputation within the community."

#### Originating Ticket:
______________
- - -
**Taken at**: https://dao.joystream.org/#/profile/memberships
**Fields**:

* Houston, we have a problem.: Loading....
* Comment: There are many parts of the app that jus Load.... forever if you go to it the first time, without an extension or anything setup.
* Feedback: bedeho

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/574f7f5a-bd07-4c5d-b01c-4b24af22c34e-annotated_df45a6bd-8670-41e7-9919-9c33e6fe7e2c.png?etag=2b5c367202eeea2276b693cf48e38246)
_[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/574f7f5a-bd07-4c5d-b01c-4b24af22c34e-annotated_df45a6bd-8670-41e7-9919-9c33e6fe7e2c.png?etag=2b5c367202eeea2276b693cf48e38246)_

**Browser**: Chrome 100 (macOS Catalina)
**Screen size**: 5120x1440
**Browser size**: 1518x948
**[Open #231 in Usersnap](https://app.usersnap.com/l/feedback/d6ca21dd-b552-41bd-b0ad-b3e5d947e2f4?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

[...]

PR #3608, opened 2022-09-13
🛐 Empty accounts placeholder

WRadoslaw:
- #2948


thesan (changes requested):
The code looks good to me but on the my accounts page it would be nice not to have all the balance widgets when there are no accounts:
image

Which makes me think that (in another PR) we should probably also remove the vesting balance widgets for people with no vesting balances (which should be most users by far)

thesan (approved):
Thanks for taking of the vesting widget too :tophat:

Issue #3508, opened 2022-08-16
Remove Black Loading Modal


dmtrjsg:
## Context

In quite a few places this loading spinner is used:

![Screenshot 2022-08-16 at 20.01.37.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/84f62845-ead8-415a-a652-3d3f917d9cca)

## Scope
- [x] Locate the places where its shown now and see if a local spinner / "bad loading states' are actually better 
> - proposals list
> - new proposal
- [ ] Design treatment for removing this before loading states are done
- [ ] Change legacy "loading" to the updated spinner and align centrally plus add some padding

<img width="347" alt="Screenshot 2022-09-14 at 12 13 47 1" src="https://user-images.githubusercontent.com/93924525/190099221-c38797b9-7518-438d-8fbe-8428f202cb1d.png">

toiletgranny:
Having discussed it with Dima earlier today, @dmtrjsg, here are the designs for the agreed scope: https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=10881%3A419797

Please note, that for the "Creating new proposal" dialog, the idea is to open up the dialog imidietaly after pressing the "Add New Proposal" button, with the animated spinning wheel inside. Once the content is loaded, it should be displayed inside of the dialog, instead of the spinning wheel.

ivanturlakov:
Tested on https://dao-git-carthage-joystream.vercel.app/#/proposals
wss://44.201.112.198.nip.io/ws-rpc

### ✅ Proposals

https://user-images.githubusercontent.com/46903215/194294006-1c62356e-43af-44f2-a0c1-264a4cc8e584.mov

### ✅ Add New Proposal

https://user-images.githubusercontent.com/46903215/194295120-763bac6d-d12b-4711-a4f0-3377354c3a60.mov

✅ Profile
✅ Working Groups
✅ Council
✅ Election
✅ Forum
✅ Members

dmtrjsg:
⚠️  Proposals

@WRadoslaw on the proposals page there seems to be a residual loading spinner, that still shows after all of the data is loaded, can you pls have a look 👀

dmtrjsg:
The loader is waiting for the proposal Vote button in relevant casess, decided to remove this spinner and we will just show button if it applies. 👍

PR #3614, opened 2022-09-16
🏴 Removing black loading modal

WRadoslaw:
- #3508


dmtrjsg:
Agreed to replace this component with the updated spinner everywhere: https://pioneer-2-storybook-git-fix-kill-black-spinner-joystream.vercel.app/?path=/story/common-loading--loading-component

Also leave the opportunity to add the text if needed. For now will not needed anywhere apart from loading avatar on new member creation flow.

thesan (changes requested):
Nice refactor :raised_hands: 

I think the WaitModal, SearchProcess and their stories/icons should be removed too to simplify the codebase.

WRadoslaw:
@thesan As SearchProcess can be removed, WaitModal is still utilized on the Onboarding modal to indicate that faucet is processing the request (as it's a normal POST request, not a tx we don't use bottom tx modal). If we wanna remove it we need to find an alternative.

thesan (approved):
Right! It makes sense for the Onboarding modal

Issue #3621, opened 2022-09-21
Unable to cover tx fees with invitation lock


dmtrjsg:
Unable to pay for tx fees with invitation lock

![Screenshot 2022-09-21 at 16.21.08.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/815a03b8-f918-4238-93b6-54996342d27e)

![Screenshot 2022-09-21 at 16.21.04.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/b7dbcbff-1a1f-4fc7-848f-d8896fb7c903)

:warning:  Tested on wss://3.92.226.32.nip.io/ws-rpc

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203022629008140) by [Unito](https://www.unito.io)

dmtrjsg:
Unable to test on 54.172.154.45

thesan:
Detailed in #3626

PR #3623, opened 2022-09-22
🗡️ Fee check patch on transfer funds modal

WRadoslaw:
- #3621


thesan:
Based on Leszek comment shown on #3626 Pioneer shouldn't allow covering tx fees with invitation locks

Issue #3619, opened 2022-09-21
Show ETA on council candidate


dmtrjsg:
![Screenshot 2022-09-21 at 13.54.48.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/1640a134-f157-476e-8ee0-1d8d55eb7f06)

polikosi:
✅ETA is showing 
 
![image](https://user-images.githubusercontent.com/80429191/194148667-cfb1222f-52b1-4de4-a57e-39103fbe9d5d.png)

PR #3628, opened 2022-09-23
🇪🇹 Update on candidate lock ETA

WRadoslaw:
- #3619

### DIFF
--- a/packages/ui/src/accounts/components/AccountItem/components/lockItems/CouncilCandidateLockItem.tsx
+++ b/packages/ui/src/accounts/components/AccountItem/components/lockItems/CouncilCandidateLockItem.tsx
@@ -39,7 +39,7 @@ export const CouncilCandidateLockItem = ({ lock, address, isRecoverable }: LockD
 
   const recoveryTime = useMemo(() => {
     if (!voteStageDuration || !revealStageDuration || !remainingPeriod) {
-      return { unrecoverableLabel: 'Recoverable after failed election' }
+      return { unrecoverableLabel: 'Recoverable after lost election or end of the council term' }
     }
     const startTime = Date.now()
     const getBlocksToEnd = () => {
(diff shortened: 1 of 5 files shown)


packages/ui/src/accounts/components/AccountItem/components/lockItems/CouncilCandidateLockItem.tsx
```diff
@@ -25,26 +27,41 @@ export const CouncilCandidateLockItem = ({ lock, address, isRecoverable }: LockD
...
   const electionId = eventData?.candidate.electionRoundId
-  const electionStart = eventData?.candidate.electionRound.referendumStageVoting?.createdAt
   const voteStageDuration = api?.consts.referendum?.voteStageDuration.toNumber()
   const revealStageDuration = api?.consts.referendum?.revealStageDuration.toNumber()
 
   const recoveryTime = useMemo(() => {
-    if (!electionStart || !voteStageDuration || !revealStageDuration) {
-      return
+    if (!voteStageDuration || !revealStageDuration || !remainingPeriod) {
+      return { unrecoverableLabel: 'Recoverable after failed election' }
```
thesan, line 42: Shouldn't it be something like: "Recoverable after lost election or end of the council term" ?

thesan (changes requested):

packages/ui/src/accounts/components/AccountItem/components/lockItems/CouncilCandidateLockItem.tsx
WRadoslaw: Not really, council candidate lock can only be withdrawn after the lost election, if one wins the election then the lock transforms to a councilor lock which is handled by another list item.

This other list item for Councilor has the following tooltip:
image

dmtrjsg:
"Recoverable after lost election or end of the council term" 👍

thesan (approved):
Please retarget to `dev` once #3640 gets merged

Issue #3470, opened 2022-08-07
[USERSNAP] Forum: cannot search for numbers


dmtrjsg:
## Scope

- [ ] Add a message "Minimum of three characters is required" as a normal validation error.
- [ ] Fix the bug of entering 3 into the field, and the reducing input to 2 chars, the validation will pass, but needs to work

## Context
**Taken at**: https://dao.joystream.org/#/forum/category/16
**Fields**:
- Houston, we have a problem.: Forum: cannot search for numbers
- Comment: The following don&#39;t work:
- Searching for a number (i.e. 18)
- Searching for a date that is put in title fields (i.e. 13.07)

Both return no results and page doesn&#39;t seem to indicate it even attempted to search.
- Feedback: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/b3a0c971-5df1-475f-9caa-34b1125675aa-annotated_db8ca673-b88d-48b8-8411-90720e0f192d.png?etag=2bb28d880151a7e3e7a538282e75500e)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/b3a0c971-5df1-475f-9caa-34b1125675aa-annotated_db8ca673-b88d-48b8-8411-90720e0f192d.png?etag=2bb28d880151a7e3e7a538282e75500e)*

**Browser**: Chrome 103 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1640x1325
**[Open #323 in Usersnap](https://app.usersnap.com/l/feedback/964fc9f0-c118-44f1-9d45-5fbc83d83978?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203238414106362) by [Unito](https://www.unito.io)

dmtrjsg:
THe aggregation right now happens on the client and there is a limit of 500 results. Searching for 2 characters would not bring the good results, but we can remove this check.. 

Suggest to add a message "Minimum of three characters is required"

oleksanderkorn:
@dmtrjsg when and where should we show the validation message?

[...]

Issue #3369, opened 2022-07-19
Search Member by ID


dmtrjsg:
## Context

Follow up use case originating from this comment: 
- https://github.com/Joystream/pioneer/issues/2657#issuecomment-1173671702

## Scope

- [ ] Enable searching member by ID

traumschule:
duplicate of #3390 #2032 (implemented)

![member search](https://user-images.githubusercontent.com/31551045/197021700-d65abd16-0244-4ad0-9d95-e305e0b1f676.png)

PR #3635, opened 2022-09-27
Forum: cannot search for numbers #3470

oleksanderkorn:
https://github.com/Joystream/pioneer/issues/3470

### DIFF
--- a/packages/ui/src/common/components/Search/SearchResultsModal.tsx
+++ b/packages/ui/src/common/components/Search/SearchResultsModal.tsx
@@ -30,10 +30,12 @@ export const SearchResultsModal = () => {
   const { forum, forumPostCount, isLoading } = useSearch(search, activeTab)
   const isValid = () => !debouncedSearch || debouncedSearch.length === 0 || debouncedSearch.length > 2
   const debouncedSearch = useDebounce(search, 400)
-  const pattern = useMemo(
-    () => (isValid() && search ? RegExp(escapeStringRegexp(search), 'ig') : null),
-    [debouncedSearch]
-  )
+  const [pattern, setPattern] = useState<RegExp | null>(null)
+  useEffect(() => {
+    if (isValid()) {
+      setPattern(RegExp(escapeStringRegexp(debouncedSearch), 'ig'))
+    }
+  }, [debouncedSearch])
   const history = useHistory()
   const [hasOverlay, setHasOverlay] = useState(true)
   useEffect(
--- a/packages/ui/src/common/components/Search/SearchResultsModal.tsx
+++ b/packages/ui/src/common/components/Search/SearchResultsModal.tsx
@@ -19,16 +19,18 @@ import { ModalWithDataCall } from '@/common/providers/modal/types'
 import { ThreadItemBreadcrumbs } from '@/forum/components/threads/ThreadItemBreadcrumbs'
 import { ForumRoutes } from '@/forum/constant'
 
+import { useDebounce } from '../../hooks/useDebounce'
+
 export type SearchResultsModalCall = ModalWithDataCall<'SearchResults', { search: string }>
 
 export const SearchResultsModal = () => {
   const { hideModal, modalData } = useModal<SearchResultsModalCall>()
-
   const [search, setSearch] = useState(modalData.search)
   const [activeTab, setActiveTab] = useState<SearchKind>('FORUM')
   const { forum, forumPostCount, isLoading } = useSearch(search, activeTab)
   const pattern = useMemo(() => (search ? RegExp(escapeStringRegexp(search), 'ig') : null), [search])
-
+  const debouncedSearch = useDebounce(search, 300)
+  const isValid = () => !debouncedSearch || debouncedSearch.length === 0 || debouncedSearch.length > 2
   const history = useHistory()
   const [hasOverlay, setHasOverlay] = useState(true)
   useEffect(
@@ -55,7 +57,15 @@ export const SearchResultsModal = () => {
       <SearchResultsSidePane>
         <SearchResultsHeader>
           <CloseButton onClick={hideModal} />
-          <SearchInput>
+          <SearchInput
+            validation={isValid() ? undefined : 'invalid'}
+            validationStyles={{
+              position: 'absolute',
+              top: '60px',
+              left: '41px',
+            }}
+            message={isValid() ? '' : 'Minimum of 3 characters is required'}
+          >
             <InputText placeholder="Search" value={search} onChange={(event) => setSearch(event.target.value)} />
           </SearchInput>
         </SearchResultsHeader>
--- a/packages/ui/src/common/components/Search/SearchResultsModal.tsx
+++ b/packages/ui/src/common/components/Search/SearchResultsModal.tsx
@@ -32,7 +32,7 @@ export const SearchResultsModal = () => {
   const debouncedSearch = useDebounce(search, 400)
   const [pattern, setPattern] = useState<RegExp | null>(null)
   useEffect(() => {
-    if (isValid()) {
+    if (debouncedSearch && isValid()) {
       setPattern(RegExp(escapeStringRegexp(debouncedSearch), 'ig'))
     }
   }, [debouncedSearch])
--- a/packages/ui/src/common/components/Search/SearchResultsModal.tsx
+++ b/packages/ui/src/common/components/Search/SearchResultsModal.tsx
@@ -27,13 +27,14 @@ export const SearchResultsModal = () => {
   const { hideModal, modalData } = useModal<SearchResultsModalCall>()
   const [search, setSearch] = useState(modalData.search)
   const [activeTab, setActiveTab] = useState<SearchKind>('FORUM')
-  const { forum, forumPostCount, isLoading } = useSearch(search, activeTab)
   const isValid = () => !debouncedSearch || debouncedSearch.length === 0 || debouncedSearch.length > 2
   const debouncedSearch = useDebounce(search, 400)
-  const [pattern, setPattern] = useState<RegExp | null>(null)
+  const [validSearch, setLastValidSearch] = useState(debouncedSearch)
+  const { forum, forumPostCount, isLoading } = useSearch(validSearch, activeTab)
+  const pattern = useMemo(() => (validSearch ? RegExp(escapeStringRegexp(validSearch), 'ig') : null), [validSearch])
   useEffect(() => {
-    if (debouncedSearch && isValid()) {
-      setPattern(RegExp(escapeStringRegexp(debouncedSearch), 'ig'))
+    if (isValid()) {
+      setLastValidSearch(debouncedSearch)
     }
   }, [debouncedSearch])
   const history = useHistory()
--- a/packages/ui/src/common/components/Search/SearchResultsModal.tsx
+++ b/packages/ui/src/common/components/Search/SearchResultsModal.tsx
@@ -33,7 +33,7 @@ export const SearchResultsModal = () => {
   const { forum, forumPostCount, isLoading } = useSearch(validSearch, activeTab)
   const pattern = useMemo(() => (validSearch ? RegExp(escapeStringRegexp(validSearch), 'ig') : null), [validSearch])
   useEffect(() => {
-    if (isValid()) {
+    if (isValid() && debouncedSearch.length !== 0) {
       setLastValidSearch(debouncedSearch)
     }
   }, [debouncedSearch])
--- a/packages/ui/src/common/components/Search/SearchResultsModal.tsx
+++ b/packages/ui/src/common/components/Search/SearchResultsModal.tsx
@@ -29,7 +29,7 @@ export const SearchResultsModal = () => {
   const [activeTab, setActiveTab] = useState<SearchKind>('FORUM')
   const { forum, forumPostCount, isLoading } = useSearch(search, activeTab)
   const pattern = useMemo(() => (search ? RegExp(escapeStringRegexp(search), 'ig') : null), [search])
-  const debouncedSearch = useDebounce(search, 300)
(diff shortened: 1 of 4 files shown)


oleksanderkorn:
@dmtrjsg please take a look, also should we only limit this to forum search, or leave it for any other search pages like Members?

oleksanderkorn:
image
I have added padding in case of validation to avoid flickering, but on pages like proposals, it breaks the alignment if the button and search bar. Any suggestions on this? Should we leave the flickering of the main content or only add this validation on forum page? Or also add paddings for the buttons by alligning them to top:
image

packages/ui/src/common/components/Search/SearchResultsModal.tsx
```diff
@@ -19,16 +19,18 @@ import { ModalWithDataCall } from '@/common/providers/modal/types'
...
 export const SearchResultsModal = () => {
   const { hideModal, modalData } = useModal<SearchResultsModalCall>()
-
   const [search, setSearch] = useState(modalData.search)
   const [activeTab, setActiveTab] = useState<SearchKind>('FORUM')
   const { forum, forumPostCount, isLoading } = useSearch(search, activeTab)
   const pattern = useMemo(() => (search ? RegExp(escapeStringRegexp(search), 'ig') : null), [search])
-
+  const debouncedSearch = useDebounce(search, 400)
+  const isValid = () => !debouncedSearch || debouncedSearch.length === 0 || debouncedSearch.length > 2
```
thesan, line 33: Great idea to debounce the search, but the pattern should use only valid `debouncedSearch` too. I think this should work:
```ts
  const debouncedSearch = useDebounce(search, 400)
  const isValid = () => !debouncedSearch || debouncedSearch.length === 0 || debouncedSearch.length > 2
  const pattern = useMemo(() => (isValid && search ? RegExp(escapeStringRegexp(search), 'ig') : null), [debouncedSearch])
```

packages/ui/src/common/components/forms/FilterBox/FilterSearchBox.tsx
```diff
@@ -40,8 +41,10 @@ interface SearchBoxProps extends ControlProps<string> {
   displayReset?: boolean
 }
 export const SearchBox = React.memo(({ value, onApply, onChange, label, displayReset }: SearchBoxProps) => {
+  const debouncedValue = useDebounce(value, 400)
   const change = onChange && (({ target }: ChangeEvent<HTMLInputElement>) => onChange(target.value))
   const keyDown = onApply && (({ key }: React.KeyboardEvent) => key === 'Enter' && onApply())
+  const isValid = () => !debouncedValue || debouncedValue.length === 0 || debouncedValue.length > 2
```
thesan, line 47: Similarly here the sidebar should only open when value is valid:
```suggestion
  const isValid = () => !debouncedValue || debouncedValue.length === 0 || debouncedValue.length > 2
  const keyDown = !isValid || !onApply ? undefined : (({ key }: React.KeyboardEvent) => key === 'Enter' && onApply())
```

packages/ui/src/common/hooks/useSearch.ts
```diff
@@ -15,7 +15,7 @@ export const useSearch = (search: string, kind: SearchKind) => {
   const searchDebounced = useDebounce(search, 400)
 
   useEffect(() => {
-    if (searchDebounced.length > 2)
+    if (searchDebounced.length > 1)
```
thesan, line 18: I don't understand. Could you explain this change please ?

thesan (changes requested):

packages/ui/src/common/components/Search/SearchResultsModal.tsx
```diff
@@ -55,7 +57,15 @@ export const SearchResultsModal = () => {
...
         <SearchResultsHeader>
           <CloseButton onClick={hideModal} />
-          <SearchInput>
+          <SearchInput
+            validation={isValid() ? undefined : 'invalid'}
+            validationStyles={{
+              position: 'absolute',
+              top: '60px',
+              left: '41px',
+            }}
```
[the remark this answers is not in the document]
oleksanderkorn: Done!

packages/ui/src/common/components/forms/FilterBox/FilterSearchBox.tsx
```diff
@@ -40,8 +41,10 @@ interface SearchBoxProps extends ControlProps<string> {
   displayReset?: boolean
 }
 export const SearchBox = React.memo(({ value, onApply, onChange, label, displayReset }: SearchBoxProps) => {
+  const debouncedValue = useDebounce(value, 400)
   const change = onChange && (({ target }: ChangeEvent<HTMLInputElement>) => onChange(target.value))
-  const keyDown = onApply && (({ key }: React.KeyboardEvent) => key === 'Enter' && onApply())
+  const isValid = () => !debouncedValue || debouncedValue.length === 0 || debouncedValue.length > 2
+  const keyDown = !isValid || !onApply ? undefined : ({ key }: React.KeyboardEvent) => key === 'Enter' && onApply()
```
thesan, line 47: My bad! `isValid()` should be called here:
```suggestion
  const keyDown = !isValid() || !onApply ? undefined : ({ key }: React.KeyboardEvent) => key === 'Enter' && onApply()
```

packages/ui/src/common/components/Search/SearchResultsModal.tsx
```diff
@@ -19,16 +19,21 @@ import { ModalWithDataCall } from '@/common/providers/modal/types'
...
   const [activeTab, setActiveTab] = useState<SearchKind>('FORUM')
   const { forum, forumPostCount, isLoading } = useSearch(search, activeTab)
-  const pattern = useMemo(() => (search ? RegExp(escapeStringRegexp(search), 'ig') : null), [search])
-
+  const isValid = () => !debouncedSearch || debouncedSearch.length === 0 || debouncedSearch.length > 2
+  const debouncedSearch = useDebounce(search, 400)
+  const pattern = useMemo(
+    () => (isValid() && search ? RegExp(escapeStringRegexp(search), 'ig') : null),
+    [debouncedSearch]
+  )
```
thesan, line 36: Looks like this doesn't work on the preview branch: https://dao-git-fork-oleksanderkorn-3470-joystream.vercel.app/#/forum
I'm still able to search for 2 chars on the side pane (although it takes time since it's a big query)

@oleksanderkorn could you debug this please ?

thesan (changes requested):

thesan:
@oleksanderkorn I think you are referring to the "but needs to work" in:
>  Fix the bug of entering 3 into the field, and the reducing input to 2 chars, the validation will pass, but needs to work

I'm not sure what it's about, but we wrote this description together on a call and agreed that a 2 chars search shouldn't happen on this particular search box.

The reason for it is that the search aggregation is currently done client side (in `useSearch`) due to hydra limitation. As a result searching for 2 chars will almost always result in fetching the max amount of results (500 posts) regardless of their relevance (since even more posts will not make it in this 500 results just based only on their creation date). Thus it will waist bandwidth on the node and clients, as well as cpu load to aggregate the results which in my experience freezes the browser on lower spec computer.

dmtrjsg:
@oleksanderkorn the problem is that we are quite limited on bringing the relevant results before the relevancy is defined on the back end, which is underway. As a quick patch the decision was made to limit the min char input to 3 chars.. do you see a problem in this approach?  @oleksanderkorn

oleksanderkorn:
@dmtrjsg i can do it with 3 chars, but I understood that the issue was that tomato was not able to search by two chars, to use the date of month. If we so not want this to happen, then it's fine I will set it to three

dmtrjsg:
@oleksanderkorn I understand this is suboptimal but it is not very useful to have the relevancy missing and sorting of results done on FE for the 2 characters either. With this change we are trying to optimise the overall system performance without sacrificing much of the ability to find the info user wants.

RE months, would this convention be best to search for?
Screenshot 2022-10-17 at 12 12 15

If this does not work for the community, and its a real problem, then we could try and get better results, by prioritising the results on the title a bit more as a follow up issue. But searching by 2 brings too many results right now and we don't show the most relevant ones, and leave most of results "not displayed" so its misleading :(

dmtrjsg:
@oleksanderkorn also, searching for numbers in the format of 4 digits works :) 

Screenshot 2022-10-17 at 12 14 25

oleksanderkorn:
@thesan @dmtrjsg I am fine with the limit of 3 chars, I have applied the changes. I was only a bit confused based on the requireeements. Please check it again now.

packages/ui/src/common/components/forms/FilterBox/FilterSearchBox.tsx
oleksanderkorn: Done!

packages/ui/src/common/components/Search/SearchResultsModal.tsx
oleksanderkorn: I put the search limit back to 3 in `useSearch` should be ok now.

traumschule:
>  I was only a bit confused based on the requireeements.

There's also potential to confuse users when the validation errors and the search still works.

For proposals the request is prevented for "26" and curiously it's impossible to enter " 26" (with leading whitespace).
Searching for `proposalId` would be nice but out of scope i guess ,)

proposals-100

Forum: "18" won't open the drawer but once open it allows to search for even single digits.
Also here results for `threadId` or `postId` would be great (for some users) but we can follow up later.

forum-18
forum-13

packages/ui/src/common/components/Search/SearchResultsModal.tsx
thesan, line 36: Turns out setting `pattern` to null isn't enough because no pattern results in the entire post showing. Which is confusing because it looks like the result was updated despite the message saying that the search is too short.

[Screencast from 10-18-2022 05:33:12 PM.webm](https://user-images.githubusercontent.com/6571453/196480328-19ecbad5-99b0-4ae1-ba05-368bea29528b.webm)

Instead the `pattern` should remain the same:
```suggestion
  const [pattern, setPattern] = useState<RegExp | null>(null)
  useEffect(() => {
    if (isValid()) {
      setPattern(RegExp(escapeStringRegexp(debouncedSearch), 'ig'))
    }
  }, [debouncedSearch])
```

thesan (changes requested):

thesan:
@traumschule 

> Also here results for threadId or postId would be great (for some users) but we can follow up later.

Yes the current code is limited by hydra but if these features are necessary for the community, Pioneer could go around and assume that depending on the search format (e.g 2 digits number) check for ids, dates (and maybe thread titles) instead of a full text search on post body and thread title, which would return MBs of useless data.

packages/ui/src/common/components/Search/SearchResultsModal.tsx
```diff
@@ -19,16 +19,23 @@ import { ModalWithDataCall } from '@/common/providers/modal/types'
...
   const [search, setSearch] = useState(modalData.search)
   const [activeTab, setActiveTab] = useState<SearchKind>('FORUM')
   const { forum, forumPostCount, isLoading } = useSearch(search, activeTab)
-  const pattern = useMemo(() => (search ? RegExp(escapeStringRegexp(search), 'ig') : null), [search])
-
+  const isValid = () => !debouncedSearch || debouncedSearch.length === 0 || debouncedSearch.length > 2
+  const debouncedSearch = useDebounce(search, 400)
+  const [pattern, setPattern] = useState<RegExp | null>(null)
+  useEffect(() => {
+    if (isValid()) {
```
thesan, line 35: @oleksanderkorn sorry I had missed that if `debouncedSearch` is empty `isValid()` is `true`:
```suggestion
    if (debouncedSearch && isValid()) {
```

thesan (changes requested):

thesan:
@oleksanderkorn any idea what might be happening here ?

[Screencast from 10-18-2022 07:11:25 PM.webm](https://user-images.githubusercontent.com/6571453/196499332-3e65d290-f10b-461e-95ec-9dbc37f8cf77.webm)

The results remain the same but their order changes depending on the search

traumschule:
> The results remain the same but their order changes depending on the search

Is it ranking by some kind of relevance algorithm?

> Yes the current code is limited by hydra but if these features are necessary for the community,

On a 2nd thought it might be hard to guess correctly what a user wants to actually search for. If design permits there could be similar drop-down like on the members page (#3369) to choose between full-text search, proposal ID, post ID, thread ID, category ID.

select-member-id

dmtrjsg:
@traumschule it could make sense ofc, but is that how the search box would be used? So the user would know the particular proposal ID.. but following this logic if one puts a full Proposal ID string, I'm pretty sure this will be returned in the top result, doesn't it?

We think that giving the option to search for Post Title and Post Body may be a good one.. wdyt?

traumschule:
>  but following this logic if one puts a full Proposal ID string, I'm pretty sure this will be returned in the top result, doesn't it?

i tried this for proposal 100 and it didn't work.


In any case from my side it's just an additional out-of-scope idea.

oleksanderkorn:
@thesan I have applied as we discussed earlier, and it works as expected for me. Please review :)

thesan (approved):
Great work @oleksanderkorn :pray: 
Sorry for the delay I thought we had merged this one already

ivanturlakov:
Tested on https://dao-git-carthage-joystream.vercel.app/#/forum
wss://44.201.112.198.nip.io/ws-rpc

✅ Added a message "Minimum of three characters is required"
✅ Search by date and number works for description, ⚠️ but not for the title

https://user-images.githubusercontent.com/46903215/200565142-d3851867-8f26-4a8a-8baa-b8fd0818cf94.mov

Issue #3626, opened 2022-09-22
Transfer full amount


dmtrjsg:
## Context

![Screenshot 2022-09-22 at 16.49.54.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/eb49ace4-06e7-48bb-9044-dfd667e8bedf)

## Scope

- Update validation on transferring all tokens if manual input. Copy "Insufficient funds to cover tx fees, adjust the transferable amount."

Out of scope:
- Add MAX button for transafers (that populates the fees with full transferable balance - tx fees)

ivanturlakov:
Tested on https://dao-git-carthage-joystream.vercel.app/#/profile
44.201.112.198
3.92.226.32

<img width="1178" alt="Снимок экрана 2022-10-11 в 13 32 41" src="https://user-images.githubusercontent.com/46903215/195067766-b1d6be9e-524d-4ecb-8cf4-63fedbd41049.png">
<img width="913" alt="Снимок экрана 2022-10-11 в 13 33 56" src="https://user-images.githubusercontent.com/46903215/195068003-cdbe5532-4ded-4d7c-934d-dd94ffadd47b.png">
<img width="915" alt="Снимок экрана 2022-10-11 в 13 34 54" src="https://user-images.githubusercontent.com/46903215/195068170-4bc8d4d6-14bd-443c-bf55-c2f655899c33.png">
<img width="916" alt="Снимок экрана 2022-10-11 в 13 35 35" src="https://user-images.githubusercontent.com/46903215/195068326-2b0c9b38-3e06-4d39-94c0-65e73e431005.png">
<img width="912" alt="Снимок экрана 2022-10-11 в 13 36 22" src="https://user-images.githubusercontent.com/46903215/195068494-db61a6e9-848c-4436-bb02-c444ef7726c2.png">

PR #3638, opened 2022-09-28
🆕 New message on for all tokens transfer

WRadoslaw:
- #3626

### DIFF
--- a/packages/ui/src/accounts/modals/TransferModal/TransferFormModal.tsx
+++ b/packages/ui/src/accounts/modals/TransferModal/TransferFormModal.tsx
@@ -20,7 +20,7 @@ import {
 } from '@/common/components/Modal'
 import { BN_ZERO } from '@/common/constants'
 import { useForm } from '@/common/hooks/useForm'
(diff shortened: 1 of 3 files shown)


thesan (changes requested):
The validation message is longer than the JOY input. As a result the fields grows when it becomes invalid, which looks weird IMO. Could you increase the width of the JOY input to fix this ?

Screenshot from 2022-09-30 20-03-18
Screenshot from 2022-09-30 20-03-21

Also I think on this step we should anticipate the transaction fees value by reproducing a simplify version of the runtime logic. But since our value might be wrong, the message should be more of a warning than a blocking validation error (i.e people should be able to go to next step anyway). Then on the next step, with the knowledge of the actual transaction fee the message should appear to make it clear why the transaction can not proceed. WDYT ?

Alternatively to avoid spending too long on this, we could just show "Insufficient funds to cover tx fees" on the last step (also this should be a default copy on all sign-in modals), WDYT @dmtrjsg ?

thesan (approved):
:+1:

Issue #3152, opened 2022-06-02
Forum - My threads Tab


dmtrjsg:
## Context
UX facelift of Forum Module

## Designs Preview

![](https://user-images.githubusercontent.com/93924525/172781259-fc187779-7bb3-434a-b653-29894325773d.png)

[LInk to Figma> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9550%3A349072)

## Scope
**My threads Tab**

AC

* [ ] Section name My threads
* [ ] Display all threads as individual cards (basically filter by creator)
* [ ] Show threads counter in a pill on tab 
> MattAugustynowicz pls update the designs to include it
* [ ] Sorty by date starting with newest on top
* [ ] Show number of replies (same design as threads on homepage)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202464580928170) by [Unito](https://www.unito.io)

MattAugustynowicz:
[👉 Figma Link](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9550%3A349072)

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/0496a057-1791-4243-bc39-cfa3b8d55c70)

dmtrjsg:
My threads would be only posted by selected member, so also need to remove filtering :)

MattAugustynowicz:
@dmtrjsg You are right, for a moment I thought there would be threads from all our memberships :)

[👉 Figma link](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9550%3A349072)

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/11008029-039f-4d32-b91c-ba2d094ea19e)

secretgir1:
Hey @dmtrjsg 
Some questions to AC:
-- Could you please explain in more details what did you mean here:
" Show threads counter in a pill on tab"?
-- Also " Display all threads as individual cards (basically filter by creator)" - My Threads Tab will be displayed just my personal threads isn't it? Can't understand why should it be filtered by creator if we have just one creator here.

dmtrjsg:
Sure @secretgir1 

>Show threads counter in a pill on tab"?
We display the number of threads contained in that tab as a "pill" label with the counter:
[...]

PR #3643, opened 2022-09-29
Feature/my threads tab

WRadoslaw:
- #3152


thesan (changes requested):
On the category page threads are still supposed to appear as list items. Only popular threads show as cards in the slider:
https://user-images.githubusercontent.com/93924525/171611749-395475b8-274a-4e5b-a9ff-3a2339a66df3.png

But in https://dao-git-feature-my-threads-tab-joystream.vercel.app/#/forum/category/1 all threads are shown as a cards.

Or as there been some design changes ?

PR #3653, opened 2022-10-04
Feature stack: Forum landing tab

WRadoslaw:
⚠️ To be reviewed after #3649

Issue #3162, opened 2022-06-02
Forum: Watchlist Threads


dmtrjsg:
## Context
Forum UX update

## Design

[LInk to Figma> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9639%3A354609)

## Scope
* [ ] Watchlist tab
* [ ] Counter of threads in the tab as pill 
> MattAugustynowicz pls upd the designs to include the above
* [ ] Hover on thread card shows remove from watchlist
* [ ] Remove from watchlist action

:information_source:

thread card is described here - #3183

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202464606651656) by [Unito](https://www.unito.io)

MattAugustynowicz:
[👉 Figma Link](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9639%3A354609)

I think it would be better if this page were a separate tab. We have a simple filtering by which we could choose the author of the post we have in the watchlist, additionally like on the main page we can see the categories in pills and the number of replies on the tile. I think we can add on the hover state the possibility of immediate removal of a given post from the watchlist

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/d18539ab-2fb2-45fa-8168-13f3817af040)

dmtrjsg:
@MattAugustynowicz 

Re filters, on the watchlist I would not worry about it as the amount of threads on watchlist would not be overwhelming for average user. Sorting with latest on top  without possibility to change it I think should suffice. @MattAugustynowicz pls remove filters from here.

MattAugustynowicz:
Resolved ✅

👉 [Figma Link](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=9639%3A354609)

![image.png](https://images.zenhubusercontent.com/61f04aadb8abb13c0fc2e178/63fc5eeb-7ea7-42a2-b5c2-41f24d6e41da)

secretgir1:
Accepted. Ready for development.

dmtrjsg:
Duplicates and implemented in 
- https://github.com/joystream/pioneer/issues/3642

ivanturlakov:
[...]

PR #3654, opened 2022-10-04
Feature stack: Forum watchlist tab

WRadoslaw:
⚠️ Should be reviewed after #3653 

Implements #3162

### DIFF
--- a/packages/ui/src/forum/components/Thread/WatchlistButton.tsx
+++ b/packages/ui/src/forum/components/Thread/WatchlistButton.tsx
@@ -11,9 +11,18 @@ interface Props {
 
 export const WatchlistButton = ({ threadId }: Props) => {
   const [watchlist, setWatchlist] = useLocalStorage<string[]>(FORUM_WATCHLIST)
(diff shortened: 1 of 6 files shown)


packages/ui/src/forum/components/Thread/WatchlistButton.tsx
```diff
@@ -11,9 +11,18 @@ interface Props {
...
+      e.stopPropagation()
+      setWatchlist(watchlist ? [...watchlist, threadId] : [threadId])
+    },
+    [watchlist]
+  )
   const removeFromWatchlist = useCallback(
-    () => setWatchlist(watchlist?.filter((id) => id !== threadId) ?? []),
+    (e) => {
+      e.stopPropagation()
+      setWatchlist(watchlist?.filter((id) => id !== threadId) ?? [])
```
thesan, line 24: Once #3719 will have made its way into `dev`. To prevent users from loosing some of their watchlist by adding threads from different browser tabs, it would be good to change these into something like:
```ts
setWatchlist((watchlist = []) => [...watchlist, threadId])
setWatchlist((watchlist = []) => watchlist.filter((id) => id !== threadId))
```

thesan (approved):
:tada:

PR #3666, opened 2022-10-05
Feature stack: Forum - Archived tab

WRadoslaw:
Should be reviewed after #3654


packages/ui/src/forum/components/CategoryCard/CategoryCard.tsx
```diff
@@ -14,10 +14,10 @@ import { ForumCategory } from '@/forum/types'
 export interface CategoryCardProps {
   category: ForumCategory
   className?: string
+  archivedStyles?: boolean
```
thesan, line 17: Please add this to the story control.
Also `isArchive` should be removed from the ForumMain story

thesan (changes requested):

packages/ui/src/app/pages/Forum/ForumArchived.tsx
```diff
@@ -38,8 +40,23 @@ export const ForumArchived = () => {
...
-            {isLoadingCategories ? <Loading /> : forumCategories && forumCategories.length > 0 && <ForumMain />}
+            {isLoadingCategories ? (
+              <Loading />
+            ) : (
+              forumCategories &&
+              forumCategories.length > 0 && (
+                <RowGapBlock gap={10}>
+                  <TextMedium value lighter>
+                    Archived Categories
+                  </TextMedium>
```
thesan, line 51: Also I can't find the new design for this but shouldn't we keep the number of archived categories ? and still show:
image
when there are none ?

thesan (changes requested):

thesan (approved):
LGTM!
But I'm confused about the failing test

PR #3671, opened 2022-10-07
🔒 New hook to group account lock

WRadoslaw:


dmtrjsg:
@WRadoslaw to change cta on Application withdrawal modal from "Recover" to "Withdraw".

packages/ui/src/accounts/hooks/useGroupLocks/useGroupLocks.ts
```diff
@@ -0,0 +1,49 @@
...
+} from '@/accounts/hooks/useGroupLocks/helpers'
+import { BalanceLock } from '@/accounts/types'
+
+interface UseGroupLocks {
+  recoverable: BalanceLock[]
+  unRecoverable: BalanceLock[]
+}
+
+export const useGroupLocks = (address: string, locks: BalanceLock[]): UseGroupLocks => {
+  const { unstakingPeriodEnd, latestElection, voteElection, application, vote } =
```
thesan, line 15: Now I get why the logic was split between a hook and helper functions in the first place.

This is better now, but this could be avoided by having this structure:
```ts
const hasWGLock = locks.some(({ type }) => type.endsWith('Worker'))
const hasCandidateLock = locks.some(({ type }) => type === 'Council Candidate')
const hasVoteLock = locks.some(({ type }) => type === 'Voting')

const isWGLockRecoverable = useIsWGLockRecoverable(hasWGLock, adress)
const isCandidateLockRecoverable = useIsCandidateLockRecoverable(hasCandidateLock, address)
const isVoteLockRecoverable = useIsVoteLockRecoverable(hasVoteLock, address)

return locks.reduce( ...
```

This way the 1st parameter of `useIsXXXLockRecoverable` is a `!skip` which prevent the queries to be send for every accounts. And also all the logic for a given lock is in its specific hook.

packages/ui/src/accounts/components/AccountItem/components/LocksDetails.tsx
```diff
@@ -2,8 +2,7 @@ import React from 'react'
 
 import { LockItemWrapper } from '@/accounts/components/AccountItem/components/LockItemWrapper'
 import { VestingLockListItem } from '@/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem'
-import { useRecoveryConditions } from '@/accounts/hooks/useRecoveryConditions'
-import { isRecoverable } from '@/accounts/model/lockTypes'
+import { useGroupLocks } from '@/accounts/hooks/useGroupLocks/useGroupLocks'
```
thesan, line 5: For consistency with the project please add a `useGroupLocks/index.ts` so that:
```suggestion
import { useGroupLocks } from '@/accounts/hooks/useGroupLocks'
```

packages/ui/src/council/components/PastVotes/PastVote/PastVote.tsx
```diff
@@ -39,8 +41,9 @@ export const PastVote = ({ vote, latestCycleId, $colLayout }: PastVoteProps) =>
   }
 
   // Reflects if the vote was cast in latest election
-  const isLatestElection = vote.cycleId === latestCycleId
-  const isVoteStakeLocked = useIsVoteStakeLocked(vote.voteFor, { isLatestElection })
+  const isVoteStakeLocked = useGroupLocks(vote.castBy, balance?.locks ?? []).unRecoverable.some(
+    (lock) => lock.type === 'Voting'
+  )
```
thesan, line 46: Here I think a `useIsVoteLockRecoverable(hasVoteLock, vote.castBy)` should work if it's exported by: `@/accounts/hooks/useGroupLocks`

thesan (changes requested):
Nice PR there is already a lot of improvements on the current code, but I'm suggesting a few more:

packages/ui/src/working-groups/components/WithdrawApplicationButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const WithdrawApplicationButton = ({ application }: WithdrawApplicationButtonProps) => {
+  const { showModal } = useModal()
+  const stakingAccountBalance = useBalance(application.stakingAccount)
+  const { recoverable } = useGroupLocks(application.stakingAccount, stakingAccountBalance?.locks ?? [])
+
+  const withdrawApplication = useCallback(() => {
+    const lock = recoverable.find((lock) => lock.type.endsWith('Worker'))
+    if (lock) {
```
thesan, line 21: Similarly to `@/council/components/PastVotes/PastVote/PastVote` it would be nicer to have:
```suggestion
  const hasWGLock = stakingAccountBalance?.locks.some(({ type }) => type.endsWith('Worker'))
  const isWGLockRecoverable = useIsWGLockRecoverable(hasWGLock, address)
  
  const withdrawApplication = useCallback(() => {
    if (!isWGLockRecoverable) return
```

packages/ui/src/working-groups/hooks/useApplications.ts
```diff
@@ -18,10 +22,14 @@ export const ApplicationStatusToTypename: Record<ApplicationStatus, ApplicationS
   withdrawn: 'ApplicationStatusWithdrawn',
 }
 
-export const useApplications = ({ applicationsStatus, openingId }: Props) => {
+export const useApplications = ({ applicationsStatus, openingId, stakingAccount, limit, skip, orderBy }: Props) => {
   const { loading, data } = useGetWorkingGroupApplicationsQuery({
+    skip,
     variables: {
+      limit,
+      orderBy,
```
thesan, line 30: Here I don't think `orderBy` has to be a parameter, it's just nicer to make it explicitly `CreatedAtDesc` (IMO it makes just as much sense to order it this way in `SelectWorkingGroupApplicationBase`)
```suggestion
      orderBy: [WorkingGroupApplicationOrderByInput.CreatedAtDesc],
```

thesan (changes requested):

thesan (changes requested):
A few more changes, I think these should be the last ones:

thesan (approved):
Thanks for all of the changes :pray:

Issue #3679, opened 2022-10-10
Could not announce candidacy


dmtrjsg:
## Context
Attempting to test the candidacy stake shown in Election I encountered a strange bug (a few actually)

## Problem
* [ ] Transaction model showed insufficient funds, even thought there were plenty
* [ ] Binding account for staking returned error : Account for staking does not exist

![Screenshot 2022-10-10 at 17.06.48.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/61adaa53-a973-4538-984e-142324ca6b74)

## Steps to reproduce
Env: Carthage
Network: wss://44.201.112.198.nip.io/ws-rpc

Election min staking:
![Screenshot 2022-10-10 at 17.07.38.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/c3fc26ce-8c0e-4bb0-bdd2-aa507d95a5a4)

Attempted to stake 90,000

![Screenshot 2022-10-10 at 17.01.35.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/3325127a-a84d-4ad5-b505-d17cc1bcfacc)

Go through annoounce candidacy flow with BlossomHungry account
![Screenshot 2022-10-10 at 17.01.49.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/bca62aab-7d52-43ad-bcd8-5f6476c7909a)

TX: https://polkadot.js.org/apps/?rpc=wss://44.201.112.198.nip.io/ws-rpc#/explorer/query/0xc8215c55b0ff9b0a18a8626ba17b5a181dfbc6d50862fcdc3c544b8bb70240ec

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203134125618756) by [Unito](https://www.unito.io)

dmtrjsg:
Address: 5He8uJYkBRt9NktabgX89CKYQDveT8Ny4Va6DkqGBkW7yagt

PR #3683, opened 2022-10-11
❌ Fix validation on carthage

WRadoslaw:
- #3679


thesan (approved):
I don't understand why this transaction was failing but the new code looks a lot better so :+1:

WRadoslaw:
> I don't understand why this transaction was failing but the new code looks a lot better so 👍

The reason was that the deps array on `confirmStakingAccount` included `state.context` which after forms refactor doesn't have staking account in itself, it resulted in the staking account being an empty string. Changing `state.context` for `form.getValues()` or just `stakingAccount` would fix it, but I as well prefer this structure better.

thesan:
Got it! Thanks for taking the time to explain it :pray:

Issue #3664, opened 2022-10-05
Talisman signer does not popup


bedeho:
When I connect Talisman signer for the first time, one needs to enter the password and select the account, however, there is no popup that actually displays the wallet in this case, which will leave the user blocked, unable to make forward progress, and unaware that Talisman signer in a special signer authentication mode.

I am on Firefox Browser 105.0.1, OSX and Talisman v1.8.0.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203148958867444) by [Unito](https://www.unito.io)

dmtrjsg:
we couldn't reproduce, but found a case where membership does not get disconnected when wallet does.

To reproduce:
- start with signed in membership (like alice)
- Disconenect wallet (Polkadot)
- connect with Talisman which does not have alice account
- see how CTAs to sign are blocked


We think that you've caught one of these cases ☝️ 

## Scope
- Disconnect membership when wallet is disconenected

polikosi:
I think everything is working fine. What do yo think? WDYT @dmtrjsg



https://user-images.githubusercontent.com/80429191/200595131-dd473ffa-282b-4622-aa48-1bf1dae345c2.mp4

dmtrjsg:
Yep all g! thanks for testing @polikosi 👍

PR #3693, opened 2022-10-12
🗝️  Keeping account provider empty while no wallet

WRadoslaw:
- #3664


thesan (approved):
I don't really understand how but it solves the bug even the case where:
1. Select an active member A
2. Disconnect wallet
3. Hide controller account of A
4. Reconnect same wallet

And un-hiding the account even reconnect the account :100:

Issue #3657, opened 2022-10-05
Downscale images before uploading to Atlas Avatar Service


kdembler:
AFAIK, currently when user creates a membership in Pioneer, they are asked to provide a file for their avatar and that file is uploaded to Atlas Avatar Service without any modification. However, it would be beneficial to first downscale that image to some reasonable size to ensure images don't get too big. Let's use the same params as Atlas uses: 192x192, webp format.

For reference, here's an article that descirbed how this can be done with Canvas API: https://imagekit.io/blog/how-to-resize-image-in-javascript/. And also relevant Atlas code (although we use library for cropping images, the export is done with Canvas as well): https://github.com/Joystream/atlas/blob/master/packages/atlas/src/components/_overlays/ImageCropModal/cropper.ts#L176-L184

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203106355891279) by [Unito](https://www.unito.io)

ivanturlakov:
Tested on https://dao-git-dev-joystream.vercel.app/#/settings?network-config=https://54.163.55.37.nip.io/network/config.json

Avatar didn't show up after a member created


<img width="888" alt="Снимок экрана 2022-11-16 в 17 41 36" src="https://user-images.githubusercontent.com/46903215/202227070-8f884082-8f7f-47f0-9c8c-a6de8d32995e.png">

ivanturlakov:
✅ Tested on https://dao-git-dev-joystream.vercel.app/#/settings
wss://34.230.5.182.nip.io/ws-rpc

### Create a New Membership
<img width="879" alt="Снимок экрана 2022-11-28 в 15 15 11" src="https://user-images.githubusercontent.com/46903215/204287705-391de103-60f5-4f51-acdc-54010f88ebaa.png">
<img width="528" alt="Снимок экрана 2022-11-28 в 15 17 39" src="https://user-images.githubusercontent.com/46903215/204288175-76a79d9e-2731-45bb-b0c7-8aabe986b24c.png">

### Source: 7kb jpeg 460*460

### Saved:  2kb webp 192*192

Path: https://atlas-services.joystream.org/avatars/f9ff9628-80f4-4a2b-8989-a9ba31259c4d.webp

PR #3702, opened 2022-10-14
⚖️ Scaling down avatar before the upload

WRadoslaw:
- #3657

### DIFF
--- a/packages/ui/src/common/components/SmallFileUpload/SmallFileUpload.tsx
+++ b/packages/ui/src/common/components/SmallFileUpload/SmallFileUpload.tsx
@@ -9,6 +9,7 @@ import { ColumnGapBlock, RowGapBlock } from '@/common/components/page/PageConten
 import { TextMedium } from '@/common/components/typography'
 import { BorderRad } from '@/common/constants'
 import { resizeImageFile } from '@/common/helpers'
+import { info } from '@/common/logger'
 import { enhancedGetErrorMessage, enhancedHasError } from '@/common/utils/validation'
 import { Avatar } from '@/memberships/components/Avatar'
 import { SUPPORTED_IMAGES } from '@/memberships/model/validation'
@@ -27,7 +28,7 @@ export const SmallFileUpload = ({ onUpload, name, initialPreview }: SmallFileUpl
   const { formState } = useFormContext()
   const [avatarPreview, setAvatarPreview] = useState<string>(initialPreview ?? '')
   useEffect(() => {
-    if (localValue.blob && SUPPORTED_IMAGES?.includes(localValue.file?.type ?? '')) {
+    if (localValue.blob && SUPPORTED_IMAGES.includes(localValue.file?.type ?? '')) {
       const objectUrl = URL.createObjectURL(localValue.blob)
       setAvatarPreview(objectUrl)
 
@@ -41,17 +42,18 @@ export const SmallFileUpload = ({ onUpload, name, initialPreview }: SmallFileUpl
     (event: React.ChangeEvent<HTMLInputElement>) => {
       const file = event.target.files?.item(0) ?? null
 
-      if (file) {
+      if (file && SUPPORTED_IMAGES.includes(file.type)) {
         try {
           resizeImageFile(file, 192, 192, 'image/webp').then((blob) => {
             setLocalValue({ blob, file })
-            onUpload(event, blob)
+            return onUpload(event, blob)
           })
         } catch (e) {
-          setLocalValue({ blob: null, file })
-          onUpload(event, file)
+          info(e)
         }
       }
+      setLocalValue({ blob: null, file })
+      onUpload(event, file)
     },
     [onUpload]
   )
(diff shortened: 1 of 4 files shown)


packages/ui/src/common/components/SmallFileUpload/SmallFileUpload.tsx
```diff
@@ -22,12 +23,13 @@ export const SmallFileUpload = ({ onUpload, name, initialPreview }: SmallFileUpl
...
-
   useEffect(() => {
     if (localValue && SUPPORTED_IMAGES?.includes(localValue.type)) {
-      const objectUrl = URL.createObjectURL(localValue)
-      setAvatarPreview(objectUrl)
-      return () => URL.revokeObjectURL(objectUrl)
+      resizeImageFile(localValue, 192, 192, 'image/webp').then((blob) => {
+        const objectUrl = URL.createObjectURL(blob)
+        setAvatarPreview(objectUrl)
+        return () => URL.revokeObjectURL(objectUrl)
```
WRadoslaw, line 31: @thesan Here I'm not sure whether cleanup will execute from a callback 🤔

packages/ui/src/common/helpers.ts
```diff
@@ -53,3 +55,28 @@ export const subtitleMapping = (value: string) => {
...
+    const reader = new FileReader()
+    reader.onload = (event) => {
+      const img = new Image()
+      img.onload = () => {
+        const canvas = document.createElement('canvas')
+        canvas.height = height
+        canvas.width = width
+        const ctx = canvas.getContext('2d')
+
+        ctx?.drawImage(img, 0, 0, width, height)
```
thesan, line 75: Currently the image are ratio is modified. After some trial and error on: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_canvas_drawimage
It looks like this should keep ratio (and center on the result image):
```suggestion
        const dW = img.width - width
        const dH = img.height - height
        const [clippedWidth, clippedHeight] = Math.abs(dW) > Math.abs(dH)
          ? [Math.floor(img.width / img.height * width), img.height]
          : [img.width, Math.floor(img.height / img.width * height)]

        const x = Math.floor(img.width/2 - clippedWidth/2)
        const y = Math.floor(img.height/2 - clippedHeight/2)

        ctx.drawImage(img, x, y, clippedWidth, clippedHeight, 0, 0, width, height);
```

packages/ui/src/memberships/components/AvatarInput.tsx
```diff
@@ -1,25 +1,35 @@
...
+        try {
+          const fileBlob = await resizeImageFile(file, 192, 192, 'image/webp')
+          formContext.setValue('avatarUri', fileBlob, { shouldValidate: true })
+        } catch (e) {
+          formContext.setValue('avatarUri', file, { shouldValidate: true })
+        }
+      }
+    },
+    [formContext.setValue]
+  )
```
thesan, line 25: Since the image is already resized inside SmallFileUpload why not just calling `onUpload(avatarPreview)` (maybe `avatarPreview` should be renamed to `localValue` and `localValue` to `file`).

So this would just be:
```suggestion
    (value) => formContext.setValue('avatarUri', value, { shouldValidate: true })
```

WDYT ?

thesan (changes requested):

packages/ui/src/common/helpers.ts
WRadoslaw: This basically depends on the image, bigger ones with context not in the center will seem meaningless when we keep the ratio and clip it.

packages/ui/src/memberships/components/AvatarInput.tsx
thesan: Right I forgot that `avatarPreview` was a url. But yes It would make more sense to me to resize the image just once and have the blob stored inside the `SmallFileUpload`.

Maybe `onUpload` could become:
```ts
(event: React.ChangeEvent<HTMLInputElement>, resizedImage?: Blob) => void
```

packages/ui/src/common/helpers.ts
thesan: Yes but IMO (once one dimension of the image matches the target) it's preferable to crop it at an arbitrary position than deform it. Especially when the original ratio is kinda close to the targeted one, it makes me feel like something is not right :thinking: 

Also here the cropping happens in one dimension only.

WRadoslaw:
@thesan I've applied the changes, lemme know WDYT

packages/ui/src/common/components/SmallFileUpload/SmallFileUpload.tsx
```diff
@@ -8,29 +8,54 @@ import { Row } from '@/common/components/Modal'
...
+        try {
+          resizeImageFile(file, 192, 192, 'image/webp').then((blob) => {
+            setLocalValue({ blob, file })
+            onUpload(event, blob)
+          })
+        } catch (e) {
+          setLocalValue({ blob: null, file })
+          onUpload(event, file)
+        }
+      }
```
thesan, line 54: Could this fix the validation ?
```suggestion
      if (file && SUPPORTED_IMAGES?.includes(file.type)) {
        try {
          resizeImageFile(file, 192, 192, 'image/webp').then((blob) => {
            setLocalValue({ blob, file })
            return onUpload(event, blob)
          })
        } catch (e) {
        }
      }
      setLocalValue({ blob: null, file })
      onUpload(event, file)
```

thesan (changes requested):
The logic looks good but when I try to upload an svg it fails quietly without the "This file type is not allowed"

thesan (approved):
:+1:

Issue #3316, opened 2022-07-01
Deleted threads shows up in multiple places


singulart:
Reproduce: 
1. Create thread
2. Delete thread

Expected result:
1. Deleted thread cannot be opened by ID
2. Deleted thread is not showing anywhere in Forum

Actual result:
1. Deleted thread is visible in `MY THREADS`
2. It is possible to click and view the deleted thread. 
3. It is possible to add Deleted thread to Watchlist
4. Watchlisted deleted thread is visible in `YOUR WATCHLIST`
5. Post from deleted thread is visible in 'LAST POST' column in the main Forum view

singulart:
Assigned to newjoiner The Muscle#4068

chrlschwb:
This seems to have been fixed by jsgenesis. Should go to QA

ivanturlakov:
Tested on https://dao-git-dev-joystream.vercel.app/#/forum/thread/40
wss://34.230.5.182.nip.io/ws-rpc

⚠️ No chance to delete my thread

No chance to deleting my post
<img width="1212" alt="Снимок экрана 2023-01-30 в 16 20 17" src="https://user-images.githubusercontent.com/46903215/215502754-633f77d0-5090-439f-9d91-3b299509c47e.png">
<img width="1213" alt="Снимок экрана 2023-01-30 в 16 20 35" src="https://user-images.githubusercontent.com/46903215/215502760-cd7d44d9-988f-4d51-87ad-255ecb80eb02.png">

PR #3715, opened 2022-10-19
issue #3316 fixed

chrlschwb:
It was a problem of Thread Status.
I have changed the "ThreadStatusModerated" to "ThreadStatusRemoved" in the hooks.

### DIFF
--- a/packages/ui/src/app/pages/Forum/ForumThread.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumThread.tsx
@@ -44,6 +44,7 @@ export const ForumThread = () => {
   }, [replyTo])
 
   const isThreadActive = !!(thread && thread.status.__typename === 'ThreadStatusActive')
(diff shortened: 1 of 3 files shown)


traumschule:
Thanks @chrlschwb! Looks like this would fix #3316 but I wonder if it also unhides threads flagged `ThreadStatusModerated`?

chrlschwb:
Hello.
Yes, the threads ThreadStatusModerated would unhide.
Let me know if it needs fixing.

traumschule (changes requested):
Yes, both ThreadStatusRemoved and ThreadStatusModerated need to be hidden.

traumschule:
This is obsolete, see https://github.com/Joystream/pioneer/issues/3153#issuecomment-1294003934

chrlschwb:
Hello, l1dev.
What is this message for?

traumschule:
It means JSG will rework the forum on their own for now.

Issue #3711, opened 2022-10-18
Revealing issue - unable to reveal more than 1 candidate on network 54.*


dmtrjsg:
## Context
During the QA test for:
* https://github.com/Joystream/pioneer/issues/3669#issuecomment-1281229788

On `54.227.116.57.nip.io` ( election `cycle_id: 713`) @ivanturlakov:
> 3 council seats
> 
> Trying to hold an election.
> For each candidate I vote from a new account (except the 4th).
> 1 - 1000
> 2 - 2000
> 3 - 3000
> 4 - 4000 (he votes for himself)
> 
> The stage of revealing the votes.
> There is an opportunity to reveal only the vote for the 3rd candidate.

On the QN only the 3rd vote was revealed:
```json
[
  {
    "votePower": "10000000000000",
    "castBy": "j4UGeAgkw76FXWmP34ZijA6rTiJaSqXHjEux6i3e7EvnmTyi2",
    "commitment": "0x30bddf42662e1db46e473d6a8a6253d630b261c716aa3ce88ffd485b47fd5cf9",
    "votecasteventcastVote": [ { "inBlock": 143409 } ],
    "voteFor": null
  },
  {
    "votePower": "20000000000000",
    "castBy": "j4SEEdc9wnZfVZ3mwcamS7ojmLBcLtYU5dopkXfRsi1285HuQ",
    "commitment": "0x803d040485c09c7c8ebe6405f441305fd24e0bafc8d96d35be95f192f223c120",
    "votecasteventcastVote": [ { "inBlock": 143417 } ],
    "voteFor": null
  },
  {
    "votePower": "30000000000000",
    "castBy": "j4SZd51TYDJLmcFvCq938CiseRGbqSyezXDWkPHPDbmBDw1oC",
    "commitment": "0x4e7c824d33bff598eb60a198c707aded587ed73235f76b65e501949e83017997",
    "votecasteventcastVote": [ { "inBlock": 143429 } ],
    "voteFor": { "memberId": "19" }
  },
  {
    "votePower": "40000000000000",
    "castBy": "j4Wtt8U69mpSiQmMSWrbU7bkcPnJqQeJjeV3Yh5XKtWMeDjqb",
    "commitment": "0x01d4d1e2353865558c73c507b310b92827b7d1884d938e52f31e13d7174de7bb",
    "votecasteventcastVote": [ { "inBlock": 143364 } ],
    "voteFor": null
  }
]
```

[...]

PR #3719, opened 2022-10-20
🗃️ Fix votes storage

thesan:
- #3711 

I'm guessing that in the issue multiple tabs were opened so the local storage item kept on getting overridden (that's the source of most local storage issue). The empty `accountId` entry was due the commitment first getting calculated before the staking account was set.


WRadoslaw (approved):
🥳

Issue #3583, opened 2022-08-29
Update Gateways WG to be called Apps


dmtrjsg:
## Context
- https://github.com/Joystream/joystream/issues/4230

## Scope

- [x] Update WG name from "Gateways" to "Apps"
- [x] Update description to "Apps group runs multiple video streaming apps working on Joystream blockchain and provides support to all exteernal app operators."
- [ ] Update Icon
- [x] Add Tooltip. "Learn more about this group [[external source icon]](https://joystream.gitbook.io/testnet-workspace/system/gateways)." 

 description and icons, same with proposal selection of WG.

toiletgranny:
@dmtrjsg a few quick clarifying questions:
1. What is/are the Pioneer page(s)/view(s) this affects?
2. Where should the name and description be updated?
3. What is the current icon for Gateways WG? In Figma, there are multiple icons with "Gateway" in their names; I don't know which we're talking about.
4. By "[external source icon]", did you mean any specific icon that's already used in Pioneer? Where should this tooltip be added?

dmtrjsg:
Hey Adam, 

Sure so 

Here's the url to the page https://dao-git-carthage-joystream.vercel.app/#/working-groups

Screenshot of the current state:
<img width="1489" alt="Screenshot 2022-10-18 at 11 19 33" src="https://user-images.githubusercontent.com/93924525/196363025-0d8247f5-b71c-4fbb-9050-bce4b2ea547b.png">

Link to tooltips issue for working groups:
https://github.com/Joystream/pioneer/issues/2680

toiletgranny:
Ok, @dmtrjsg, this is ready to be reviewed. Design in Figma [can be found here](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11275%3A434605). A few notes:

1. I put the "Learn more" link at the end of the description, as putting interactive elements in tooltips is typically considered a bad UX practice. Hope this recommendation is okay.
2. It appears there's a character limit on each WG description, which is implied and inherited by the layout and the way each Working group row has been designed and implemented. For this reason, I had to cut down the description you provided by just a touch so that it's contained within max ~105 characters. I hope the sorter version is still technically correct: "Apps group runs multiple video streaming apps on Joystream blockchain and supports all external operators."

![image](https://user-images.githubusercontent.com/35307309/196412982-3c8f5021-e3bf-4175-9427-338fea47010b.png)

[...]

PR #3720, opened 2022-10-20
🇩🇬 Rename gateways group

WRadoslaw:
- #3583


thesan (changes requested):
Please also edit the `defaultGroups` in [@/app/pages/WorkingGroups/WorkingGroups](https://github.com/Joystream/pioneer/blob/91756b75233c955bab010f67a4705c5ecf5bb587/packages/ui/src/app/pages/WorkingGroups/WorkingGroups.tsx#L22). It's used to show WG info before the QN response. Currently it shows:
Screenshot from 2022-10-27 17-55-53

thesan (changes requested):

packages/ui/src/app/pages/WorkingGroups/WorkingGroups.tsx
```diff
@@ -11,10 +12,26 @@ import { WorkingGroupsTabs } from './components/WorkingGroupsTabs'
 export const WorkingGroups = () => {
   const { groups } = useWorkingGroups()
 
+  const sortedGroups = useMemo(
+    () =>
+      !groups
```
thesan, line 17: My bad it should still be:
```suggestion
      !groups.length
```

thesan (changes requested):

packages/ui/src/app/pages/WorkingGroups/WorkingGroups.tsx
WRadoslaw: I've remembered to change it in deps array, but forgot to change it here :D

thesan (approved):
:100:

Issue #3723, opened 2022-10-20
After X amount of time failing to establish the `stateSubscription` end the tx flow successfully


thesan:
Context in #3722

PR #3731, opened 2022-10-21
🧷 Subscription timeout

WRadoslaw:
- #3723


packages/ui/src/common/hooks/useQueryNodeTransactionStatus.ts
```diff
@@ -20,6 +26,16 @@ export function useQueryNodeTransactionStatus(blockHash?: Hash | string, shouldS
...
 
+  useEffect(() => {
+    if (loading || queryNodeStateError) {
+      const timeout = setTimeout(() => {
+        setStatus('confirmed')
+      }, 10_000)
+
+      return () => clearTimeout(timeout)
+    }
+  }, [loading, queryNodeStateError])
```
thesan, line 37: IMO `queryNodeStateError` shouldn't reset the timeout:
```suggestion
  useEffect(() => {
    if (loading) {
      const timeout = setTimeout(() => {
        setStatus('confirmed')
      }, 10_000)

      return () => clearTimeout(timeout)
    }
  }, [loading])
```

thesan (changes requested):

thesan (approved):
:100:

PR #3741, opened 2022-10-23
Restore `Use Max` button on `TransferFormModal` (#1816)

### LINKED ISSUE #1816: Button to handle max balance transaction
On My Profile page, we have option to transfer tokens from one account to another. Goal is to implement button on TransferModal  that let user transfer whole balance with "Use Max" label. That feature was implemented before, but logic didn't allow to transfer exactly whole balance. With help of @jodator, possible (and desirable) solution is to implement it with transaction that do the calculation of amount plus fee that covers whole account balance.

traumschule:
Fixes #1816


traumschule:
This failed preliminary qa, thanks @ivanturlakov for the quick response! 
Insufficient funds to cover the transaction
`useTransactionFee` subtracted a fee of 0.07 tJOY but not enough to cover the tx. This could also be a validation error.
Possible remedy: always subtract existential deposit (~10 tJOY) to also prevent reaping of the account. WDYT?

thesan:
@traumschule I don't think it's a validation error:
image
Here nothing was subtracted from the transferable balance. Which makes sense because when `setMax` gets called `maxAmount` is `0` in most case or at least a lower than the value it should be.

The problem is that even if the value in the field was updated correctly, the transaction fee `A` for transferring the whole `transferableBalance` are higher than the transaction fee `B` for transferring `transferableBalance - A`. As a result the transaction will work but the user will be left with `A - B` as their new transferable balance. A way to find out the the real "MAX" amount is to reproduce the chain logic in an equation (and check the result once with query).

So this feature is tricky to implement (especially since AFAIK the chain tx fee logic might still be tweaked until mainnet) and is not a really high prio IMO. So I think it would be better to work on this later.

WDYT ?

traumschule:
Time for another test round. Can you still spot the validation error?

sendAll
signGood

oleksanderkorn (approved):
Looks good, works as expected for me: 
image

PR #3746, opened 2022-10-24
Add handle / name tooltips on `BuyMembershipFormModal` (#3024)

### LINKED ISSUE #3024: Tooltip: Handle vs Name improvements
## Scope

Add tooltips to create membership flow and new member flow. 

- [ ] Handle
This is your username that will be displayed in most of the places where you perform actions that others see, such as forum posts, present yourself as council candidate or create a new proposal. Think of it as a public name. Handles are unique.

- [ ] Name
Name: This is your name, which is separate to your handle. It will only be displayed on the membership details pages. SomeDAO participants choose to use their real names, while others choose to use their nicknames. Name does not have to be unique.

## Context
This language seems to be causing a lot of questions from the community, it would be good to have them include an example explaining the difference.

traumschule:
Fixes #3024


thesan (approved):
:+1:

PR #3758, opened 2022-10-25
👾 Sort proposal discussion by block (#3236)

### LINKED ISSUE #3236: Incorrect display of comments in a proposal
## Scope

- [ ] Use block count for sorting

 ⚠️ if this fails then the issue is with Hydra - post the results of attempt to comments
__________

During the discussion it was possible to find out that the proposal does not correctly display more than 15 comments - https://dao.joystream.org/#/proposals/preview/274

The 16th comment moved up the discussion. You can see from the time in the screenshot that it is located incorrectly
image

traumschule:
Fixes #3236


oleksanderkorn (approved):

oleksanderkorn:
@traumschule lgtm, test is flaky so rerstarting the test should make it pass i hope, but i can't restart it, do not have permissions.

Issue #3273, opened 2022-06-22
[feature request] Forum: tooltip to show full thread title


dmtrjsg:
## Scope

 - [ ] Truncate thread title after 2 lines 
 - [ ] Show full thread name on hover over on title in the tooltip
 
## Guidance

 > "Radosław" add this line of code to the title component for truncation:
truncateLines={2}

## Context

**Taken at**: https://dao.joystream.org/#/forum/category/11
**Fields**:
- Which feature are you missing?: Forum: tooltip to show full thread title
- What problem would that solve for you?: Can&#39;t see full thread titles, want to hover mouse and see the full thread title.
- Share your Github or Discord so we can get in touch: tomato#9464

**Browser**: Chrome 102 (Windows 10)
**Screen size**: 3440x1440
**[Open #22 in Usersnap](https://app.usersnap.com/l/feedback/21b38c92-781e-4fc1-9ed0-063bb3394674?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203239316151657) by [Unito](https://www.unito.io)

adovrodion:
Bounty was created: https://dao.joystream.org/#/bounty/preview/159

traumschule:
@mkbeefcake please check this one

polikosi:
tested on https://dao-git-fork-mkbeefcake-fix-tooltipshowthreadtitle-joystream.vercel.app/#/forum/category/1
✅
![image](https://user-images.githubusercontent.com/80429191/199026205-7e8bb637-9f8b-4630-af5f-b00996821a44.png)

traumschule:
(would assign @mkbeefcake but lack permission. @dmtrjsg @mnaamani)

PR #3760, opened 2022-10-25
added tooltip for thread title

mkbeefcake:
Fixed) Forum: tooltip to show full thread title #3273

### DIFF
--- a/packages/ui/src/forum/components/threads/ThreadListItem.tsx
+++ b/packages/ui/src/forum/components/threads/ThreadListItem.tsx
@@ -16,6 +16,7 @@ import { useMember } from '@/memberships/hooks/useMembership'
 
 import { LatestActivity, LatestActivityRowGapBlock } from './LatestActivity'
 import { ThreadTags } from './ThreadTags'
+import { Tooltip, TooltipDefault } from '@/common/components/Tooltip'
 
 interface ThreadListItemProps {
   thread: ForumThread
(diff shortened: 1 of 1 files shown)


packages/ui/src/forum/components/threads/ThreadListItem.tsx
```diff
@@ -35,10 +36,12 @@ export const ThreadListItem = ({ thread, isArchive }: ThreadListItemProps) => {
     >
       {thread.isSticky && <ThreadPinIcon />}
 
-      <Thread>
-        <TextBig bold>{thread.title}</TextBig>
-        {thread.tags.length > 0 && <ThreadTags tags={thread.tags} />}
-      </Thread>
+      <Tooltip tooltipText={thread.title}>
+        <Thread>
+          <TextBig bold>{thread.title}</TextBig>
```
traumschule, line 41: > "Radosław" add this line of code to the title component for truncation:
> truncateLines={2}

traumschule (changes requested):
LGTM. From the image this is already truncated. CSS or how does it work?
thread-title

packages/ui/src/forum/components/threads/ThreadListItem.tsx
mkbeefcake: image
This is the CSS which is for truncated features (I just checked that it already fixed)

WRadoslaw: `FullDots` will truncate first line only. `truncateLines` props on all `Text` components allows to pick number of line which will be truncated. `truncateLines={1}` should work as well

WRadoslaw (approved):

PR #3774, opened 2022-10-30
Show default target of WG Openings as 1 (#2539)

### LINKED ISSUE #2539: working-groups/* issues
## Scope

Sub-Issues to fix

1️⃣ 
**Context:** Pioneer FE creates the WGL opening correctly when done via GUI proposal flow; while on CLI creation of openings is possible via CLI, where the rule of setting the default hiring limit is not implemented. 
**To-do:**
- If the limit is not specified (on creating opening via CLI), display default limit as 1 on the opening card and details views. SP -1.

2️⃣ 
**Context:** described in the report below
**To-do:**
- [ ] Currently Staking amount to be fixed. 
🛠️ Tech tips:  Use the `sumStakes` function for this widget.
 SP - 1
 
⛔  **Out of scope**
 
- #2913 

________

Although the whole opening/application/hiring process has been improved significantly, there are still a few issues:

Screen Shot 2022-03-13 at 12 37 07
- When not setting a target, it defaults to 2. This **may** be ok for the "Worker" openings (although I don't think it's the best), for Lead openings, it makes no sense.
- "Currently Staking" (I'm currently staking for 5 roles - 5k for each) seems to be a set of the stake amounts, without a separator:
   - `[5000,5000,5000,5000,5000]`
   - displayed:
      - without  "," as the thousand separator
      - a space as stake separator
      - () for omitted due to not enough space
  - `(5000 500)0 5000 5000 5000`

Screen Shot 2022-03-13 at 12 41 27

- Also shows the `staked` amount as 100JOY and not hired, despite all of all but Storage Worker with ID 1 being hired.

ref #2537

traumschule:
Fixes #2539 part 1

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpening.tsx
@@ -209,7 +209,7 @@ const ApplicationStats = ({
       ) : (
         <StatiscticContentColumn>
           <StatisticHeader title="Hiring limit" />
-          <NumericValue>{hiring.limit}</NumericValue>
+          <NumericValue>{hiring.limit || 1}</NumericValue>
         </StatiscticContentColumn>
       )}
     </MultiColumnsStatistic>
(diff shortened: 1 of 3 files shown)


WRadoslaw (approved):
LGTM

PR #3776, opened 2022-10-30
Handle cancelled membership Modal tx (#2815)

### LINKED ISSUE #2815: Cancelling membership creation signing should not stop it from being re-triggerable
## Context 
Once membership creation flow is cancelled on the last stage, new membership button stops working, and so do other modals. 

https://www.loom.com/share/4729d10fee334974960e3551f7f902fd

## Refactoring scope

- [ ] We don't handle this for all transactions outside of bounties.
- [ ] State in xmachine is not handled by sript or by user-click
- [ ] Must be done globally  
- [ ] + Refactor bounties in-line with this approach

## Technical Guidelines

this was already fixed in #3264 Pls use the same approach here

traumschule:
Fixes #2815, membership modals part

If this looks ok i'll fix the other modals too.

### DIFF
--- a/packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipModal.tsx
+++ b/packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipModal.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 
 import { useApi } from '@/api/hooks/useApi'
+import { FailureModal } from '@/common/components/FailureModal'
 import { useFirstObservableValue } from '@/common/hooks/useFirstObservableValue'
 import { useMachine } from '@/common/hooks/useMachine'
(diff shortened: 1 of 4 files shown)


WRadoslaw:
Hey!

Looks like everything works as expected when canceling each of these modals. About the part with adding the modal for the canceled state - it was decided upon not to display such modal as we have the bottom one informing about canceled transaction.

image

**FYI**: RN common states across the modal are handled inside `GlobalModals.tsx` in `useGlobalModalHandler` hook. It was covered in modals refactor

traumschule:
I understand the design has changed, still we need to fix the bug. What should 'cancelled' step return instead?

traumschule:
Never mind, you claim it's fixed :)

Issue #3770, opened 2022-10-28
Resubscribing to QN on every transaction


dmtrjsg:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203259055086107) by [Unito](https://www.unito.io)

PR #3784, opened 2022-10-31
🪠 Resubscribe to QN state to validate tx

thesan:
- #3770

### DIFF
--- a/packages/ui/src/app/components/ConnectionStatusDot.tsx
+++ b/packages/ui/src/app/components/ConnectionStatusDot.tsx
@@ -1,5 +1,5 @@
 import { AnimatePresence, ForwardRefComponent, HTMLMotionProps, motion } from 'framer-motion'
-import React, { useMemo } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import styled, { StyledComponent } from 'styled-components'
 
 import { useApi } from '@/api/hooks/useApi'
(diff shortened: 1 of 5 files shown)


packages/ui/src/app/components/ConnectionStatusDot.tsx
```diff
@@ -18,16 +18,16 @@ interface ConnectionStatusDotProps {
...
+  const [isQnLate, setIsQnLate] = useState(false)
 
-  const isQnLate = useMemo(() => {
+  useEffect(() => {
     if (queryNodeState && header) {
-      return (
-        (header.toJSON().number as number) - Number(queryNodeState.indexerHead) > MAX_INDEXER_BLOCKS_BEHIND_NODE_HEAD
-      )
+      const processorDelay = Number(header.number) - Number(queryNodeState.indexerHead)
+      setIsQnLate(processorDelay > MAX_INDEXER_BLOCKS_BEHIND_NODE_HEAD)
```
traumschule, line 28: I assume `queryNodeState` is updated on every processed block which ideally happens very soon after block production except the node is under load / syncing. Should we allow a delay of a few blocks or would that defeat the purpose?

traumschule (approved):
Code looks ok, needs to be tested in action.

packages/ui/src/app/components/ConnectionStatusDot.tsx
thesan: Yes `MAX_INDEXER_BLOCKS_BEHIND_NODE_HEAD` is the allowed delay currently 20 blocks

PR #3793, opened 2022-11-01
Ease validation for `minContext` and `maxContext` (#2601)

### LINKED ISSUE #2601: Update validation copy on "Create WGL opening" Proposal
Update the validation copy of the top field  to ... equal or greater than {{execution constraint value}}

traumschule:
Fixes #2601

### DIFF
--- a/packages/ui/src/common/utils/validation.tsx
+++ b/packages/ui/src/common/utils/validation.tsx
@@ -33,7 +33,7 @@ export const maxContext = (
     }
 
     const validationValue = new BN(get(this.options.context, contextPath))
-    if (validationValue && validationValue.lte(new BN(value))) {
+    if (validationValue && validationValue.lt(new BN(value))) {
       return this.createError({
         message: msg,
         params: {
@@ -63,7 +63,7 @@ export const minContext = (
     }
 
     const validationValue = new BN(get(this.options.context, contextPath))
-    if (validationValue && validationValue.gte(new BN(value))) {
+    if (validationValue && validationValue.gt(new BN(value))) {
       return this.createError({
         message: msg,
         params: {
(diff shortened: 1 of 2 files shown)


packages/ui/src/common/utils/validation.tsx
```diff
@@ -33,7 +33,7 @@ export const maxContext = (
     }
 
     const validationValue = new BN(get(this.options.context, contextPath))
-    if (validationValue && validationValue.lt(new BN(value))) {
+    if (validationValue && validationValue.lte(new BN(value))) {
```
thesan, line 36: The previous logic is correct:
It returns an error if `validationValue` is strictly less than `value`.
So `value` is valid when it is greater or equal to `validationValue`

So only the validation messages were wrong

packages/ui/src/common/utils/validation.tsx
```diff
@@ -63,7 +63,7 @@ export const minContext = (
     }
 
     const validationValue = new BN(get(this.options.context, contextPath))
-    if (validationValue && validationValue.gt(new BN(value))) {
+    if (validationValue && validationValue.gte(new BN(value))) {
```
thesan, line 66: Same thing here

thesan (changes requested):

packages/ui/src/common/utils/validation.tsx
traumschule: Makes sense.

packages/ui/src/common/utils/validation.tsx
traumschule: Fixed in [46462da](https://github.com/Joystream/pioneer/pull/3793/commits/46462da0b5188b7e92efb660b0ff707266d56083)

thesan (approved):
Sorry for the delay I forgot about this one. Whenever you can please try to remember to re-request reviews after a PR is fixed

Issue #3660, opened 2022-10-05
Some voting/candidacy stakes migth be wrongfully locked


thesan:
## Context

`@/accounts/hooks/useRecoveryConditions` assumes that voting and candidacy locks always relate to the latest election. This assumption looks wrong. However it's strange that this was not reported earlier, so maybe there is more to it.

Also the current logic is kind of unintuitive. `useRecoveryConditions` is only called inside `LocksDetails` so `recoverable` and `nonRecoverable` can be grouped in the component. IMO `useRecoveryConditions` should be replaced by something like: 
```js
const [recoverable, nonRecoverable] = useGroupLocks(address, locks)
```

## Scope

- Check whether voting locks become unrecoverable when the next election start or when the voted for candidate is elected in a later election
- Also check if this assumption affect the candidacy/councilor lock
- Simplify the logic a bit

polikosi:
![image](https://user-images.githubusercontent.com/80429191/197527543-2fe71a8c-0e6d-4091-ba5d-de7dae566d11.png)

**candidacy stakes**
![image](https://user-images.githubusercontent.com/80429191/197528234-02e916fd-7f3d-477d-b9f5-b6f86cf267ef.png)
**voting stakes**
![image](https://user-images.githubusercontent.com/80429191/197529349-cef6888a-aefa-4825-b90b-8ee3a9ffb7e3.png)


**voting stakes can be recovered when the next election start**
![image](https://user-images.githubusercontent.com/80429191/197531976-03a4148c-0b9e-4b2d-afaa-2ef423429806.png)

**Councilor stakes can be recovered when the next election start**
![image](https://user-images.githubusercontent.com/80429191/197532345-ad799c4c-7317-4408-ac8b-96ee114770d1.png)

**candidancy who not won can be recover immediately**
![image](https://user-images.githubusercontent.com/80429191/197532819-c7c97117-5300-4f27-926a-36d429f4c8c9.png)

🚫once the next election begins voting stakes can't be recover
![image](https://user-images.githubusercontent.com/80429191/197540554-60064ca0-49f4-477f-8d13-d7bd6fb76e60.png)


@dmtrjsg 
🚫update ESTIMATED TIME TO RECOVER for voting. So that it showed the same time as the Councilor
![image](https://user-images.githubusercontent.com/80429191/197533889-94e90e8e-d251-4d7a-bccf-9555ffa31a61.png)

polikosi:
[...]

PR #3795, opened 2022-11-02
⏳ Fix the recovery time of the voting and candidacy locks

thesan:
Fix the issue found during [3660 QA](https://github.com/Joystream/pioneer/issues/3660#issuecomment-1297078053)


traumschule (approved):
LGTM, few duplicate strings but this should go into QA asap.

WRadoslaw (approved):

PR #3796, opened 2022-11-02
Don't overwrite watchlist (#3769)

### LINKED ISSUE #3769: Isolate localstorage writes from the forum tread watchlist
Once #3719 will have made its way into `dev`. To prevent users from loosing some of their watchlist by adding threads from different browser tabs, it would be good to change these into something like:
```ts
setWatchlist((watchlist = []) => [...watchlist, threadId])
setWatchlist((watchlist = []) => watchlist.filter((id) => id !== threadId))
```

_Originally posted by @thesan in https://github.com/Joystream/pioneer/pull/3654#discussion_r1007942049_

traumschule:
Fixes #3769

2nd commit is optional, your choice.

### DIFF
--- a/packages/ui/src/forum/components/Thread/WatchlistButton.tsx
+++ b/packages/ui/src/forum/components/Thread/WatchlistButton.tsx
@@ -21,7 +21,7 @@ export const WatchlistButton = ({ threadId }: Props) => {
   const removeFromWatchlist = useCallback(
     (e) => {
       e.preventDefault()
-      setWatchlist(watchlist?.filter((id) => id !== threadId) ?? [])
+      if (watchlist) setWatchlist([...watchlist, threadId].filter((id) => id !== threadId))
     },
     [watchlist]
   )
--- a/packages/ui/src/forum/components/Thread/WatchlistButton.tsx
+++ b/packages/ui/src/forum/components/Thread/WatchlistButton.tsx
@@ -14,14 +14,14 @@ export const WatchlistButton = ({ threadId }: Props) => {
   const addToWatchlist = useCallback(
     (e) => {
       e.stopPropagation()
-      setWatchlist(watchlist ? [...watchlist, threadId] : [threadId])
+      setWatchlist((watchlist = []) => [...watchlist, threadId])
     },
     [watchlist]
   )
   const removeFromWatchlist = useCallback(
     (e) => {
       e.preventDefault()
-      if (watchlist) setWatchlist([...watchlist, threadId].filter((id) => id !== threadId))
+      setWatchlist((watchlist = []) => watchlist.filter((id) => id !== threadId))
     },
     [watchlist]
   )


packages/ui/src/forum/components/Thread/WatchlistButton.tsx
```diff
@@ -13,7 +13,7 @@ export const WatchlistButton = ({ threadId }: Props) => {
   const [watchlist, setWatchlist] = useLocalStorage<string[]>(FORUM_WATCHLIST)
   const addToWatchlist = useCallback(() => setWatchlist(watchlist ? [...watchlist, threadId] : [threadId]), [watchlist])
   const removeFromWatchlist = useCallback(
-    () => setWatchlist(watchlist?.filter((id) => id !== threadId) ?? []),
```
thesan, line 16: @traumschule currently this code: `setWatchlist(watchlist?.filter((id) => id !== threadId) ?? []),` is fine, it works just as well as `setWatchlist(watchlist ? [...watchlist, threadId].filter((id) => id !== threadId) : [])` the issue I mentioned in #3769 is the value of `watchlist` comes from the local storage state at the moment the component was mounted. It's fine in most cases but if someone had different threads opened in different browser tabs/windows: if 2 threads are added in 2 different tabs/windows the watchlist will only have the latest addition.

A similar problem caused #3711 fixed here: #3719 by mimicking the `useState` api (i.e when providing a function perform an updated) except that update function are provided with an up to date value from the local storage. But this PR was merged in `carthage` this is why we should wait for `carthage` to be merged into `dev` to fix this.

I'm closing this PR in the meantime.

traumschule: Sounds good and sorry for the extra work! I'll keep this as low-prio draft on my end.

traumschule:
As discussed above, feel free to assign a target when this comes on the agenda.

thesan:
@traumschule #3719 is now on `dev` so this can be worked on

packages/ui/src/forum/components/Thread/WatchlistButton.tsx
```diff
@@ -21,7 +21,7 @@ export const WatchlistButton = ({ threadId }: Props) => {
   const removeFromWatchlist = useCallback(
     (e) => {
       e.preventDefault()
-      setWatchlist(watchlist?.filter((id) => id !== threadId) ?? [])
+      if (watchlist) setWatchlist([...watchlist, threadId].filter((id) => id !== threadId))
```
thesan, line 24: The idea behind this issue is that by the time `addToWatchlist` or `removeFromWatchlist` are called, the `watchlist` variable might be out of date due to changes made from other window (or potentially other components since `useLocalStorage` doesn't rely on any reactive global context AFAIK although it's probably irrelevant for this use case).

So the proposed solution is to pass an update function to `setWatchlist` (similarly to how react `useState` works [here](https://beta.reactjs.org/apis/react/useState#updating-state-based-on-the-previous-state)). The key is that this update function is given the value of the local storage at the time of the update opposed to the time of the component render (incidentally it also updates the value for the rest of the component after being called).

This way this line would be:
```suggestion
      setWatchlist((watchlist = []) => watchlist.filter((id) => id !== threadId))
```

And line 17:
```ts
setWatchlist((watchlist = []) => [...watchlist, threadId])
```

thesan (changes requested):

traumschule:
new (related?) type errror
```
ERROR in src/bounty/components/BountyDiscussion/BountyDiscussion.tsx:23:8
TS2741: Property 'replyToPost' is missing in type '{ threadId: string; isLoading: boolean; isThreadActive: boolean; isDiscussion: true; }' but required in type 'PostListProps'.
    21 |
    22 |   const replyToPost = useCallback(
  > 23 |     (post: ForumPost) => {
       |        ^^^^^^^^
    24 |       if (thread) {
    25 |         return push(`${generatePath(ForumRoutes.thread, { id: thread.id })}?post=${post.id}`)
    26 |       }
```

oleksanderkorn (approved):

freakstatic (approved):
LGTM!

thesan (approved):

PR #3797, opened 2022-11-03
Votes backup CTA on `VoteForCouncilSuccessModal` (#1633)

### LINKED ISSUE #1633: Expose and backup voting salt

traumschule:
Fixes #1633

### DIFF
--- a/packages/ui/src/council/modals/VoteForCouncil/VoteForCouncilSuccessModal.tsx
+++ b/packages/ui/src/council/modals/VoteForCouncil/VoteForCouncilSuccessModal.tsx
@@ -6,6 +6,7 @@ import { ButtonGhost } from '@/common/components/buttons'
 import { SuccessIcon } from '@/common/components/icons'
 import { Modal, ModalFooter, ModalHeader, SuccessModalBody } from '@/common/components/Modal'
 import { TextMedium } from '@/common/components/typography'
+import { BackupVotesButton } from '@/council/components/election/BackupVotesButton'
 import { ElectionRoutes } from '@/council/constants'
 import { useCandidate } from '@/council/hooks/useCandidate'
 import { SelectedMember } from '@/memberships/components/SelectMember'
@@ -30,9 +31,10 @@ export const VoteForCouncilSuccessModal = ({ onClose, candidateId }: Props) => {
 
       <SuccessModalBody>
         <TextMedium margin="l" light>
-          You have just successfully voted for the Candidate
+          You have just successfully voted for the Candidate. This is a good time to backup your votes.
         </TextMedium>
-
+        <BackupVotesButton cycleId={candidate?.cycleId} />
+        ￼
         <SelectedMember size="l" member={candidate?.member} hideGroup />
       </SuccessModalBody>
 
--- a/packages/ui/src/council/modals/VoteForCouncil/VoteForCouncilSuccessModal.tsx
+++ b/packages/ui/src/council/modals/VoteForCouncil/VoteForCouncilSuccessModal.tsx
@@ -33,8 +33,7 @@ export const VoteForCouncilSuccessModal = ({ onClose, candidateId }: Props) => {
         <TextMedium margin="l" light>
           You have just successfully voted for the Candidate. This is a good time to backup your votes.
         </TextMedium>
-        <BackupVotesButton cycleId={candidate?.cycleId} />
-        ￼
+        <BackupVotesButton cycleId={candidate?.cycleId} /> ￼
         <SelectedMember size="l" member={candidate?.member} hideGroup />
       </SuccessModalBody>
 
--- a/packages/ui/test/council/pages/Election.test.tsx
+++ b/packages/ui/test/council/pages/Election.test.tsx
@@ -103,9 +103,6 @@ describe('UI: Election page', () => {
     },
   }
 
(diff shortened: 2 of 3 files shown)


packages/ui/src/council/modals/VoteForCouncil/VoteForCouncilSuccessModal.tsx
```diff
@@ -30,9 +31,10 @@ export const VoteForCouncilSuccessModal = ({ onClose, candidateId }: Props) => {
 
       <SuccessModalBody>
         <TextMedium margin="l" light>
-          You have just successfully voted for the Candidate
+          You have just successfully voted for the Candidate. This is a good time to backup your votes.
         </TextMedium>
-
+        <BackupVotesButton cycleId={candidate?.cycleId} />
+        ￼
```
thesan, line 37: @traumschule is this on purpose ?

thesan (changes requested):
The tests fails because the `@testing-library` mocked version of `URL` is missing the `createObjectURL` method (I think it might be missing `revokeObjectURL`too). To fix the tests this method has to be mocked. I think it's done somewhere else in the codebase.

packages/ui/src/council/modals/VoteForCouncil/VoteForCouncilSuccessModal.tsx
traumschule: i added empty lines around `BackupVotesButton` for clarify but prettier had another view, or do you mean something else?

traumschule:
```
yarn test election
yarn run v1.22.19
$ jest election
 PASS  test/council/components/ElectionTabs.test.tsx
 PASS  test/council/pages/PastElection.test.tsx (11.096 s)
 PASS  test/council/components/ElectionVotes.test.tsx (11.591 s)
 PASS  test/council/pages/Election.test.tsx (13.933 s)

Test Suites: 4 passed, 4 total
Tests:       43 passed, 43 total
Snapshots:   0 total
Time:        15.806 s
Ran all test suites matching /election/i.
Done in 16.71s.
```

thesan (approved):
I just moved the URL mocks in the jest `setup.ts`, so this shouldn't be a problem anymore

traumschule:
forum and proposal tests failed, seems unrelated this time.

dmtrjsg:
Pls update the copy to 

"Import your file with voting salt here and submit the transaction to restore votes."

Screenshot 2022-11-24 at 13 16 40

Issue #2977, opened 2022-05-05
Back button on openings/# leads to nothing if referrer is empty


eikebartels:
To reproduce: 

Open new tab and open https://dao.joystream.org/#/working-groups/openings/operationsWorkingGroupAlpha-9 
press the back button 
![image](https://user-images.githubusercontent.com/767979/166952236-436797d4-821d-4871-92bd-affc2877e4e0.png)

Expected behaviour:
👉 ✅ Option 2: Button leads to list page https://dao.joystream.org/#/working-groups/openings 

Actually behaviour: 
Button leads to empty tab page

traumschule:
use solution suggested in #3768

ivanturlakov:
✅ Tested on https://dao-git-staging-joystream.vercel.app/#/working-groups/openings
wss://34.230.5.182.nip.io/ws-rpc

Works as expected

https://user-images.githubusercontent.com/46903215/210771130-e192852a-eddc-4f3a-8e85-16dc20631c08.mov

Issue #3768, opened 2022-10-28
Joystream Pioneer Forum Back Link


Songoku1691:
### Scope 
- [ ]  Add page number to the url so "back" in-app button brings to the right previous page in case of paginated prev page.

### Context

Hello Guys;

I observed a limitation in Joystream Pioneer Forum. I'm not sure if it's a bug (not working) or improvement (no feature of this function).
Please see this Atlas video: https://play.joystream.org/video/27767

You will notice I clicked a forum thread on Page 9 of a Category. When I clicked back, I assumed I would go back to Page 9 of Category. Instead, I was brought to Page 1. So if I want to check the next thread on that page, I will need to click again Page 9.

I hope I explained it well.

Thank you

dmtrjsg:
Thanks you've explained it very well and extra points for the nice music 😄   I agree this is pretty annoying for a heavy Forum user..

vrrayz:
I'll like to take this up as well. I just dropped a fix for [#2977](https://github.com/Joystream/pioneer/issues/2977) in this PR [here](https://github.com/Joystream/pioneer/pull/3798)

vrrayz:
I just fixed this here #4147

ivanturlakov:
✅ Tested on https://dao-git-dev-joystream.vercel.app/#/forum/category/1

This case works as expected


https://user-images.githubusercontent.com/46903215/222966435-c3f19996-532d-4333-b922-5f6ad44a55d4.mov



Also, I found another scenario with "back btn" bug #4215

PR #3798, opened 2022-11-03
Issue 2977

vrrayz:
Back button on openings now list to list page as requested from this [this issue](https://github.com/Joystream/pioneer/issues/2977)


traumschule (approved):
LGTM! Keep on it with #3768 if you want.

WRadoslaw (approved):
LGTM!

WRadoslaw (changes requested):
Actually, I just noticed that this solution causes one bug. Problem is that when you go into opening details from a specific WG page, the back button will always transfer you to general openings instead of given WG ones.

Two cases have to be considered before pushing `customLink` to the stack:

- Whether `history.length` is bigger than 2 (if stack eq `1` then, the user clicks the link and get redirected, if stack eq `2` there is a possibility that the user opened a new tab in the browser and paste in the link
- Another possibility is that the user clicks on the URL to the openings list and then clicks on the given opening, in that case, checking for stack length will fail (stack length will equal `2`, but `goBack` should be invoked anyway), but we can check the current action on the history. I see that when you paste the URL or click on the link action will eq `POP` while if you navigate from Pioneer forward (Opening list -> Opening details) it will be `PUSH`

Integrating these two rules should do the work. WDYT?

vrrayz:
> Actually, I just noticed that this solution causes one bug. Problem is that when you go into opening details from a specific WG page, the back button will always transfer you to general openings instead of given WG ones.
> 
> Two cases have to be considered before pushing `customLink` to the stack:
> 
> * Whether `history.length` is bigger than 2 (if stack eq `1` then, the user clicks the link and get redirected, if stack eq `2` there is a possibility that the user opened a new tab in the browser and paste in the link
> * Another possibility is that the user clicks on the URL to the openings list and then clicks on the given opening, in that case, checking for stack length will fail (stack length will equal `2`, but `goBack` should be invoked anyway), but we can check the current action on the history. I see that when you paste the URL or click on the link action will eq `POP` while if you navigate from Pioneer forward (Opening list -> Opening details) it will be `PUSH`
> 
> Integrating these two rules should do the work. WDYT?

I just made modifications based on your second suggestion. I think tracking the link action is all that's needed. I tested out for those cases you mentioned, and they work fine now

traumschule (approved):
lgtm! tested entering opening details from `/working-groups/openings` and `/working-groups/forum`.

oleksanderkorn (changes requested):
Why does it have all the types changes here? I guess this can be done in a separate change if needed.

traumschule:
You are right, didn't look at the code. @vrrayz resubmit the fix on a new branch.

Issue #3382, opened 2022-07-22
Lock icons overlap balance on account select dropdown


ivanturlakov:
Tested on staging

<img width="1249" alt="Снимок экрана 2022-07-22 в 13 14 34" src="https://user-images.githubusercontent.com/46903215/180419057-ece744be-da5d-44d5-ba84-fe11bdf91808.png">

dmtrjsg:
looks like this may be caused by 
- https://github.com/Joystream/pioneer/pull/3351

@singulart adding to project 55 to get assigned as part of stab9 release.

singulart:
Should this be in progress or in To do?

singulart:
@dmtrjsg

dmtrjsg:
@singulart In progress I would say to be released together with the recent stabilisation9 scope which is due next.

adovrodion:
Bounty was created: https://dao.joystream.org/#/bounty/preview/208

adovrodion:
Bounty was created: https://dao.joystream.org/#/bounty/preview/259

adovrodion:
Bounty was created: https://dao.joystream.org/#/bounty/preview/431

XxFlameCatxX:
Like this? I'm planning to fix it. It seems like a CSS bug, but since the project manage with different programming language i will need to adapt to it

dmtrjsg:
@traumschule 

please decide if this can be done together with 
- https://github.com/Joystream/pioneer/issues/3787#issuecomment-1298098690

and feel free to assign both to a single person if that makes sense from your pov.

Just saying that Lock icons are misplaced in three locations.

XxFlameCatxX:
okay, changing margin-top don't do much thing. It seems a responsvity error. Which resolution are you using? Maybe it's related to resolution thing

PR #3803, opened 2022-11-03
Adding "align-items" (need testing)

XxFlameCatxX:
Depiste this update it need testing, i have no idea how responsive is the joystream, so it's possibly the responsivity. But i will need it being tested with different resolutions. I can't test it wihout having a bigger screen. SO you guys could test it for me, i don't know if it will work

Problem: https://github.com/Joystream/pioneer/issues/3382


oleksanderkorn (changes requested):

oleksanderkorn:
@XxFlameCatxX it looks like the `top` posititon of `AccountLocksWrapper` can be adjusted to add more space between icons wrapper and account balance, this works for me, wdyt?
```
export const AccountLocksWrapper = styled(ColumnGapBlock)`
  position: absolute;
  top: 20px;
  align-items: center;
`
```

image

.gitignore
```diff
@@ -6,4 +6,4 @@ build-storybook
 *.log
 .DS_Store
 packages/ui/.env
-packages/ui/src/services/i18n/utils/csv
+packages/ui/src/services/i18n/utils/csv
```
oleksanderkorn, line 9:
```suggestion
packages/ui/src/services/i18n/utils/csv

```

oleksanderkorn (changes requested):
add new line

oleksanderkorn (approved):

traumschule:
Failing check in [DeleteThreadModal](https://github.com/Joystream/pioneer/pull/3803/checks#step:7:55) seems unrelated.

packages/ui/src/accounts/components/AccountItem/components/styles.ts
```diff
@@ -72,6 +72,7 @@ export const LocksButtons = styled.div`
   align-self: end;
   display: flex;
   column-gap: 10px;
+  align-items: center;
```
oleksanderkorn, line 75: I thought i wrote this before but could not find my comment:
This  line can be removed, has no impact at all.

oleksanderkorn (changes requested):

packages/ui/src/accounts/components/AccountItem/components/styles.ts
traumschule: Had the same feeling. Likely because of commit overlap with #3828.

XxFlameCatxX:
closed

Issue #3251, opened 2022-06-16
Loss of tJOY in the application


adovrodion:
## Scope

- [ ] Fix recovering the role staking after role was closed via proposal. 

Add a button to "recover lock" and follow the general designs for it

Out of scope:
- [x] Fix showing past openings closed via proposal. Opening that was closed via proposal is missing in the openings history of given WG. (resolved in -https://github.com/Joystream/pioneer/issues/3348)

## Context

There have been complaints from users that their tJOYs are missing somewhere in the staking.
https://discord.com/channels/811216481340751934/812345044094156811/984719368673976350
https://discord.com/channels/811216481340751934/815192060276506664/984516585404366898

And today the consuls were recreating the openings for the group leaders' positions. 
The opening had application from user. One user wrote back that he did not get tJOY back after closing the opening through proposal. The tokens are considered blocked in the wallet.
Also, the opening that was closed to the councils through proposal does not appear in the history of the group openings.

traumschule:
are the tokens burned or locked? a screenshot or wallet address would help to debug.

adovrodion:
> are the tokens burned or locked? a screenshot or wallet address would help to debug.
@traumschule 
![image](https://user-images.githubusercontent.com/84832436/174251717-6b77d658-9af9-453b-8efa-cdc065f0b035.png)

5FdtkcMwGtPWH7bX7tLto39SxTG5zwyxBCXfXcmFHykCHp74
5D5K5fmQoxC2aubHM7DwPUntHEEJvdkprDwPxicbDJbcvxMF

dmtrjsg:
@thesan to check if there is an extrensic to recover apply on opening stake for this action..

dmtrjsg:
@adovrodion do we have any info on what specific role was cancelled via proposal?

adovrodion:
@dmtrjsg This openings:
https://dao.joystream.org/#/working-groups/openings/operationsWorkingGroupGamma-6 
https://dao.joystream.org/#/working-groups/openings/contentWorkingGroup-12

dmtrjsg:
When someone applies to application and application is cancelled via proposal - runtime does not pick it up and staking amount gets locked for eternity. Reclaim staking amount exrensic does not exist, so is not supported for the working groups. 

There is only withdrawal of the application, but if the application is closed it is unclear how this can be done. Our (pioneer team) assumption is that staking needs to be automatically returned. 

Right now user can only withdraw application for the active opening. 

[...]

PR #3823, opened 2022-11-10
🎊 Consider all statues in checking WG lock state

WRadoslaw:
- #3251


packages/ui/src/accounts/hooks/useGroupLocks/helpers.ts
```diff
@@ -18,19 +18,16 @@ export const useIsWGLockRecoverable = (hasWGLock: boolean, stakingAccount: strin
...
-  if (applications[0]?.status === 'ApplicationStatusPending') {
-    return true
-  }
-
-  if (applications[0]?.status === 'ApplicationStatusAccepted') {
+  if (status === 'ApplicationStatusAccepted') {
     if (unstakingPeriodEnd) {
       return Date.parse(unstakingPeriodEnd) > Date.now()
     }
     return true
```
thesan, line 27: @WRadoslaw could you remind me why if there is no unstaking period the stake is recoverable and why is the unstaking period happening when the application is accepted please ? I don't remember how this is supposed to work...

WRadoslaw: Good question. I think this logic was inherit from Rafal's lock refactor PR, but now when I look at it, it seems invalid. Since `useWorkerUnstakingPeriodEnd` will return a falsy value if the worker didn't leave/got kicked.

Unstaking period happens only when the given application was already accepted, so the applicant is now a worker.   More [here](https://joystream.gitbook.io/testnet-workspace/system/working-groups#worker).

```suggestion
      return Date.parse(unstakingPeriodEnd) > Date.now()
    }
    return false
```

thesan (approved):
:+1:

Issue #3736, opened 2022-10-21
CI keeps on failing at `UI: DeleteThreadModal › Transaction success` 


thesan:
See: https://github.com/Joystream/pioneer/actions/runs/3297294384/jobs/5437939367

thesan:
It happened again here: https://github.com/Joystream/pioneer/actions/runs/3565964834/jobs/5991807716#step:7:547

PR #3824, opened 2022-11-10
🧪 Test fix on DeleteThreadModal tests

WRadoslaw:
- #3736

### DIFF
--- a/packages/ui/test/setup.ts
+++ b/packages/ui/test/setup.ts
@@ -64,7 +64,7 @@ const defaultMockedTransactionFee: UseTransaction = {
 const mockedTransactionFee = jest.fn(() => defaultMockedTransactionFee)
 
 export const mockTransactionFee = (value: Partial<UseTransaction>) => {
-  mockedTransactionFee.mockReturnValue({ ...defaultMockedTransactionFee, ...value })
+  mockedTransactionFee.mockReturnValue({ ...mockedTransactionFee(), ...value })
 }
 
 jest.mock('@/accounts/hooks/useTransactionFee', () => ({
--- a/packages/ui/test/setup.ts
+++ b/packages/ui/test/setup.ts
@@ -54,15 +54,21 @@ export const mockDefaultBalance = {
   total: BN_THOUSAND,
 }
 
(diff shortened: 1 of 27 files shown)


thesan (commented):
@WRadoslaw it looks fine, but I see that: `mockedTransactionFee.feeInfo = { ... }` is called in a bunch of other places. Could this cause similar failures ? WDYT of changing the transaction fee mock inside `setup.ts` to:
```ts
const defaultMockedTransactionFee = {
  transaction: undefined,
  setTransaction: () => undefined,
  setSigner: () => undefined,
  feeInfo: { transactionFee: BN_ZERO, canAfford: true },
}

const mockedTransactionFee = jest.fn<UseTransaction, []>(() => defaultMockedTransactionFee)

export const updateMockedTransactionFee = (update: Partial<UseTransaction>): void => {
  mockedTransactionFee.mockReturnValue({ ...defaultMockedTransactionFee, ...update })
}

jest.mock('@/accounts/hooks/useTransactionFee', () => ({
  useTransactionFee: mockedTransactionFee,
}))
```

This way `updateMockedTransactionFee({ feeInfo: { ... } })` can be called to set transaction a fees value scoped to a test or suite, without mutating the global value.

packages/ui/test/setup.ts
```diff
@@ -54,15 +54,21 @@ export const mockDefaultBalance = {
...
   transaction: undefined,
   setTransaction: () => undefined,
   setSigner: () => undefined,
   feeInfo: { transactionFee: BN_ZERO, canAfford: true },
 }
 
+const mockedTransactionFee = jest.fn(() => defaultMockedTransactionFee)
+
+export const mockTransactionFee = (value: Partial<UseTransaction>) => {
+  mockedTransactionFee.mockReturnValue({ ...defaultMockedTransactionFee, ...value })
```
thesan, line 67:
```suggestion
  mockedTransactionFee.mockReturnValue({ ...(mockedTransactionFee() ?? defaultMockedTransactionFee), ...value })
```

thesan (approved):

Issue #2309, opened 2022-02-23
Opening card for Fill WGL Proposal


dmtrjsg:
### Scope

- [ ] Update opening card based on mockups  [Figma >](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-Design?node-id=199%3A3016)

<img width="445" alt="Screenshot 2022-03-04 at 15 33 45" src="https://user-images.githubusercontent.com/93924525/156792675-0516bd86-0f8b-4359-87d4-322a1142f387.png">

- [ ] Add padding for ID
- [ ] Add "lead" badge for WGL component variation

Out of scope ⛔ 

- [ ] Add reward 

> for Fill working group lead proposal

#### Usersnap ticket details

**Taken at**: https://pioneer-2.vercel.app/#/proposals/current
**Fields**:
- Discrod Handle: -
- Give us some details: Show me something I can relate to.
- Houston, we have a problem.: This isn&#39;t a good way to distinguish the openings.
- -
- 
![Screen Shot 2022-02-23 at 15 07 26](https://user-images.githubusercontent.com/35237943/155335486-30095ac5-4c06-4a73-bd6d-fb8a10e65897.png)
*replaced*

**Browser**: Firefox 97 (macOS Catalina)
**Screen size**: 2560x1440
**Browser size**: 2560x1304
**[Open #13 in Usersnap](https://app.usersnap.com/l/feedback/0c731d09-1a02-4d5a-9415-323630a771e7?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

traumschule:
@gyroflaw can you have a look please

gyroflaw:
Ok, I will take a look

traumschule:
Just checked figma and wonder why it shows `LEAD` badge for all openings. Probably  safe to ignore.
![lead-badge](https://user-images.githubusercontent.com/31551045/206426422-160763e6-181f-43f4-bb0a-29fd4eb62806.png)

ivanturlakov:
Tested on https://dao-git-dev-joystream.vercel.app/#/working-groups/openings

### ✅ Opening card
[...]

PR #3829, opened 2022-11-14
Fix/WGL-Proposal

gyroflaw:
Fix https://github.com/Joystream/pioneer/issues/2309


oleksanderkorn (approved):
Should we mark them `past` by default? I see it is also used here:
image

packages/ui/src/working-groups/components/OpeningsList/OpeningsList.tsx
```diff
@@ -10,7 +10,7 @@ export interface OpeningsListProps {
   past?: boolean
 }
 
-export const OpeningsList = ({ openings, past }: OpeningsListProps) => {
+export const OpeningsList = ({ openings, past = true }: OpeningsListProps) => {
```
traumschule, line 13: agree, there's possible downside.
```suggestion
export const OpeningsList = ({ openings, past }: OpeningsListProps) => {
```

traumschule (changes requested):

traumschule (approved):

packages/ui/src/working-groups/components/OpeningsList/Opening/OpeningListItem.tsx
```diff
@@ -33,9 +33,7 @@ export const OpeningListItem = ({ opening, past, onClick }: OpeningListItemProps
       <ToggleableItemInfo>
         <ToggleableItemInfoTop>
           <Subscription title={`ID: ${opening.id}`}>ID: {opening.runtimeId}</Subscription>
-          {isInFuture(opening.expectedEnding) && (
-            <Subscription>Ends {relativeTime(opening.expectedEnding)}</Subscription>
-          )}
+          <Subscription>Duration: {relativeTime(opening.createdAtBlock.timestamp)}</Subscription>
```
thesan, line 36: image

I'm not 100% sure but in my understanding we should still show the time left until the `opening.expectedEnding` here. In which case it would still be a countdown and it should still stop showing after the `expectedEnding` has passed. Please could you confirm @dmtrjsg ?

thesan (changes requested):
I don't understand the changes to `Option`, `Select`, `OptionsListWorkingGroupOpening`, and `SelectWorkingGroupOpening`, I don't see any new prop passed to any of these components. Am I missing something or is it fixing some warning somewhere ?

packages/ui/src/working-groups/components/OpeningsList/Opening/OpeningListItem.tsx
dmtrjsg: @thesan 💯 agreed, let's change the label from "Duration" to "Ends in" and calculate the countdown, instead of time elapsed from creation. 

Sorry for this, as the designs seem to be incorrect in Figma. Thank you all for understanding :)

gyroflaw: Fixed

gyroflaw:
> I don't understand the changes to `Option`, `Select`, `OptionsListWorkingGroupOpening`, and `SelectWorkingGroupOpening`, I don't see any new prop passed to any of these components. Am I missing something or is it fixing some warning somewhere ?

All these changes are required due to styling.
We use `Styled-component`(below library) for styling and [here](https://github.com/gyroflaw/pioneer/blob/506edb8c4e346fb9d2cc477e19de7f5a22173b9e/packages/ui/src/working-groups/components/SelectWorkingGroupOpening/OptionsListWorkingGroupOpening.tsx#L40) styles `Option` component

 ```ts
 const StyledOption = styled(Option)`
 ```

 and the library generates a CSS class for that style and assigns it to the component `className` props. If we omitted `{...props}` in the component this style won't be applied.
 
image

packages/ui/src/working-groups/components/OpeningsList/Opening/OpeningListItem.tsx
```diff
@@ -33,7 +33,9 @@ export const OpeningListItem = ({ opening, past, onClick }: OpeningListItemProps
       <ToggleableItemInfo>
         <ToggleableItemInfoTop>
           <Subscription title={`ID: ${opening.id}`}>ID: {opening.runtimeId}</Subscription>
-          <Subscription>Duration: {relativeTime(opening.createdAtBlock.timestamp)}</Subscription>
+          {isInFuture(opening.expectedEnding) && (
+            <Subscription>Ends {relativeTime(opening.expectedEnding)}</Subscription>
```
traumschule, line 37: Looks good for decided proposals but "Ends" vs "[Ends in](https://github.com/Joystream/pioneer/pull/3829#discussion_r1054157981)".

filled

gyroflaw: Done

packages/ui/src/common/components/selects/components.tsx
```diff
@@ -24,9 +24,11 @@ interface Props {
   disabled?: boolean
 }
 
-export const Option = ({ children, onClick, disabled }: Props) => (
+export const Option = ({ children, onClick, disabled, ...props }: Props) => (
   <OptionComponentContainer onClick={onClick} disabled={disabled}>
-    <OptionComponent disabled={disabled}>{children}</OptionComponent>
+    <OptionComponent disabled={disabled} {...props}>
```
thesan, line 29: @gyroflaw
> All these changes are required due to styling. We use `Styled-component`(below library) for styling and [here](https://github.com/gyroflaw/pioneer/blob/506edb8c4e346fb9d2cc477e19de7f5a22173b9e/packages/ui/src/working-groups/components/SelectWorkingGroupOpening/OptionsListWorkingGroupOpening.tsx#L40) styles `Option` component
> 
> ```ts
> const StyledOption = styled(Option)`
> ```
> 
> and the library generates a CSS class for that style and assigns it to the component `className` props. If we omitted `{...props}` in the component this style won't be applied.
> 
> image

Right I missed that for `Option`, however only the `className` prop is needed to get the styled method to work (see: https://styled-components.com/docs/basics#styling-any-component).

It would better to just pass `className` because it makes it clear that the style is what is being passed:
```ts
export const Option = ({ children, onClick, disabled, className }: Props) => (
  <OptionComponentContainer onClick={onClick} disabled={disabled}>
    <OptionComponent disabled={disabled} className={className}>{children}</OptionComponent>
```

packages/ui/src/common/components/selects/select.tsx
```diff
@@ -48,6 +48,7 @@ export const Select = <T extends any, V extends any = T>({
   renderList,
   className,
   onBlur,
+  ...props
```
thesan, line 51: Here it shouldn't be needed since `className` is already forwarded
```suggestion
```

packages/ui/src/working-groups/components/SelectWorkingGroupOpening/SelectWorkingGroupOpening.tsx
```diff
@@ -35,6 +35,7 @@ export const SelectWorkingGroupOpeningBase = ({
   className,
   groupId,
   openingsPositionType,
+  ...props
```
thesan, line 38: Same here
```suggestion
```

packages/ui/src/working-groups/components/SelectWorkingGroupOpening/OptionsListWorkingGroupOpening.tsx
```diff
@@ -13,13 +16,13 @@ interface Props {
   onChange: (option: WorkingGroupOpening) => void
 }
 
-export const OptionsListWorkingGroupOpening = React.memo(({ allOpenings, onChange }: Props) => (
-  <OptionsListComponent>
+export const OptionsListWorkingGroupOpening = React.memo(({ allOpenings, onChange, ...props }: Props) => (
+  <OptionsListComponent {...props}>
```
thesan, line 20: Here also the `className` would be enough to extend this component style but I don't see this done anywhere. So I think the change should just be reverted:
```suggestion
export const OptionsListWorkingGroupOpening = React.memo(({ allOpenings, onChange }: Props) => (
  <OptionsListComponent>
```

thesan (changes requested):
It's just about coding style but on a project like this it's important to (as much as possible) only keep the changes the changes and the component properties that are needed

thesan (approved):
LGTM!

PR #3832, opened 2022-11-14
👿 Improvement on statistics layout

WRadoslaw:


thesan (approved):
Good idea the statistic wrapped weirdly

Issue #3651, opened 2022-10-03
Forum UX Upd: duplicated component


thesan:
## Context
Currently the codebase has 2 category cards:
- [`CategoryCard`](https://pioneer-2-storybook.vercel.app/?path=/story/forum-facelift-categorycard--default): Looks good but has no props so far
- [`CategoryListItem`](https://pioneer-2-storybook.vercel.app/?path=/story/forum-categories-categorylistitem--default): Functional but it is missing the subcategory on hover feature. Also AFAICS the story mocks members for no reason

and 2 thread cards:
- [`ThreadCard`](https://pioneer-2-storybook.vercel.app/?path=/story/forum-facelift-threadcard--default): Looks good but has no props so far
- [`ThreadItem`](https://pioneer-2-storybook.vercel.app/?path=/story/forum-threads-threaditem--default): Functional but it relies on `useThreadOriginalPost` internally which makes it slower and require mocks for tests/storybook. Also the some of the storybook controls don't work or are not relevant

## Scope
- [X] [`CategoryListItem`](https://pioneer-2-storybook.vercel.app/?path=/story/forum-categories-categorylistitem--default)
- [x] [`ThreadItem`](https://pioneer-2-storybook.vercel.app/?path=/story/forum-threads-threaditem--default)

(Part of #3029)

Issue #3352, opened 2022-07-13
Caching and Polling: Forum module


dmtrjsg:
## Context

As part of the overall caching and polling work (- https://github.com/Joystream/pioneer/issues/2311) Forum was left out of scope with a view to do it after the redesign work is done

This ticket addresses caching and polling for Forum

⚠️ Frequency is defined here: https://github.com/Joystream/pioneer/pull/3221

## Scope

⚠️ Do after the respective pages from #3029 are ~completed~ merged

* [ ] Create a hook to enable polling "useRefetchQueries" (already used in Bounty.tsx, packages/ui/src/app/pages/Council/Council.tsx and Election..)
* [ ] Implement caching and polling for the redesigned forum module

ivanturlakov:
Tested on https://dao-git-staging-joystream.vercel.app/#/settings?network-config=https://34.230.5.182.nip.io/network/config.json

⚠️ Latest threads didn't update
✅ Categories
✅ Watchlist
⚠️ Thread page didn't update
✅ Archived

PR #3833, opened 2022-11-14
🚀 Forum facelift

WRadoslaw:
Most of:
- #3029


thesan (approved):
:rocket: 

Only these 2 are left to close #3029:
- #3651
- #3352

PR #3845, opened 2022-11-15
Hide Bounties button on `MoveFundsModal`

traumschule:
This was raised in #pioneer-testing

### DIFF
--- a/packages/ui/src/accounts/modals/MoveFoundsModal/MoveFundsModalButtons.tsx
+++ b/packages/ui/src/accounts/modals/MoveFoundsModal/MoveFundsModalButtons.tsx
@@ -16,14 +16,18 @@ interface MoveFundsModalItemProps {
 export const MoveFundsModalButtons = ({ insufficientBalances, noFreeAccounts }: MoveFundsModalItemProps) => {
   const { showModal, hideModal } = useModal()
 
+  const BountiesButton = (
+    <div onClick={() => hideModal()}>
+      <LinkButtonPrimary size="medium" to={BountyRoutes.currentBounties}>
+        Bounties
+      </LinkButtonPrimary>
+    </div>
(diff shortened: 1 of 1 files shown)


freakstatic (approved):
LGTM!

oleksanderkorn (approved):

thesan:
If I remember correctly we decided to keep the button so people who wish to contribute would know of this option which should be available in the future. @dmtrjsg WDYT ?

dmtrjsg:
I think Theo has a point that's true, but either way would work really so I think hiding the button is actually good as it seems now it will take quite a while before it is back on.

packages/ui/src/accounts/modals/MoveFoundsModal/MoveFundsModalButtons.tsx
```diff
@@ -13,22 +13,29 @@ interface MoveFundsModalItemProps {
...
+  hide?: boolean
+}
+
+const LinkButton = ({ to, label, onClick, hide = false }: LinkButtonProps) => (
+  <div onClick={onClick}>
+    <LinkButtonPrimary size="medium" to={to}>
+      {label}
+    </LinkButtonPrimary>
+  </div>
+)
```
thesan, line 29: I don't see where is the `hide` used. Am I missing something ?

IMO just deleting the button or commenting it would be fine.

thesan (changes requested):

packages/ui/src/accounts/modals/MoveFoundsModal/MoveFundsModalButtons.tsx
traumschule: ab3ad24 finishes the intended implementation. Ready to remove though.

PR #3858, opened 2022-11-18
🌼 Considering blob in upload function

WRadoslaw:
- #3657


thesan (approved):
Turn out `atlas-services.joystream.org/avatars` doesn't support svgs. But I left the code there anyway I just removed it from the `SUPPORTED_IMAGES`.

Please review the changes in a comment

Issue #3527, opened 2022-08-18
Change tJoy to JOY


dmtrjsg:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202829081590251) by [Unito](https://www.unito.io)

polikosi:
Tested on https://pioneer-2-git-mainnet-joystream.vercel.app/
wss://34.230.5.182.nip.io/ws-rpc
✅I couldn't find tJoy in Pioneer UI

polikosi:
@traumschule @dmtrjsg  sorry, I find tJoy

I want to create a post with a zero account balance
![image](https://user-images.githubusercontent.com/80429191/206260196-796ebfaa-3461-41c7-a5f4-5d9042799e37.png)

traumschule:
Please test again on https://dao-git-fork-traumschule-fix-tjoy-joystream.vercel.app

thesan:
#3941 is now merged so it can be tested on: https://pioneer-2-git-mainnet-joystream.vercel.app/

polikosi:
✅fixed on: https://pioneer-2-git-mainnet-joystream.vercel.app/
![image](https://user-images.githubusercontent.com/80429191/206431687-f58329dc-7154-43dc-acc7-1532f0315e6a.png)

Issue #3400, opened 2022-07-28
Remove Testnet Banner


dmtrjsg:
Remove this banner:

![](https://user-images.githubusercontent.com/93924525/181470872-c0398c8a-78a3-49a3-8a2d-92f471ab55dd.png)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203143438951371) by [Unito](https://www.unito.io)

polikosi:
Tested on https://pioneer-2-git-mainnet-joystream.vercel.app/
wss://34.230.5.182.nip.io/ws-rpc

✅ Testnet banner doesn't appear

PR #3859, opened 2022-11-18
♾️  Initial mainnet changes

WRadoslaw:
- #3400
- #3527


thesan:
@WRadoslaw is this based on a version of `dev` which includes #3824 ? If not we should at least make sure that `mainnet` includes it

WRadoslaw:
> @WRadoslaw is this based on a version of `dev` which includes #3824 ? If not we should at least make sure that `mainnet` includes it

True, I've made sure that it was merged before I debranched from `dev`.

thesan (approved):

Issue #2491, opened 2022-03-07
[USERSNAP] `MaxWhiteListSize = 20` not enforced at creation


dmtrjsg:
**Taken at**: https://pioneer-2.vercel.app/#/proposals/preview/85
**Fields**:
- Houston, we have a problem.: `MaxWhiteListSize = 20` not enforced at creation
- Give us some details: I could add 21...
- -
- Discrod Handle: bwhm

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/975b6c0b-e700-4888-8e6b-0892dee343b4-annotated_a3ba18fa-0418-4b8f-82ea-d868c9f41d02.png?etag=97b1b78e6a32fb675ee2fdea9f9387c3)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/975b6c0b-e700-4888-8e6b-0892dee343b4-annotated_a3ba18fa-0418-4b8f-82ea-d868c9f41d02.png?etag=97b1b78e6a32fb675ee2fdea9f9387c3)*

**Browser**: Firefox 97 (macOS Catalina)
**Screen size**: 2859x1608
**Browser size**: 2144x1117
**Labels**: bug-non-critical
**[Open #100 in Usersnap](https://app.usersnap.com/l/feedback/5a8616fe-f133-4865-bc3c-f7777dde12a0?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

chrlschwb:
fixed this with PR 3861

PR #3861, opened 2022-11-18
Lead opening proposal whitelist limit (#2491)

chrlschwb:
Issue #2491.
When creating a WG lead opening, a whitelist limit of 20 was set, but the limit was not activated.
Created a new variable to implement this.
Error message "Maximum whitelist size of 20 members is reached." added.


packages/ui/src/proposals/modals/AddNewProposal/components/TriggerAndDiscussionStep.tsx
```diff
@@ -24,7 +26,12 @@ export const TriggerAndDiscussionStep = () => {
   ])
 
   const addMemberToWhitelist = (member: Member) => {
-    setValue('triggerAndDiscussion.discussionWhitelist', [...discussionWhitelist, member], { shouldValidate: true })
+    if (count < 20){
+      setValue('triggerAndDiscussion.discussionWhitelist', [...discussionWhitelist, member], { shouldValidate: true })
+      setCount(count+1)
+    }else{
+      setSelected(true)
+    }
```
traumschule, line 34: Please extend existing validation on the select field like [here](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/proposals/modals/AddNewProposal/components/TriggerAndDiscussionStep.tsx#L75), see also [machine.ts](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/proposals/modals/AddNewProposal/machine.ts)

traumschule (changes requested):
This looks ok on itself but we need to follow the implemented logic. Marking this as draft. Would add to description: Fixes #2491 (follow-up to #3855)
cc @WRadoslaw

WRadoslaw (changes requested):
Hey! Thanks for your contribution! 

About the issue, I would prefer to keep validation strictly in the form vendor. Please check out `packages/ui/src/proposals/modals/AddNewProposal/helpers.ts` line `173` I suppose. The discussion whitelist is already validated there, so it's only about adding an additional method on `yup`. This may come in handy -> https://github.com/jquense/yup#array

packages/ui/src/proposals/modals/AddNewProposal/components/TriggerAndDiscussionStep.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
```
WRadoslaw, line 1: This file do not require any changes

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
```diff
@@ -171,7 +171,7 @@ export const schemaFactory = (api?: ProxyApi) => {
       }),
       isDiscussionClosed: Yup.boolean(),
       discussionWhitelist: Yup.array().when('isDiscussionClosed', {
-        is: true,
+        is: Yup.array().length() < 20,
```
WRadoslaw, line 174:
```suggestion
        is: true,
```

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
```diff
@@ -171,7 +171,7 @@ export const schemaFactory = (api?: ProxyApi) => {
...
         then: Yup.array().required('Field is required'),
```
WRadoslaw, line 175:
```suggestion
        then: Yup.array().max(20, 'Maximum whitelist size of ${max} members is reached.').required('Field is required'),
```

WRadoslaw (changes requested):

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
chrlschwb: changed accordingly

traumschule:
There's too much wrong here, please recreate after testing locally. @gyroflaw please help out with pioneer setup.
- deleted yarn.lock
- added node as dependency
- removed react from imports
- not using prettier
```
15:01:06.689 | \| src/utils.ts(11,22): error TS2769: No overload matches this call.
15:01:06.689 | \| Overload 1 of 2, '(o: {}): string[]', gave the following error.
15:01:06.689 | \| Argument of type 'T' is not assignable to parameter of type '{}'.
15:01:06.689 | \| Overload 2 of 2, '(o: object): string[]', gave the following error.
15:01:06.689 | \| Argument of type 'T' is not assignable to parameter of type 'object'.
15:01:06.717 | \| error Command failed with exit code 2.
```

Issue #2596, opened 2022-03-18
Fix account selector markup (auto-fill address)


dmtrjsg:
![Screenshot 2022-03-18 at 17.45.23.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/755ac4bf-e5ab-409d-82d7-54ee8d42b529)

adovrodion:
Bounty link: https://dao.joystream.org/#/bounty/preview/52

adovrodion:
Bounty link: https://dao.joystream.org/#/bounty/preview/476

chrlschwb:
2596

traumschule:
Do we have influence over what browser suggests based on field name and saved suggestions? It's possible to change field name to avoid collisions.

PR #3880, opened 2022-11-23
Removed account info auto-fill from account selector (#2596)

chrlschwb:
#2596
Removed <AccountInfo> component from 
pioneer\packages\ui\src\accounts\components\SelectAccount\OptionAccount.tsx

### DIFF
--- a/packages/ui/src/accounts/components/SelectAccount/OptionAccount.tsx
+++ b/packages/ui/src/accounts/components/SelectAccount/OptionAccount.tsx
@@ -23,7 +23,7 @@ export const OptionAccount = ({ option, isForStaking }: Props) => {
 
   return (
     <>
-      <AccountInfo account={option} locked={isLocked} />
+      {/*<AccountInfo account={option} locked={isLocked} />  // Address info commented out */}
       <BalanceInfoInRow>
         <InfoTitle>{balanceType} balance</InfoTitle>
         <InfoValueWithLocks>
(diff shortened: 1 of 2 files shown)


traumschule:
#2596

traumschule:
This removed the account address. Positioning issue is unrelated.

missing-address

packages/ui/src/accounts/components/SelectAccount/OptionAccount.tsx
```diff
@@ -23,7 +23,7 @@ export const OptionAccount = ({ option, isForStaking }: Props) => {
 
   return (
     <>
-      <AccountInfo account={option} locked={isLocked} />
+      {/*<AccountInfo account={option} locked={isLocked} />  // Address info commented out */}
```
traumschule, line 26: This breaks account selection and does not fix the issue.

traumschule (changes requested):

oleksanderkorn (changes requested):
Worse to try `autocomplete="none"`:
image

Previous solution `autocomplete="new-password"` have been working but not anymore in the latest chrome.

There have been an ongoing discussion and this may still break in future versions:
https://stackoverflow.com/questions/15738259/disabling-chrome-autofill

traumschule:
I don't think this approach is the right solution. Feel free to reopen with a screenshot.

Issue #3885, opened 2022-11-24
Hide tooltip on locked accounts as soon as contents of scroll bar (accounts) start to get scrolled


dmtrjsg:
## Context

Tooltip is flipping when accounts are scrolled
![Screenshot 2022-11-24 at 17.02.06.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/b160edc3-deb8-4588-8ca0-29dafa569570)

## Scope
- [ ] Close tooltip as soon as user starts scrolling the accounts up or down

dmtrjsg:
@traumschule pls feel free to assign

WRadoslaw:
@oleksanderkorn This might fix the issue

```
const [isDetached, setIsDetached] = useState(false)

  const { styles, attributes, state } = usePopper(referenceElementRef, popperElementRef, {
    placement: isDetached ? 'top-start' : 'bottom-start',
    modifiers: [
      {
        name: 'offset',
        options: {
          offset: offset ?? [0, 0],
        },
      },
    ],
  })

  useEffect(() => {
    if (referenceElementRef && boundaryElement) {
      const refBox = referenceElementRef.getBoundingClientRect()
      const boundaryBox = boundaryElement.getBoundingClientRect()

      setIsDetached(boundaryBox.bottom + boundaryBox.y < refBox.bottom + refBox.y)
    }
  }, [!state, !boundaryElement])
  ```

PR #3888, opened 2022-11-24
Hide tooltip on locked accounts as soon as contents of scroll bar (accounts) start to get scrolled #3885

oleksanderkorn:
@WRadoslaw applied your suggestion, it is not flipping, but still not ideal, the tooltip link is hidden later than the tooltip itself, because of `transition: ${Transitions.all};` (resetting transition for opacity to 0s helps )

But the requirements are to hide on scroll is started.

Fixes: https://github.com/Joystream/pioneer/issues/3885

### DIFF
--- a/packages/ui/src/common/components/Tooltip/Tooltip.tsx
+++ b/packages/ui/src/common/components/Tooltip/Tooltip.tsx
@@ -57,9 +57,10 @@ export const Tooltip = ({
   const [referenceElementRef, setReferenceElementRef] = useState<HTMLElement | null>(null)
   const [popperElementRef, setPopperElementRef] = useState<HTMLDivElement | null>(null)
   const [boundaryElement, setBoundaryElement] = useState<HTMLElement | null>(null)
+  const [isDetached, setIsDetached] = useState(false)
 
-  const { styles, attributes } = usePopper(referenceElementRef, popperElementRef, {
-    placement: 'bottom-start',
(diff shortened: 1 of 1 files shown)


WRadoslaw:
Yeah, unfortunately I have no idea why solution based on the popper API would suddenly stop working. About the scope, hiding tooltip on scroll won't change a thing, because this problem occurs when drop-down is totally static

traumschule:
GM! if i try to make it look back, the window manager helps by resizing the window without the UI catching a rerender.

dropdown
tooltip

thesan:
@traumschule I think this only happen if the window size is reduced way passed what the app supports:


@dmtrjsg, @oleksanderkorn but I'm not sure about what the issue is about... Here's what I'm getting on ff:
[Screencast from 2022-12-12 17-45-15.webm](https://user-images.githubusercontent.com/6571453/207104891-c981210a-1bda-4f6e-a7b6-4d91e9c0600b.webm)

Is this behavior wrong ?

traumschule:
It's not great :)

thesan (changes requested):
This solution doesn't work for me ATM. The tooltips still scroll with the select options. I'll set this PR as a draft for now. I'll try to keep this in mind and add a suggestion later but rn I have no ideas.

@oleksanderkorn please update the PR if you think of something, but in the meantime I don't think this should be a priority.

cc @chrlschwb

Issue #3122, opened 2022-05-30
[USERSNAP] Forum: The 1st post of a thread only includes the blockheight and no date/time


dmtrjsg:
## Scope

- Show date/time of 1st post

____
**Taken at**: https://dao.joystream.org/#/forum/thread/168
**Fields**:
- Houston, we have a problem.: Forum: The 1st post of a thread only includes the blockheight and no date/time
- Comment: subsequent posts do include this information
- Feedback: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/49787688-d0c2-4798-b986-376bf433eb76-annotated_ae5586dc-159b-43d4-9e90-81590ccb7df8.png?etag=b80c3f264790e2bed5ef57d406cc69b7)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/49787688-d0c2-4798-b986-376bf433eb76-annotated_ae5586dc-159b-43d4-9e90-81590ccb7df8.png?etag=b80c3f264790e2bed5ef57d406cc69b7)*

**Browser**: Chrome 101 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1176x1050
**[Open #265 in Usersnap](https://app.usersnap.com/l/feedback/78272eff-66f5-4f41-9388-587cdb073e04?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

traumschule:
Just to be sure

polikosi:
tested on https://dao-git-carthage-joystream.vercel.app/#/settings?network-config=https://44.201.112.198.nip.io/network/config.json

a thread only includes the blockheight
![image](https://user-images.githubusercontent.com/80429191/199012823-6f421d4f-abea-4ef9-8ec6-710d016f2f2b.png)

dmtrjsg:
@traumschule pls feel free to assign for fix

polikosi:
✅tested on https://dao-git-staging-joystream.vercel.app/#/settings?network-config=https://34.230.5.182.nip.io/network/config.json
[...]

PR #3893, opened 2022-11-27
thread datetime

traumschule:
#3122

thread-time-end


dmtrjsg:
@traumschule I can see that WEST timezone is added to the originating thread.. is it hardcoded or taken from User browser settings? How is it showing for the other posts in a thread?

traumschule:
that's from `BlockTime` component using `formatDateString`.

oleksanderkorn (approved):

thesan (approved):

Issue #2807, opened 2022-04-06
Profile: Improve `Recoverable Balance` column


traumschule:
1. [ ] Sorting by recoverable balance has no effect / ignores vested tokens.
2. [ ] Show vested total and and claimable.
3. [ ] Add voting stake to total recoverable when no longer locked.
4. [ ] Clarify total recoverable.

## Part 4

I do not understand this screen, specifically `Recoverable Balance`. [Raised here](https://discord.com/channels/811216481340751934/910968995379822633/960484575312773151)
- It is not clear how each stake contributes to total recoverable balance. Why are only 8.7M recoverable?
- There should be a recover button for all stakes (possibly disabled) or a short explanation why it is not recoverable (eg. `locked for proposal`, `locked for candidate`, `activate nomination *Unbond*`, etc.).
- Tooltips for details if suitable.

![](https://cdn.discordapp.com/attachments/812344657794301972/960217105326833764/profile.png)

[Response](https://discord.com/channels/811216481340751934/910968995379822633/960830912256426074):
> "Do you know what does the sum for Recoverable Balance mean, 8M from what?"
> - recoverable balance is smth that you staked before but now the underlying process is over so you can recover (meaning transfer funds from module acc to your member's role or staking acc.)

traumschule:
Ref: https://github.com/Joystream/pioneer/issues/2649#issuecomment-1077862588

dmtrjsg:
This is been worked on right at the moment, feel free to check in on the progress of this task next week:  
- #2300

traumschule:
What is X: `200,000 + 5,533.38 - X = 200,562.46`
![cm-difference](https://user-images.githubusercontent.com/31551045/204184404-3f1c38d4-ebad-4809-8b43-8208c97cc972.png)

PR #3894, opened 2022-11-27
Clarify recoverable balance (#2807)

traumschule:
#2807

before / after (right)
locks-comparison

to triage: after 5eb7f8b `total` is not expected to be less than `vestingTotall`:
total-lower-than-vestingTotal

### DIFF
--- a/packages/ui/src/accounts/components/AccountItem/AccountItem.tsx
+++ b/packages/ui/src/accounts/components/AccountItem/AccountItem.tsx
@@ -6,6 +6,7 @@ import { AccountInfo } from '@/accounts/components/AccountInfo'
 import { AccountLocks } from '@/accounts/components/AccountLocks'
 import { TransferButton } from '@/accounts/components/TransferButton'
 import { useBalance } from '@/accounts/hooks/useBalance'
+import { addClaimable } from '@/accounts/model/sortAccounts'
 import { Account } from '@/accounts/types'
 import { DropDownButton, DropDownToggle } from '@/common/components/buttons/DropDownToggle'
 import { TableListItemAsLinkHover } from '@/common/components/List'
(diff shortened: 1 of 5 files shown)


traumschule:
The [failing test](https://github.com/Joystream/pioneer/actions/runs/3561791080/jobs/5983004572#step:7:150) was built with the assumption that a non-rivalrous lock will reduce the amount the voting lock can recover - it does not say 'amount that can be turned into free or unreserved balance".

recoverable-test

To illustrate comments in 71ca0dc according to the failing test the vote can only partially recover due to the unrecoverable "bound" stake. This may come down to a wording question but in this logic it wouldn't be possible to "recover" the whole amount and vote for someone else several elections later with the same amount (which would also be possible without recovering) .
However since both aren't rivals at least one can fully  "recover" and only the fraction exclusive the binding can be turned into "transferable" again - but this is not the meaning of "recoverable".

Since it only affects voting my suggestion is to not test vote recovery (always fully because not slashable) but something more interesting like
- how much can a validator recover after voting and getting slashed.
- there are edge cases when `total` gets lower than a stake (after slashes)
- how much can a 'total' vote recover after a rivalrous lock got slashed
- vesting balances can be higher than `total` or `total recoverable` (see above)
- both binding and proposal stake share the same tokens:
handbook
You can recover the proposal stake fully but it won't affect transferable balance. Am i alone with this understanding?

=> either only use "recoverable" in context to specific locks (according to rivalry)
=> on account level refer for the "capped" part to maximal free-able or amount that can be turned into transferable tokens, like 'free-after-recovery'.

Also worth noting that the handbook page on [locks](https://joystream.gitbook.io/testnet-workspace/system/accounts-and-staking#rivalry) does not mention "recover". :(

(feels like a total sidetrack like everything in this PR)

PR #3898, opened 2022-11-28
⬆️ Yarn update

WRadoslaw:


thesan (changes requested):
It looks good but we should avoid changes to the `@joystream` packages. So this would be the right moment to remove `@joystream/metadata-protobuf` and `@joystream/types` from this repos and let yarn get them from npm.

As a result I think `packages/chain-metadata.json` can be removed too. and the `build` script can become: `"build": "wsrun --package @joystream/pioneer -c build` or maybe even just `yarn workspace @joystream/pioneer build` I'm not sure what's the output difference between the 2. Hopefully this should be enough :crossed_fingers:

thesan (approved):
:100:

Issue #3073, opened 2022-05-16
Create Proposal for Channel Payouts


dmtrjsg:
## Context

Details are covered here:
- [ ] https://github.com/Joystream/pioneer/issues/2934

ℹ️ DM: in my head this shall broadly follow [this pattern](https://www.figma.com/file/pgNl83wgAub186krVkdiuo/Proposals?node-id=210%3A13897) for runtime upgrade, with some light touch validation thrown in, without going on an channel-specific level, since number of channels can be very (very-very) large ℹ️ 

### Key Challenges:
☝️ 1️⃣  _If a channel owner was promised e.g. 10 in a given proposal, but did not cash it out, then they could be promised any new value in the next one, say 5, and this would be a permanent loss without recourse. It is in other words at the discretion of the council to properly update what a channel is owed over time if the channel does not cash out. This is largely a moot point in practice, because it is to begin with at the discretion of the council what to pay out in any proposal regardless, at least as a technical matter.

☝️ 2️⃣  A cash-out proposal does not need to address what all channels are owed, it can only cover a subset, or event be empty. For anyone not mentioned, this is equivalent to being owed 0.

☝️  These parameters can also be changed in the proposal. While they can be changed they don't have to. E.g. member can only change one parameter and keep all else unchanged. The Payload is not mandatory for the proposal execution either.

<img width="959" alt="Screenshot 2022-05-18 at 15 29 51" src="https://user-images.githubusercontent.com/93924525/169066550-eee5eafc-9500-4602-8cdf-1a821f8ff821.png">

### User Story: as Pioneer Member I want to create proposal for Channel Payouts

⚠️ **assumption about operations:** Presumably proposal description will be pointing to the related Incentives programme rules and perhaps some sort of an External Tool for incentives scheme scoring inputs so proposal creator will have access to it.  

**Acceptance criteria:**

- [ ] Follows proposals pattern with general and specific parameters. Here, in this issue we only focus on the variation for the specific parameters
- [ ] Accepts JSON as input for the specific parameters
Proposal Specific Params:
ChannelID
Payout
Rationale
- [ ] Validation error on bad file format after uploading (Payout payload cannot be processed. Please check the file format and extension and try again.)
- [ ] ⚠️ importantly: Allows to update the parameters from point 3 in the context section, also r[eferenced here> ](https://github.com/Joystream/joystream/issues/3263)

>The only validation  that is technically feasible and ROI justified would be:
- [ ] Add parameter values

[...]

PR #3906, opened 2022-11-29
🤑 Channel payouts proposal

WRadoslaw:
- #3073 

:warning: once ready merge this PR into a new `ephesus` branch


packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/ChannelIncentivesPayout/ChannelIncentivesPayout.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const expectedDataSizeFee = useObservable(() => api?.query.storage.dataObjectPerMegabyteFee(), [api?.isConnected])
+  const expectedDataObjectStateBloatBond = useObservable(
+    () => api?.query.storage.dataObjectStateBloatBondValue(),
+    [api?.isConnected]
+  )
+
+  useEffect(() => {
+    if (expectedDataSizeFee && expectedDataObjectStateBloatBond) {
+      setValue('channelIncentivesPayout.expectedDataSizeFee', expectedDataSizeFee)
+      setValue('channelIncentivesPayout.expectedDataObjectStateBloatBond', expectedDataObjectStateBloatBond)
```
thesan, line 35: It's nice to have the `objectCreationParamsContent` and the `objectCreationParamsSize` validated separately, but to keep `getSpecificParameters` simple the payload object could be build here:
 ```suggestion
      setValue('channelIncentivesPayout.payload', {
        uploaderAccount: creator.controllerAccount,
        objectCreationParams: { size_: objectCreationParamsSize, ipfsContentId: objectCreationParamsContent },
        expectedDataSizeFee: expectedDataSizeFee,
        expectedDataObjectStateBloatBond: expectedDataObjectStateBloatBond,
      })
```

(with `const [objectCreationParamsSize, objectCreationParamsContent] = watch(['objectCreationParamsSize', 'objectCreationParamsContent'])`)

WDYT?

thesan (changes requested):

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/ChannelIncentivesPayout/ChannelIncentivesPayout.tsx
WRadoslaw: It is kinda validated separately. Both of them are validated at the time of upload. I choose that way because FileDropzone has a nicer validation UI than just plain red text. 

So at the time of setting values both size and params content are validated.

thesan: I got that but what I'm proposing is to keep these separate validations but also to have `channelIncentivesPayout.payload` (not validated) but set (has a react hook form data) inside the `useEffect` in  `ChannelIncentivesPayout` instead of inside `getSpecificParameters`

thesan:
@WRadoslaw more importantly the CI reports problems with the `@joystream/js` or the patch... I thought I left a comment about that yesterday but it looks like it got lost due to bad network :cry:

Issue #3909, opened 2022-11-30
[USERSNAP] can't create proposals


dmtrjsg:
**Taken at**: https://dao.joystream.org/#/proposals/current
**Fields**:
- Houston, we have a problem.: can&#39;t create proposals
- Comment: can&#39;t create proposals is we have many proposals approved and in the gracing period

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/f147eadb-c1c8-4703-bf7e-4395de381e12-annotated_d453cf05-ce34-4c58-85fe-b4b8ff2a6410.png?etag=7a5556d8f65885ed6775d6cfb7624b8f)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/f147eadb-c1c8-4703-bf7e-4395de381e12-annotated_d453cf05-ce34-4c58-85fe-b4b8ff2a6410.png?etag=7a5556d8f65885ed6775d6cfb7624b8f)*

**Browser**: Chrome 107 (macOS Catalina)
**Screen size**: 1920x1080
**Browser size**: 1720x761
**[Open #348 in Usersnap](https://app.usersnap.com/l/feedback/68504119-5192-400b-a0c2-d77ab7aa6dd9?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
@Lezek123 could you advise here pls on how the limitations are set up rn on the concurrent number of active proposals? (Suspect the limit covers both gracing and open)

<img width="660" alt="Screenshot 2022-12-02 at 19 52 19" src="https://user-images.githubusercontent.com/93924525/205332512-4452cdd5-c9e4-448d-bf21-c0f37be43ce0.png">

Lezek123:
@dmtrjsg The limit is 20 active proposals: https://github.com/Joystream/joystream/blob/master/runtime/src/lib.rs#L1571. _Active_ refers to proposals that haven't reached their final status yet. The number of active proposals can be either retrieved from the query node:
```graphql
query {
  proposalsConnection(where: { isFinalized_eq: false }) {
    totalCount
  }
}
```
Or from chain via `api.query.proposalsEngine.activeProposalCount`

[...]

PR #3916, opened 2022-12-02
Tooltip for deactivated `AddProposalButton` (#3909)

traumschule:
#3909

Wrap `TransactionButton` in a tooltip when disabled. The `placement` prop seemed necessary to better orient the arrow.

Note ideally we don't want to hardcode constants, however there is no quick solution with [useProposalConstants](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/proposals/hooks/useProposalConstants.ts) because it is designed to query `proposalsCodex` not `proposalsEngine.activeProposalCount` and can't be easily extended without changing [its return type](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/proposals/types/constants.ts).
Waiting for the go-ahead.


packages/ui/src/proposals/components/AddProposalButton.tsx
```diff
@@ -34,4 +35,21 @@ export const AddProposalButton = () => {
       Add new proposal
     </TransactionButton>
   )
+
+  if (!api) return <Tooltip tooltipText="Connecting to api">{txButton()}</Tooltip>
+
+  if (!areProposalSlotsAvailable)
+    return (
+      <Tooltip
+        tooltipTitle="MAX_ACTIVE_PROPOSALS"
```
oleksanderkorn, line 44: Wondering should we rename the title to  something like this?
```suggestion
        tooltipTitle="Max active proposals limit reached"
```

oleksanderkorn (changes requested):

oleksanderkorn (approved):

traumschule:
## before
```
FAIL test/proposals/components/AddProposalButton.test.tsx (8.094 s)
 |   ● UI: AddProposalButton › Active proposals under proposal limit
 | 
 |     expect(element).not.toBeDisabled()
 | 
 |     Received element is disabled:
 |       <button class="ButtonPrimaryStyles-sc-1s7nmh2-1 bOBZxM" disabled="" />
 | 
 |       17 |     renderComponent()
 |       18 |
 |     > 19 |     expect(await getButton('Add new proposal')).not.toBeDisabled()
```

## after
```
 PASS  test/proposals/components/AddProposalButton.test.tsx
  UI: AddProposalButton
    ✓ Active proposals under proposal limit (109 ms)
    ✓ Active proposals limit met (17 ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total

```

oleksanderkorn (approved):

thesan (changes requested):
That's a really nice solution but ATM the tooltip shows even when they're no current proposals:
image

on: https://dao-git-fork-traumschule-fix-add-proposal-button-3909-joystream.vercel.app/#/proposals/current?network-config=https://34.230.5.182.nip.io/network/config.json

packages/ui/src/proposals/components/AddProposalButton.tsx
```diff
@@ -1,37 +1,44 @@
...
+        tooltipTitle: 'Max active proposals limit reached',
+        tooltipText: `The creation of proposals is currently disabled because the number of deciding or gracing proposals reached the limit of ${maxProposals}.`,
+        tooltipLinkText: 'Proposal System Constants',
+        tooltipLinkURL: 'https://joystream.gitbook.io/testnet-workspace/system/proposal-system#constants',
+      }
+    return {
+      tooltipText: 'Use the proposal engine to suggest a change to the Council.',
+      tooltipLinkText: 'Learn about the Proposal System',
+      tooltipLinkURL: 'https://joystream.gitbook.io/testnet-workspace/system/proposal-system',
+    }
```
traumschule, line 33: Looks like a logic error. This might work:
```suggestion
    if (!api?.isConnected || availableSlots > 0) {
      return {
        tooltipText: 'Use the proposal engine to suggest a change to the Council.',
        tooltipLinkText: 'Learn about the Proposal System',
        tooltipLinkURL: 'https://joystream.gitbook.io/testnet-workspace/system/proposal-system',
      }
    } else {
      return {
        tooltipTitle: 'Max active proposals limit reached',
        tooltipText: `The creation of proposals is currently disabled because the number of deciding or gracing proposals reached the limit of ${maxProposals}.`,
        tooltipLinkText: 'Proposal System Constants',
        tooltipLinkURL: 'https://joystream.gitbook.io/testnet-workspace/system/proposal-system#constants',
      }
    }
```

traumschule:
Now the default is to show the general info with handbook link from the start even without connection or wallet. Clicking the button will open the modal to create a membership.
Should be fine because even new users will notice the `Connect Wallet` button at some point when asked to fill the fields.

proposals-tooltip

packages/ui/src/proposals/components/AddProposalButton.tsx
```diff
@@ -1,37 +1,46 @@
...
-      style="primary"
-      size="medium"
-      onClick={addNewProposalModal}
-      disabled={!areProposalSlotsAvailable}
-    >
-      <PlusIcon />
-      Add new proposal
-    </TransactionButton>
+    <Tooltip placement="bottom-end" {...tooltipProps}>
+      <TransactionButton style="primary" size="medium" onClick={addProposalModal} disabled={!availableSlots}>
```
thesan, line 40: `TransactionButton` also renders a [Tooltip when a transaction is pending](https://github.com/Joystream/pioneer/blob/6fb1e843fefeb4bf3a72674417821d0fab25e254/packages/ui/src/common/components/buttons/TransactionButton.tsx#L18). As a result in case of a pending transaction, this will result in the Tooltip beeing nested (I'm not sure what will actually be shown). To avoid this situation, `TransactionButton` could accept `TooltipContentProp`:
- If these are defined ⇒ display the Tooltip
- If a transaction is pending ⇒ display the: "Please wait until the current transaction is over" regardless of whether a different tooltip text was passed.

WDYT ?

thesan (changes requested):

packages/ui/src/proposals/components/AddProposalButton.tsx
```diff
@@ -1,37 +1,46 @@
...
-    [api?.isConnected]
-  )
-  const areProposalSlotsAvailable = api && maxProposals && currentProposals?.lt(maxProposals)
+  const maxProposals = useMemo(() => api?.consts.proposalsEngine.maxActiveProposalLimit, [api?.isConnected])
+  const proposals = useFirstObservableValue(() => api?.query.proposalsEngine.activeProposalCount(), [api?.isConnected])
+  const availableSlots = useMemo(() => Number(maxProposals) - Number(proposals), [proposals, maxProposals])
+
+  const tooltipProps = useMemo(() => {
+    if (!api?.isConnected) return { tooltipText: 'Connecting to api' }
+    if (!api?.isConnected || availableSlots > 0) {
```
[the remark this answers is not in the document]
traumschule: The new logic is that until connection is established and when there are slots the default tooltip is shown. Without the first condition a user would see the error "Max active proposals limit reached" during loading.

packages/ui/src/proposals/components/AddProposalButton.tsx
traumschule: This makes sense.

traumschule:
Merged in 3.9.0

Issue #2327, opened 2022-02-24
Update Working Group Budget


dmtrjsg:
- [ ] Update Working Group Budget
- If budget_update is non-negative, then this it must be no less than the council budget, otherwise the absolute value must be no less than the current group budget.

bwhm:
@dmtrjsg 
Note that there are some potential pitfalls here. The budget will often change in between the creation and potential execution of the proposal, for more than one reason:
1. The budget may be increased/decreased due to another proposal
   - This would be FAR outside the scope of Olympia, but this can be somewhat accounted for by checking if there are any other active (budget) proposals for this group.
2. The budget will change if there is a payout in between.
   - This change can be estimated, but as a group can hire/fire in between, it won't be correct.

AFAICT, the best we can do here is provide some information and warnings to the user.

**At creation**  
Show:
- the current budget 
   - implemented but incorrect values ref #2443
- the expected size of the (next) payout(s)
- the blockheight + expected time until the (next) payout(s)
Warnings:
- If the user wants to **decrease** - always warn "If the budget is less than this amount at attempted execution, this proposal will fail to execute, and the budget size will not be changed"
- If the user wants to **decrease** more than the current budget warn "Unless the budget is increase between now and attempted execution, this proposal will fail to execute, and the budget size will not be changed"
   - For both, feel free to make the warning better :)

**Proposals page**
Assuming the proposal is created, in the proposal preview page
show:
- the current budget
- expected value of the budget assuming it gets passed?
- if the proposal is to **decrease** more than the current budget warn "Unless the budget is increase between now and attempted execution, this proposal will fail to execute, and the budget size will not be changed"

dmtrjsg:
At creation
Show:

~the current budget for the WG~ ✅  - Already done
~the current budget for the Council~ - ✅  Already done
➕ 👉  the expected size of the (next) payout(s)  + blockheight + expected time until the (next) payout(s).  + message "Expected payouts can be impacted by hiring and firing decisions made before the date of payout."
> WG payouts are done at regular times, and cover Lead and Workers salary. 

➕  👉  Add _custom_ Warnings:
If the user wants to decrease - always warn "If the budget is less than this amount at attempted execution, this proposal will fail to execute, and the budget size will not be changed"

If the user wants to decrease more than the current budget warn "Unless the budget is increase between now and attempted execution, this proposal will fail to execute, and the budget size will not be changed"~ - we have this covered by generic message:


[...]

PR #3920, opened 2022-12-04
Set maximum WG budget increase amount (#2327)

traumschule:
#2327 

BN seems to have a limit of 10000000000000000.

### DIFF
--- a/packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
+++ b/packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
@@ -275,9 +275,7 @@ export const schemaFactory = (api?: ProxyApi) => {
     updateWorkingGroupBudget: Yup.object().shape({
       isPositive: Yup.boolean(),
       groupId: Yup.string().required('Field is required'),
-      budgetUpdate: BNSchema.test(moreThanMixed(0, 'Amount must be greater than zero.'))
-        .test(maxMixed(new BN(999999999999999), 'Amount must be less than 99999.', true))
-        .required('Field is required'),
+      budgetUpdate: BNSchema.test(moreThanMixed(0, 'Amount must be greater than zero.')).required('Field is required'),
     }),
     setInitialInvitationCount: Yup.object().shape({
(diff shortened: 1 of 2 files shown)


oleksanderkorn (approved):
LGTM

thesan (changes requested):
@traumschule here: https://github.com/Joystream/pioneer/issues/2327#issuecomment-1336443690

You identified a bug where a javascript number was passed to `asUpdateWorkingGroupBudget` instead of a `BN`, I'm not sure why the type system doesn't detect this, but it's probably because somehow the `fragment.amount` is typed `any`.

Here's an example of `BN` with values > $2^{53} – 1$ (`Number.MAX_SAFE_INTEGER`): https://codepen.io/thesan/pen/poKmpzb?editors=0012

That said on chain most amount are encoded as `u64` and currently the default  `maxAllowedValue` of `<TokenInput />` is `powerOf2(128)` so it should be changed to `powerOf2(64)`

thesan (changes requested):

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
```diff
@@ -275,7 +275,9 @@ export const schemaFactory = (api?: ProxyApi) => {
     updateWorkingGroupBudget: Yup.object().shape({
       isPositive: Yup.boolean(),
       groupId: Yup.string().required('Field is required'),
-      budgetUpdate: BNSchema.test(moreThanMixed(0, 'Amount must be greater than zero')).required('Field is required'),
+      budgetUpdate: BNSchema.test(moreThanMixed(0, 'Amount must be greater than zero.'))
+        .test(maxMixed(new BN(999999999999999), 'Amount must be less than 99999.', true))
+        .required('Field is required'),
```
traumschule, line 280:
```suggestion
      budgetUpdate: BNSchema.test(moreThanMixed(0, 'Amount must be greater than zero.')).required('Field is required'),
```

traumschule:
No longer needed.

Issue #3940, opened 2022-12-07
Update text on the Insufficient Funds


polikosi:
## Scope
- [ ] Translation  added to the JSON is there, but maybe there's a typo..

## Context
I want to create a post with a zero account balance
![image](https://user-images.githubusercontent.com/80429191/206260675-a9e9df6a-edcf-4773-b14a-e1fca52aa658.png)

traumschule:
Does this happen for both en / ru language settings? If not please test #3942

polikosi:
When I change the language, nothing happens, the page doesn't translate.
![image](https://user-images.githubusercontent.com/80429191/206432260-52fe8329-c00e-4faf-99c4-b04f00eca5ab.png)

traumschule:
Makes sense: #1984

thesan:
@polikosi this should have been be fixed by #3958

polikosi:
tested on: https://pioneer-2-git-mainnet-joystream.vercel.app/#/settings
wss://34.230.5.182.nip.io/ws-rpc
![image](https://user-images.githubusercontent.com/80429191/209336240-70248f74-8b6d-425c-bb5d-67deb706bfb9.png)

PR #3942, opened 2022-12-07
i18n: update ru

traumschule:
add missing keys (#3940)


oleksanderkorn (approved):

thesan:
@traumschule I'm closing this PR as:

1. I quietly fixed #3940 in #3958 last month.
2. About adding the empty keys, the way I understand it, i18n will fallback to `en` when the `ru` key doesn't exist. You can check this behavior by setting `localStorage.setItem('i18nextLng', 'ru')` and check the settings page for e.g. Therefore adding all these English value on the Russian `json` just creates noise.

PR #3946, opened 2022-12-08
Fix tooltips for Election Round and Minimal Stake

gyroflaw:


traumschule (commented):
Thanks! The inline `TooltipExternalLink` doesn't make the code easier to read but is according to specification.
For conformity let's use the same tooltip on `/working-groups/openings` ([OpeningDetails](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/working-groups/components/OpeningsList/Opening/OpeningDetails.tsx#L65)).

rival

`tooltipText="Minimum tokens free of rivalrous locks required as application stake to this role."`

It's a bargain which link is more important, in general i'd go with one link only.
AnnounceCandidacyModal has a tooltip also mentioning rivalrous locks without link.
There could be a hint with link in the handbook directly.

oleksanderkorn (approved):

thesan (approved):

PR #3948, opened 2022-12-08
Initialize `discussionPosts` before thread is loaded (#3236)

traumschule:
#3236


thesan (changes requested):
The same issue still happens on https://pioneer-2-git-fork-traumschule-fix-sort-propos-88f43d-joystream.vercel.app/#/proposals/preview/41

I'll give debugging this a try too

thesan:
@traumschule I just figured it out: `createdAtBlock` is a block object so `Comparator(...).string` won't work. But `createdAt` is a ISO 8601 date time string which I thing is sortable alphabetically.

#3965 should solve it so I'm closing this PR.

PR #3953, opened 2022-12-08
Sort by recoverable balance and show vesting total, claimable (#2807)

traumschule:
#2807 part 1+2

### DIFF
--- a/packages/ui/src/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem.tsx
+++ b/packages/ui/src/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem.tsx
@@ -9,7 +9,6 @@ import {
   DetailsName,
   LockWrapper,
   TitleCell,
-  ValueCell,
   VestingStyledDropDown,
 } from '@/accounts/components/AccountItem/components/styles'
 import { lockIcon } from '@/accounts/components/AccountLocks'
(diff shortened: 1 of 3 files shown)


oleksanderkorn (approved):
lgtm

packages/ui/src/accounts/components/AccountItem/components/VestingLockItem/VestingLockItem.tsx
```diff
@@ -38,9 +37,11 @@ export const VestingLockListItem = ({
           <DetailsName>Vesting</DetailsName>
         </TitleCell>
         {!isDropped && (
-          <ValueCell>
+          <>
+            <TokenValue value={locked.add(vested)} />
             <TokenValue value={locked} />
-          </ValueCell>
+            <TokenValue value={vested} />
+          </>
```
thesan, line 44: The current solution doesn't work:
Screenshot from 2022-12-21 17-23-31

This component is given the `DeriveBalancesVesting` which doesn't include `vestedClaimable` but:
Recoverable is `vestedClaimable` and what's actually locked is `locked - vested + vestedClaimable` (although there might be different way to get to these result) and finally the total should always be `locked`.

So one way to fix this is to pass `vestedClaimable` to the `VestingLockListItem`.

thesan (changes requested):

PR #3964, opened 2022-12-09
Hide Languages tab (#3955)

### LINKED ISSUE #3955: Hide Languages tab from Settings

traumschule:
Fixes #3955


oleksanderkorn (approved):
Just a thought, should we remove it instead of commenting out?
We can put it back after from git history when its needed.

thesan (approved):
@oleksanderkorn I agree that deleting the code would be better (and it would be easy to find back on a small file like this one). But then there are already a bunch of commented out code on the app so it's fine for now. Maybe we could clean those out later.

Issue #3974, opened 2022-12-13
Fix icon on the Polkadot Block Explorer Menu item


dmtrjsg:
## Context

Explorer menu tab needs to match the designs

Expected:
Default mode:
![Screenshot 2022-12-13 at 14.00.37.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/5b341ae2-093f-4c89-ad26-205dbba7b603)

On Hover:
![Screenshot 2022-12-13 at 14.00.46.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/0891fa65-0376-4ff8-ba40-aa72822abaee)

Currently:
![Screenshot 2022-12-13 at 13.08.43.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/713b6697-7076-4887-9e77-0a6ee197ef6f)

## Scope

- [ ] Update Icon styling to match Figma designs

[Link to Figma > ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=9977%3A407524&t=jcJpTbfVIxdkeUYw-1)

traumschule:
Please feel free to assign someone else.

vrrayz:
Please I'd like to take up this issue

PR #3979, opened 2022-12-13
Fix icon on the Polkadot Block Explorer Menu item #3974

oleksanderkorn:
- #3974
image
image

### DIFF
--- a/packages/ui/src/common/components/page/Sidebar/NavigationLink.tsx
+++ b/packages/ui/src/common/components/page/Sidebar/NavigationLink.tsx
@@ -119,7 +119,21 @@ const NavigationItemLinkChildren = styled.div`
   ${Overflow.FullDots};
 
   svg {
-    width: 20px;
+    width: 16px;
+    height: 16px;
+
+    path {
+      fill: ${Colors.Black[400]}!important;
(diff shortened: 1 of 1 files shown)


traumschule (approved):
:ok_hand:

traumschule:
On a closer look
1. The link icon should be aligned right inside the bar.
2. Tooltip is currently not implemented.

traumschule:
@thesan i suggest to merge this good fix and let @vrrayz do the fine-tuning.

packages/ui/src/common/components/page/Sidebar/NavigationLink.tsx
```diff
@@ -119,7 +119,21 @@ const NavigationItemLinkChildren = styled.div`
   ${Overflow.FullDots};
 
   svg {
-    width: 20px;
+    width: 16px;
+    height: 16px;
+
+    path {
+      fill: ${Colors.Black[400]}!important;
```
thesan, line 126: @oleksanderkorn does [`currentcolor`](https://css-tricks.com/currentcolor/) so the hover could be removed ?
```suggestion
      fill: currentcolor !important;
```

thesan (approved):

Issue #3913, opened 2022-12-02
Forum: Category Page Layout  > Subcategories


dmtrjsg:
## Context
![Screenshot 2022-12-02 at 13.02.16.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/418c0111-e652-4448-8f69-b71363a6a438)

## Scope
- [ ] Review the merged PRs and see why the category page layout doesn't match the designs. Possibly apply relevant fixes if needed.

dmtrjsg:
@traumschule ready to be assigned, would be nice to get this one out within the Forum earlier than other forum tasks if possible ofc. You to decide ultimately 👍

traumschule:
wrong tab, sorry :)

mkbeefcake:
Does it mean i don't work on this issue?

traumschule:
just the prior actions. i'd like you to have a look if you feel comfortable but feel free to unassign.

mkbeefcake:
Ok, i will look this.

polikosi:
Tested on: https://pioneer-2-a87ruv6qh-joystream.vercel.app/#/settings?network-config=https://34.230.5.182.nip.io/network/config.json
![image](https://user-images.githubusercontent.com/80429191/215534993-070c54ae-000e-4fba-8c19-03ec7cdcf93f.png)

traumschule:
https://github.com/Joystream/pioneer/pull/3984#pullrequestreview-1236261315
> display the categories in a grid, just like how they look on the forum home page:
![](https://user-images.githubusercontent.com/6571453/210615206-5607ee54-da3e-4c90-b8e2-2fb037727562.png)
> Not in a in a slider.
> As a result a some components will become unused: <ForumCategoryList/> <CategoryListItem />, and maybe some others... Please remove them in this PR (as well as there stories if need).

PR #3984, opened 2022-12-15
fixed subcategory layout

mkbeefcake:
Fixed the issue
https://github.com/Joystream/pioneer/issues/3913
Screenshot_3

### DIFF
--- a/packages/ui/src/app/pages/Forum/ForumCategory.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumCategory.tsx
@@ -5,6 +5,7 @@ import styled from 'styled-components'
 import { ForumThreadOrderByInput } from '@/common/api/queries'
 import { TransactionButton } from '@/common/components/buttons/TransactionButton'
 import { EmptyPagePlaceholder } from '@/common/components/EmptyPagePlaceholder/EmptyPagePlaceholder'
+import { HorizontalScroller } from '@/common/components/HorizontalScroller/HorizontalScroller'
 import { PlusIcon } from '@/common/components/icons/PlusIcon'
 import { ItemCount } from '@/common/components/ItemCount'
 import { Loading } from '@/common/components/Loading'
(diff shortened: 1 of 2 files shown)


packages/ui/src/app/pages/Forum/ForumCategory.tsx
```diff
@@ -99,13 +101,15 @@ export const ForumCategory = () => {
...
-              <ItemCount count={category.subcategories.length}>
-                {isArchive ? 'Archived categories' : 'Categories'}
-              </ItemCount>
-              <ForumCategoryList categories={category.subcategories} isArchive={isArchive} />
+              <HorizontalScroller
+                title={isArchive ? 'Archived categories' : 'Categories'}
+                items={category.subcategories.map((category) => (
+                  <CategoryCard key={category.id} category={category} />
+                ))}
+                count={category.subcategories.length}
```
traumschule, line 109: design shows 2

traumschule (approved):
there are differences for category slider and thread row spacing but i'd let it pass.

design
design

after
subcat

packages/ui/src/app/pages/Forum/ForumCategory.tsx
mkbeefcake: I fixed this problem and updated it. please check it again.

traumschule (approved):
is according to design but depending on category titles and descriptions it looks pretty empty. maybe 3 or 4 will fit as well. leaving this up to further review. thanks @mkbeefcake!

mkbeefcake:
My pleasure

oleksanderkorn (approved):

thesan (changes requested):
Sorry it wasn't very clear in the design (because it shows only 2 categories) but what we meant was to display the categories in a grid, just like how they look on the forum home page:
image
Not in a in a slider.

As a result a some components will become unused: `<ForumCategoryList/>` `<CategoryListItem />`, and maybe some others... Please remove them in this PR (as well as there stories if need).

traumschule:
The horizontal slider doesn't make it easier to view and navigate through categories. I agree grid layout is preferable.

PR #3993, opened 2022-12-22
added refreshcount to prevent reload

### LINKED ISSUE #3968: Proposal discussion auto-scroll distraction
Triaging #3967 i noticed an auto-scroll behavior that keeps pulling back to the post every few seconds after clicking the 'replies to' link and scrolling further.

It should be fine to scroll only once.

mkbeefcake:
fix #3968 
The page is periodically refreshed by caused from fetching queries. 
So I added refreshcount query variable for preventing it.

### DIFF
--- a/packages/ui/src/proposals/components/ProposalDiscussions.tsx
+++ b/packages/ui/src/proposals/components/ProposalDiscussions.tsx
@@ -1,5 +1,5 @@
 import { ForumPostMetadata } from '@joystream/metadata-protobuf'
-import React, { RefObject, useMemo, useRef } from 'react'
+import React, { RefObject, useMemo, useRef, useEffect, useState } from 'react'
 import { generatePath } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


oleksanderkorn (approved):

thesan (changes requested):
@mkbeefcake sorry I missed that when triaging the issue but it looks like #3968 is just a special case of #4129.

This solution doesn't address the underlying issue which should be located in `PostListItem`.

Could you try to remove the `ref.current` in the `useEffect` dependencies [here](https://github.com/Joystream/pioneer/blob/0d308b26de798e8d41b8a7680b18b395ec42f03e/packages/ui/src/forum/components/PostList/PostListItem.tsx#L73), and if it solves the problem make sure it doesn't break something else. If it doesn't solve the problem properly stringifying the location should do the trick.

@chrlschwb it looks like someone is already working on #4129 could you coordinate the efforts with the assigned developer ?

PR #3996, opened 2022-12-24
Fix icon on the Polkadot Block Explorer Menu item #3974

vrrayz:
Fixed Explorer Menu Item to match with the Figma Design from issue #3974

Screenshot (441)


packages/ui/src/common/components/icons/symbols/LinkSymbol.tsx
```diff
@@ -8,22 +8,27 @@ import { SymbolProps } from './common'
...
+      /> */}
       <path
         className="primaryPart"
-        d="M15.4999 1.09998H21.9999L22.8999 1.99998V8.49998H21.0999V4.17277L10.6363 14.6364L9.36353 13.3636L19.8271 2.89998H15.4999V1.09998Z"
+        d="M9 1H8V3H9H11.5858L5.29289 9.29289L4.58579 10L6 11.4142L6.70711 10.7071L13 4.41421V7V8H15V7V2L14 1H9ZM2 3L1 4V14L2 15H12L13 14V11V10H11V11V13H3V5H5H6V3H5H2Z"
         fill={color ?? Colors.Blue[500]}
+        fillRule="evenodd"
+        clipRule="evenodd"
       />
+      {/* <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
traumschule, line 29: if this solution is accepted the comment should be removed

traumschule (commented):
lgtm

traumschule (approved):

oleksanderkorn (changes requested):
`LinkSymbol` is used in many more places, are you sure these changes will not break it anywhere?

Also i see commented code which should be either  uncommented or removed if not needed. 
Example of the usage in another place, which is now changed:
Before:
image
After:
image

vrrayz:
> `LinkSymbol` is used in many more places, are you sure these changes will not break it anywhere?
> 
> Also i see commented code which should be either uncommented or removed if not needed. Example of the usage in another place, which is now changed: Before: image After: image

Thanks for the review., i'll check it and get back to you asap

vrrayz:
Okay @oleksanderkorn and @traumschule , I have made the fix as suggested. please review again

Issue #4011, opened 2023-01-04
Regression on `UpdateWorkingGroupBudgetProposal`


traumschule:
App crash for https://pioneerapp.xyz/#/proposals/preview/26

```
Uncaught Error: Assertion failed
    at i (bn.js:6:21)
    at r._initNumber (bn.js:128:7)
    at r._init (bn.js:82:19)
    at new r (bn.js:39:12)
    at Object.UpdateWorkingGroupBudgetProposalDetails (ProposalDetails.ts:225:11)
    at n.asProposalDetails (ProposalDetails.ts:363:31)
```
>   amount: new BN(fragment.amount),

However amount is not empty. Either related to loading states or QN issue?
```
"data": {
    "proposalByUniqueInput": {
      "id": "26",
      "details": {
        "__typename": "UpdateWorkingGroupBudgetProposalDetails",
        "amount": 16500000000000000
      }
```
https://graphql-console.subsquid.io/?graphql_api=https://query.joystream.org/graphql
```
query MyQuery {
  proposalByUniqueInput(where: {id: "26"}) {
    id
    details {
      __typename
      ... on UpdateWorkingGroupBudgetProposalDetails {
        __typename
        amount
      }
    }
  }
}
```

PR #4012, opened 2023-01-04
Hotfix for #4011

traumschule:
Catch empty amount #4011

Fixes: https://dao-git-fork-traumschule-patch-2-joystream.vercel.app/#/proposals/preview/26


thesan:
@traumschule I'm closing this one in favor of #4018 for the reason I explained there.

But also #4018 as the advantage of showing the correct `Current Council Budget`: https://dao-git-fix-proposal-view-crash-joystream.vercel.app/#/proposals/preview/26 (although there could be in some cases a very small loss of precision ref  Joystream/joystream#4503).

That said I am confused by the fact that adding a `?? 0` prevents the crash since once the QN returns the value, `fragment.amount` is not `undefined` or `null` and its value is `> Number.MAX_SAFE_INTEGER` :confused:

PR #4017, opened 2023-01-05
🔥 Hotfix on proposal preview

WRadoslaw:

PR #3995, opened 2022-12-23
🩹 Fix proposal view crashing due to high amounts of JOY

thesan:
- #3994

https://dao-git-fix-proposal-view-crash-joystream.vercel.app/#/proposals/preview/12?network-config=https://atlas-next.joystream.org/network/config.json

Issue #3994, opened 2022-12-22
Update WG budget proposal details page shows blank


dmtrjsg:
Env: https://dao-git-ephesus-joystream.vercel.app/#/proposals/current

Proposal: https://dao-git-ephesus-joystream.vercel.app/#/proposals/preview/12

![Screenshot 2022-12-22 at 16.23.27.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/211e3576-eef0-4bed-a7a0-195034a6a339)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203601281548392) by [Unito](https://www.unito.io)

traumschule:
I believe there is a problem when no lead is set. The WG should be hidden from the selection.

![ephesus-wg-select](https://user-images.githubusercontent.com/31551045/209198792-3935ba8b-a90f-4ad8-a7ba-a2afe1c31917.png)

![ephesus-leads](https://user-images.githubusercontent.com/31551045/209198070-76386d7e-eed2-4aee-a926-f4f6b2a7457a.png)

However the proposal shows all details: https://dao-git-ephesus-joystream.vercel.app/#/proposals/preview/5
It is rejected as expected since there's no council.

![proposal5](https://user-images.githubusercontent.com/31551045/209198467-db92aead-d5c5-4d6e-b9cb-5b86370ee3e0.png)

thesan:
So this comes from the QN returning the update amount as a `number`. The problem is that the amount is `50000000000000000` which is (one digit) greater than `Number.MAX_SAFE_INTEGER` as a result the `bn.js` throws an error.

Try this on https://atlas-next.joystream.org/query-node/server/graphql:
```gql
query {
  proposals (where: { details_json: { isTypeOf_eq: "UpdateWorkingGroupBudgetProposalDetails" } }) {
    details {
      ...on UpdateWorkingGroupBudgetProposalDetails { amount }
    }
  }
}
```

[...]

PR #4018, opened 2023-01-05
🩹 Fix proposal view crashing due to high amounts of JOY on main

thesan:
- #3994
- #4011

@traumschule my bad because I wasn't expecting that this problem would occur so soon on `mainnet` I had merged #3995 on `dev`.

To avoid conflicts and because this addresses the issue on all proposals I'll re-merge this branch on `main` rather than #4012 or #4017.

Testable here: https://dao-git-fix-proposal-view-crash-joystream.vercel.app/#/proposals/preview/26

### DIFF
--- a/packages/ui/src/common/utils.ts
+++ b/packages/ui/src/common/utils.ts
@@ -1,3 +1,5 @@
+export * from './utils/bn'
+
 import { Reducer } from './types/helpers'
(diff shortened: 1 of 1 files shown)


thesan:
Hmm I need to rebase it...

Issue #4029, opened 2023-01-07
[USERSNAP] Forums: Reply view should start at 0, not 1


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/forum
**Fields**:
- Houston, we have a problem.: Forums: Reply view should start at 0, not 1
- Comment: default view shows all threads with 1 reply when they haven&#39;t been replied to yet. It seems to be counting the initial post (thread post) as being a reply when it isn&#39;t.

Should view as 0
- Feedback: tomato#9464]

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/39f2f44c-f9f6-4053-a858-679f98c0835d-annotated_8a49ff02-e203-4d73-881a-ed66e1c221d6.png?etag=57bef7bae119be37e84344f3d8636a9f)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/39f2f44c-f9f6-4053-a858-679f98c0835d-annotated_8a49ff02-e203-4d73-881a-ed66e1c221d6.png?etag=57bef7bae119be37e84344f3d8636a9f)*

**Browser**: Chrome 108 (macOS Catalina)
**Screen size**: 1440x900
**Browser size**: 1227x665
**[Open #363 in Usersnap](https://app.usersnap.com/l/feedback/5be452bc-206c-420a-8ff6-6d534d224e0b?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

ivanturlakov:
✅ Tested on https://dao-git-staging-joystream.vercel.app/#/forum
wss://34.230.5.182.nip.io/ws-rpc

Reply counter is hidden if there are 0 replies and starts from 1

<img width="1051" alt="Снимок экрана 2023-02-15 в 17 41 30" src="https://user-images.githubusercontent.com/46903215/219077823-1bef1f6e-346d-485f-9208-73d8ba590422.png">

PR #4030, opened 2023-01-07
Start replies count at zero and hide if empty (#4029)

traumschule:
#4029

counter


mochet:
@traumschule I would think it better to always show the reply counter even if it has 0 replies for consistency sake?

oleksanderkorn (approved):

thesan (approved):
:100:

Issue #4041, opened 2023-01-09
[USERSNAP] Wrong values in "My Roles" tab


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/working-groups/my-roles
**Fields**:
- Houston, we have a problem.: Wrong values in &#34;My Roles&#34; tab
- Comment: see image

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/95b67b59-e69c-487d-aa29-32488b3e9d03-annotated_86abc92b-213f-4c2a-8273-4c7671892ccc.png?etag=e88c6e6d10b1da3541d78daa4f3e7177)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/95b67b59-e69c-487d-aa29-32488b3e9d03-annotated_86abc92b-213f-4c2a-8273-4c7671892ccc.png?etag=e88c6e6d10b1da3541d78daa4f3e7177)*

**Browser**: Chrome 108 (Windows 10)
**Screen size**: 1600x900
**Browser size**: 1366x657
**[Open #367 in Usersnap](https://app.usersnap.com/l/feedback/e5ee8320-7913-4aba-9fcc-44ae4e96b8fa?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

thesan:
@traumschule please keep the `high-prio` labels, even after a PR has been opened. It makes it easier to prioritize issues during the review and QA phase too.

polikosi:
tested on: https://dao-git-fork-chrlschwb-fix-4041-myrolesvalues-joystream.vercel.app/#/settings?network-config=https://54.205.24.16.nip.io/network/config.json

it's ok for me
![image](https://user-images.githubusercontent.com/80429191/220595096-96ae2f30-365c-4dfc-9e58-25c91bd5ebca.png)

PR #4056, opened 2023-01-10
Fix wrong values in MyRoles tab (#4041)

chrlschwb:
issue #4041

Earned total, Next payment in, 
Total owed reward values have problem on QN side


packages/ui/src/working-groups/components/Roles/RolesList.tsx
```diff
@@ -31,18 +35,28 @@ export interface RolesListProps {
...
-      </ListItem>
-    ))}
-  </List>
-)
+export const RolesList = ({ workers }: RolesListProps) => {
+  const blockNumber = useCurrentBlockNumber()
+  const { api } = useApi()
+  const nextPayout = useMemo(
+    () => blockNumber && getNextPayout(workers, blockNumber, api),
+    [workers.length, blockNumber?.toNumber()]
```
traumschule, line 43: 'Next payment in' 0 JOY => leave box on `RolesListItem` empty if zero
That total paid is zero may be unrelated?

bwg-job

I suggest to rename the (passed) variable to `nextPayment` and adjust `RolesListItem` accordingly.

traumschule (changes requested):
Better already, still a few remarks. In general i'd question the use of next payment date because it's unlikely someone will actually sit there and wait for it.

`TOTAL OWED REWARD` seems still outdated? Hard to tell without a breakdown. Details in a tooltip would be great but out of scope.

packages/ui/src/working-groups/components/Roles/RolesList.tsx
```diff
@@ -80,13 +97,14 @@ const RolesListItem = ({ worker }: { worker: Worker }) => {
...
-            <TokenValue value={worker.rewardPerBlock} />
+            <TokenValue value={earnings} />
           </TextInlineBig>
           <ToggleableSubscriptionWide>Earned total</ToggleableSubscriptionWide>
         </OpenItemSummaryColumn>
         <OpenItemSummaryColumn>
           <TextInlineBig>
-            <TokenValue value={earnings} />
+            <NumericValue>{payout?.gte(BN_ZERO) ? payout?.toString() + ' blocks' : '–'}</NumericValue>
+            {/* {payout && <TextMedium lighter>({payout && blocksToTime(payout)})</TextMedium>} */}
```
traumschule, line 107: [getNextPayout](https://github.com/Joystream/pioneer/pull/738/commits/1dd175f01cb2749f5c7e1bff01a60c5628b6c45f) returns BN ("Big Number"). WDYT about renaming `payout` to `blocks` to make the code more clear?

We expect "payout" to be defined. Why the switch to `NumericValue` and then pass `-`? No need to re-implement [blocksToTime](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/common/model/blocksToTime.ts):
```suggestion
            <TextMedium lighter>{blocksToTime(payout)}</TextMedium>
```

traumschule (changes requested):
In terms of code readability `payout` could also mean a JOY value.

`TOTAL OWED REWARD` is usually outdated once non-zero even if filled afterwards.

packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRoles.tsx
```diff
@@ -17,9 +17,10 @@ import { useMyWorkers } from '@/working-groups/hooks/useMyWorkers'
 import { WorkingGroupsTabs } from '../components/WorkingGroupsTabs'
 
 export const MyRoles = () => {
-  const { isLoading, workers } = useMyWorkers()
+  const { isLoading, workers } = useMyWorkers();
+
+  const owedReward = workers.reduce((a, b) => a.add(new BN(b.owedReward)), new BN(0));
```
traumschule, line 22: > `TOTAL OWED REWARD`
> 21,997.53

should be 0

traumschule (approved):
lgtm, except for owed. let's merge and address that in separate PR.

gyroflaw (approved):
LGTM 👍

packages/ui/src/working-groups/components/Roles/RolesList.tsx
thesan: I guess it can be `undefined` when `blockNumber` is `undefined` (before the api connection/ before the block height is returned).

But I took a look at `getNextPayout` this function is weird... I don't see why it should return a `BN` to begin with (it's an amount of blocks it won't go over any js limits). It returns `BN(-1)` when no worker are passed :confused: and the reduce mutates it's accumulator. I think: `Math.max(0, Math.min(...nextPayoutPerGroup))` should work (given `nextPayoutPerGroup` becomes `number[]`)

thesan (changes requested):
@chrlschwb I agree with @gyroflaw it looks like the changes on:
- packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRoles.tsx
- packages/ui/src/working-groups/components/NextPayoutStat.tsx
- packages/ui/src/working-groups/hooks/useMyWorkers.ts
Are not necessary.

Especially the changes on `packages/ui/src/app/pages/WorkingGroups/MyRoles/MyRoles.tsx`, semicolons should be removed by the linting. Are you skipping the post commits hooks ?

chrlschwb:
BN is not BigNumber. BN is a class value of  "pioneer/node_modules/@types/bn.js/index.d.ts". So I can't calculation min and max value of "Math.max(0, Math.min(...nextPayoutPerGroup))". Math.min is library of Javascript. So member of Math is number. but nextPayoutPerGroup is BN.So I can't it.
But based on BN I've already done what you're trying to do.

thesan:
> BN is not BigNumber. BN is a class value of "pioneer/node_modules/@types/bn.js/index.d.ts". So I can't calculation min and max value of "Math.max(0, Math.min(...nextPayoutPerGroup))". Math.min is library of Javascript. So member of Math is number. but nextPayoutPerGroup is BN.So I can't it. But based on BN I've already done what you're trying to do.

Yes this is why I said: "given `nextPayoutPerGroup` becomes `number[]`)" It's not mandatory but it would be nice because `nextPayoutPerGroup` represent blocks (interval or height I don't remember rn), so there is no reason `BN` should be used here (if js numbers are good enough to store ms timestamps, they are good enough for block heights).

packages/ui/src/working-groups/model/getNextPayout.ts
```diff
@@ -12,7 +12,7 @@ export function getNextPayout(workers: Pick<Worker, 'group'>[], blockNumber: BN,
   const nextPayoutPerGroup = userGroups.map((name) => api?.consts[name].rewardPeriod.toBn()).map(blocksUntilNext)
 
   if (nextPayoutPerGroup.length) {
-    return nextPayoutPerGroup.reduce((closest, time) => (closest = BN.min(closest ?? BN_ZERO, time ?? BN_MILLION)))
+    return nextPayoutPerGroup.reduce((closest, time) => (closest = BN.min(closest ?? BN_ZERO, time ?? BN_MAX_INTEGER)))
   }
   return new BN(-1)
 }
```
thesan, line 18: Now that you edited this file. Here's what I'm proposing:
```ts
const nextPayoutPerGroup = userGroups.flatMap((name) => {
  const interval = api?.consts[name].rewardPeriod.toNumber()
  return interval ? [interval - (blockNumber % interval)] : []
}

if (nextPayoutPerGroup.length === 0) {
  return
}

return Math.min(...nextPayoutPerGroup)
```
Of course the returned type is now different so a few adjustment are necessary but at least the values should always make sense now.

thesan (changes requested):

packages/ui/src/working-groups/model/getNextPayout.ts
chrlschwb: I changed it the way you do.

traumschule:
`Earned in Past 24 hours` shows amount of my first role, not the total. Please check.

packages/ui/src/working-groups/model/getNextPayout.ts
thesan: @chrlschwb I really think the `getNextPayout` function should be rewritten, this code is confusing. However it's not really related to this PR which appear to fix what it's supposed to fix.

So please just revert the changes on this file, because although they are both wrong at least `BN_MILLION` blocks makes more sense than `BN_MAX_INTEGER` blocks. I'll fix this function in another PR.

thesan (changes requested):
@chrlschwb please check:

> `Earned in Past 24 hours` shows amount of my first role, not the total. Please check.

chrlschwb:
> `Earned in Past 24 hours` shows amount of my first role, not the total. Please check.

Can you please show me screenshot of the problem? I cannot test because I have only one role.
You are not commenting about the CM rewards, I hope. CM rewards are not integrated here.

traumschule:
24h

ivanturlakov:
Tested on https://dao-git-fork-chrlschwb-fix-4041-myrolesvalues-joystream.vercel.app/#/working-groups/my-roles
wss://54.205.24.16.nip.io/ws-rpc

@chrlschwb numbers are looks correct

### User hired to 2 roles
Снимок экрана 2023-03-01 в 16 37 45

### First payout
Снимок экрана 2023-03-02 в 13 37 10

### Second payout
Снимок экрана 2023-03-03 в 14 11 11

thesan (changes requested):
@chrlschwb looks like it go lost in the long thread but I replied to you [here](https://github.com/Joystream/pioneer/pull/4056#discussion_r1117684541) that the PR is good but please just reverse the changes on `packages/ui/src/working-groups/model/getNextPayout.ts`

traumschule (changes requested):
Value for 24h is still [wrong](https://github.com/Joystream/pioneer/pull/4056#issuecomment-1448426361). Please revise and remove unrelated changes (`.env.example`, `NextPayoutStat.tsx`).

> Total owed reward values have problem on QN side

Is there a QN issue about this?

traumschule:
@ivanturlakov did you test it with multiple roles or just one?

packages/ui/.env.example
```diff
@@ -1,31 +0,0 @@
-# TESTNET Endpoints
```
thesan, line 1: Thank you for removing some of the unnecessary changes but now this file was removed for no reason. Please restore it.

thesan (changes requested):

ivanturlakov:
@traumschule I [tested](https://github.com/Joystream/pioneer/pull/4056#issuecomment-1453454799) with two roles

traumschule:
Well [it's](https://github.com/Joystream/pioneer/pull/4056#pullrequestreview-1335038481) still wrong. 
> EARNED IN PAST 24 hours 
> 2,796.57
>
> Distribution Lead
> Reward per 14490 blocks
> 2,796.57

packages/ui/.env.example
chrlschwb: fixed it, please review

thesan (approved):
@traumschule IMO if this PR fixes for sure the values for the one role case, it's already a good improvement and worth merging. Please create a new issue for the multiple role case.

PR #4061, opened 2023-01-11
Fixed enhance election lifecycle issue

### LINKED ISSUE #4006: Enhance election lifecycle clarity
# Background

Currently, the election works like this

```
Announcing Period: 14 days
Voting Period: 3 days
Reveal Period: 3 days
Idle Period: 1 days
Total: 21 days = 3 weeks
```

However, when you land on the election screen, you jus see the current period and how much time is left. This makes it hard for people to understand the full cycle, in terms of duration and structure, and this deters participation from both prospective candidates and voters.

# Proposal

Add some sort of visual representation to this screen which solves this.

# Solution

image

mkbeefcake:
Fixed #4006

### DIFF
--- a/packages/ui/src/app/pages/Election/Election.tsx
+++ b/packages/ui/src/app/pages/Election/Election.tsx
@@ -26,6 +26,7 @@ import { useCurrentElection } from '@/council/hooks/useCurrentElection'
 import { useElectionStage } from '@/council/hooks/useElectionStage'
 import { Election as ElectionType } from '@/council/types/Election'
 
+import { ElectionPeriod } from './components/ElectionPeriod'
 import { ElectionTabs } from './components/ElectionTabs'
 
 const displayElectionRound = (election: ElectionType | undefined): string => {
@@ -108,6 +109,15 @@ export const Election = () => {
           <TextHuge id="election-round-value" bold>
             {displayElectionRound(election)}
           </TextHuge>
+        </StatisticItem>
+      </Statistics>
+      <Statistics>
+        <StatisticItem
+          title="Election Period"
+          tooltipText="This shows the Election period detailed information"
+          size="s"
+        >
+          <ElectionPeriod />
         </StatisticItem>
       </Statistics>
       {electionStage === 'announcing' && (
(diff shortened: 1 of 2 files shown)


bedeho:
This should not be merged as is. The original problem described a problem, a solution was randomly suggested in a [comment](https://github.com/Joystream/pioneer/issues/4006#issuecomment-1378496296), there was no review, and here the PR is open for evaluation. I think this particular solution is bad, but thats not even the main point. If we develop the product this way, its going to end up with a totally broken experience. 

**Developers should not start work on something with material UX implications without some sort of PM process being involved, which almost certianly requires a designer for a proper solution, or at least a deliberate decision to not involve a designer because something is urgent and can be fixed later.**

# My comments on the actual solution itself 

From Discord

```
ok, well, this is not the way to do design review, but no, this is not a good solution. These parts of the UI are for dynamic data, this is a bandaid, but not a good UX.
If those periods were mutable, then one could argue it, but not sure even then.
a designer has to do this, not sure who did this
I mean, this doesnt even show what order things are
```

@dmtrjsg

traumschule:
Thanks, that's very clear! Proper design is [on its way](https://discord.com/channels/811216481340751934/1035125653722365993/1062727103797284864). My first idea was a very simple horizontal timeline visualizing the phases (with tooltips) at the top with a dot indicating current position.

election-times

Just sharing a community view on "Developers should not start work on something":
- When implementation and design both may take significant amount of time for independent reasons it can be beneficial to run things in parallel vs sequentially. Sometimes that's wasteful though.
- Our solid review process is able to throw it back to design after implementation if needed.

traumschule (approved):
This should go live asap. Sophisticated designs can be implemented later.

traumschule:
@mkbeefcake can you please have a look at [failing tests](https://github.com/Joystream/pioneer/actions/runs/3892595602/jobs/6644191914).

dmtrjsg:
@traumschule @mkbeefcake Guys let's please work on the new designs and not merge this version.. is it possible to expedite the version which Adam prepared for us? What's the ETA on it?

mkbeefcake:
Hello @traumschule @dmtrjsg 
I am on the trip to my home village for Lunar day, so I told to Chaos77 about this before. so I am worried I couldn't start this asap..

traumschule:
> This should not be merged as is.

The months delay here sets a bad example. The offered solution is good enough for the community and should be merged imo. JSG can improve it later if capacity is available.

Issue #4048, opened 2023-01-09
[USERSNAP] Forum post text overflowing container


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/forum/thread/17
**Fields**:
- Houston, we have a problem.: Forum post text overflowing container
- Comment: All the text should be contained inside the white card

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/702cabfe-b3a6-4732-8829-f05215a83e78-annotated_83cc0764-6f92-4d8b-bdc8-84f1de497ef6.png?etag=a04730664b7254246be4c3c584f7b175)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/702cabfe-b3a6-4732-8829-f05215a83e78-annotated_83cc0764-6f92-4d8b-bdc8-84f1de497ef6.png?etag=a04730664b7254246be4c3c584f7b175)*

**Browser**: Chrome 108 (macOS Catalina)
**Screen size**: 1728x1117
**Browser size**: 1404x971
**[Open #372 in Usersnap](https://app.usersnap.com/l/feedback/68042f9b-19ef-4baa-8372-19912f1e32e7?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

traumschule:
One solution is to truncate or force-break links. What is better?

chrlschwb:
sometimes the text is inside the card and sometimes it's overflowing. What's the cause?

traumschule:
look out for `position: absolute` within `.markdown-preview ol`

IlyaSmiyukha:
resolved with this PR https://github.com/Joystream/pioneer/pull/4296
#4256 duplicated issue

ivanturlakov:
✅ Tested on https://dao-git-dev-joystream.vercel.app/#/forum/thread/17
Mainnet

<img width="1206" alt="Screenshot 2023-04-02 at 15 43 21" src="https://user-images.githubusercontent.com/46903215/229353526-f11ebae7-b77e-48f3-a0ad-354b124cf554.png">

PR #4067, opened 2023-01-11
Forum post text overflow fix (#4048)

chrlschwb:
#4048

### DIFF
--- a/packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
+++ b/packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
@@ -80,6 +80,7 @@ export const MarkdownPreview = ({ markdown, append, ...styleProps }: MarkdownPre
       <MarkdownPreviewStyles {...styleProps} />
       <ReactMarkdown rehypePlugins={rehypePlugins} remarkPlugins={[remarkGfm]} components={components} rawSourcePos>
         {/* {stripBackslashes(markdown)} */}
+        test
       </ReactMarkdown>
       {appendAfter && <p>{append}</p>}
     </div>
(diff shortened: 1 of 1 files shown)


traumschule (approved):
LGTM

links

packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
```diff
@@ -43,7 +43,7 @@ export const MarkdownPreview = ({ markdown, append, ...styleProps }: MarkdownPre
     const p: Components['p'] = ({ children, sourcePosition }) => (
       <p>
         {children}
-        {shouldAppend(sourcePosition) && <> {append}</>}
+        {shouldAppend(sourcePosition) && <p> {append}</p>}
       </p>
```
thesan, line 47: I completely forgot what this `append` is about but having a `p` element inside another `p` element is considered incorrect html. As a result react will show a warning in development mode. Is this change really necessary ?

packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
```diff
@@ -59,7 +59,7 @@ export const MarkdownPreview = ({ markdown, append, ...styleProps }: MarkdownPre
             {children}
           </Mention>
         ) : (
-          <a style={{ overflowWrap: 'break-word' }} href={href}>
+          <a style={{ overflowWrap: 'anywhere' }} href={href}>
```
thesan, line 62: Good solution but please add it in the `MarkdownPreviewStyles` instead of inline.

Also could you try to add it at the `.markdown-preview *` level ? Because even though it's unlikely this issue would occur on anything else than a hyperlink it still might. Eg on a long inline code like this: `firstValueFrom(api.query.referendum.accountsOptedOut.keys()).then((result)`

thesan (changes requested):

Issue #3642, opened 2022-09-29
Forum: Watchlist tab


WRadoslaw:
Prepare watchlist screen based on the My Thread tab implementation

(Part of https://github.com/Joystream/pioneer/issues/3029)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203073054886860) by [Unito](https://www.unito.io)

ivanturlakov:
Tested on https://pioneer-2-git-mainnet-joystream.vercel.app/#/forum/watchlist
wss://34.230.5.182.nip.io/ws-rpc

Didn't see changes since the last [test](https://github.com/Joystream/pioneer/issues/3162#issuecomment-1333466395)
@dmtrjsg Am I using the correct testing env?

https://user-images.githubusercontent.com/46903215/206699789-c0360ec9-dcd2-4c37-a10b-933f8048d86d.mov

dmtrjsg:
@traumschule  the watchlist needs to be implemented as slider that we already have here> 

https://dao.joystream.org/#/forum

pls feel free to assign

ivanturlakov:
✅ Tested on https://dao-2grqxk4u7-joystream.vercel.app/#/forum/watchlist
wss://34.230.5.182.nip.io/ws-rpc

Works as expected on this env


https://user-images.githubusercontent.com/46903215/212340459-1005a99b-8c80-4b4f-9443-725b194209b7.mov

PR #4071, opened 2023-01-11
Update watchlist view to horizontal scroller

gyroflaw:
Fix https://github.com/Joystream/pioneer/issues/3642

### DIFF
--- a/packages/ui/src/app/pages/Forum/ForumWatchlist.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumWatchlist.tsx
@@ -15,7 +15,7 @@ import { useWatchlistedThreads } from '@/forum/hooks/useWatchlistedThreads'
 import { ForumTabs } from './components/ForumTabs'
 
 export const ForumWatchlist = () => {
-  const { threads, isLoading } = useWatchlistedThreads({ threadsPerPage: -1 })
+  const { threads, isLoading } = useWatchlistedThreads()
   const isRefetched = useRefetchQueries({
     interval: MILLISECONDS_PER_BLOCK,
     include: ['GetForumThreads', 'GetForumThreadsCount'],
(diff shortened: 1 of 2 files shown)


traumschule (approved):
[Tested here](https://github.com/Joystream/pioneer/issues/3642#issuecomment-1381920007)

packages/ui/src/forum/hooks/useWatchlistedThreads.ts
```diff
@@ -17,16 +17,16 @@ interface UseMyThreads {
...
 
   const status_json = {
     isTypeOf_not: 'ThreadStatusModerated',
   }
   const variables = {
     where: { id_in: watchlist ?? [], status_json },
-    limit: threadsPerPage,
-    offset: (page - 1) * threadsPerPage,
+    limit: threadsPerPage === -1 ? watchlist?.length || 5 : threadsPerPage,
+    offset: threadsPerPage === -1 ? 0 : (page - 1) * threadsPerPage,
```
thesan, line 29: Currently this hook is used in only one place so the logic can be very simple:
1. `page` and `threadsPerPage` are not needed in the props.
2. no need to pass an `offset` and the limit is always `watchlist?.length`. It's fine if it's `undefined | 0` in this case the query should be skipped.
3. No need to return `pageCount`.
4. The `totalCount` is just `watchlist.length`, no need for `useGetForumThreadsCountQuery`

Finally I think a `skipQuery` property should be provided here so this hook can be called to know the `totalCount` without actually sending the query.

packages/ui/src/app/pages/Forum/ForumWatchlist.tsx
```diff
@@ -1,38 +1,45 @@
...
+                threads.map((thread) => <StyledThreadCard thread={thread} watchlistButton />)
+              ) : (
+                <div>No threads found</div>
+              )
+            }
+          />
+        ) : (
+          <HorizontalScroller items={<ThreadCardSkeleton count={10} />} />
+        )}
+      </RowGapBlock>
```
thesan, line 42: The code looks good but for me (both on Firefox and Chrome) it shows like this when there are less than 6 thread in my watchlist:
image

But I guess this can be handled in another issue...

thesan (changes requested):

thesan (approved):
LGTM!

Issue #4043, opened 2023-01-09
[USERSNAP] Council: Councilor Reward is either wrong or unclear


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/council
**Fields**:
- Houston, we have a problem.: Council: Councilor Reward is either wrong or unclear
- Comment: The current amount shown can easily be read as 1.16 $JOY / 14,400 blocks but it is really not clear and the maths don&#39;t seem to add up if I use the QN to find this information manually.
- Feedback: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/4f624cd3-1673-40ac-9c6a-c5d2e01ea018-annotated_dbe8836c-a103-4c55-9ab3-826d495d3983.png?etag=ce76c2c0f72626cdc517a04eba2cbcd6)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/4f624cd3-1673-40ac-9c6a-c5d2e01ea018-annotated_dbe8836c-a103-4c55-9ab3-826d495d3983.png?etag=ce76c2c0f72626cdc517a04eba2cbcd6)*

**Browser**: Chrome 108 (macOS Catalina)
**Screen size**: 1440x900
**Browser size**: 1396x722
**[Open #368 in Usersnap](https://app.usersnap.com/l/feedback/83f33c52-3a80-42c1-8da8-5810864cb705?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

mochet:
It should be noted that other parts of Pioneer clearly state the amount is per block:
<img width="194" alt="image" src="https://user-images.githubusercontent.com/50830161/211301233-24bbd0c3-d7cf-4876-b332-d362ecfecaf2.png">

bedeho:
Per block generally is going to be too low, if there is a natural period, we should use that.

mochet:
I suggest the following. To be used universally (since there is an inconsistency between WGs and council payments):
* Payments (per block)
* Payment period (payment is made after period has passed)

bedeho:
@DzhideX confirmed that the value here is actually _per block_, so the UX is just super confusing here.
[...]

PR #4083, opened 2023-01-13
Show Councilor reward per day and week (#4043)

traumschule:
#4043

It's easy to switch to monthly if preferred.

reward

### DIFF
--- a/packages/ui/src/app/pages/Council/Council.tsx
+++ b/packages/ui/src/app/pages/Council/Council.tsx
@@ -39,6 +39,8 @@ export const Council = () => {
   const header = <PageHeaderWithHint title="Council" hintType="council" tabs={<CouncilTabs />} />
 
   const isCouncilorLoading = !isRefetched && (isLoading || isLoadingCouncilors)
+
+  const rewardPerDay = useMemo(() => reward?.period?.mul(reward?.amount ?? asBN(0)) ?? asBN(0), [reward])
   const main = (
     <MainPanel>
       <Statistics>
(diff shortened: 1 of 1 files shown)


chrlschwb (approved):

oleksanderkorn (approved):

packages/ui/src/app/pages/Council/Council.tsx
```diff
@@ -54,13 +54,15 @@ export const Council = () => {
...
-            { label: 'Amount', value: reward.amount },
-            { label: 'Period length', value: reward.period, type: 'blocks' },
-          ]}
-        />
+        {reward?.amount && reward?.period && (
+          <MultiValueStat
+            title="Councilor Reward"
+            values={[
+              { label: 'Per Day', value: asBN(+reward.amount * +reward.period) },
+              { label: 'Per Week', value: asBN(7 * +reward.amount * +reward.period) },
```
thesan, line 62: To avoid any risk of going over the max integer limit avoid casting these to numbers. Instead `reward.amount.mul(reward.period)` should work.

But also currently the reward period happens to be a day on mainnet but it could be changed in a runtime upgrade. So instead of relying on the current period length, the reward amount per day should be: `reward.amount.muln(Math.round(A_DAY / MILLISECONDS_PER_BLOCK))`.

Then:
```suggestion
              { label: 'Per Day', value: rewadPerDay },
              { label: 'Per Week', value: rewadPerDay.muln(7) },
```

thesan (changes requested):

packages/ui/src/app/pages/Council/Council.tsx
thesan: @traumschule

traumschule: Thanks! I almost committed your suggestion :)

oleksanderkorn (approved):
LGTM

thesan (approved):
:ok_hand:

Issue #4111, opened 2023-01-23
Multiple constitutionality proposals display upd


dmtrjsg:
## Context

At present proposals that need to undergo multiple constitutionality periods are displayed with multiple tabs. The future rounds of voting are disabled, while the last round is not represented in the tab. 

Here's an example: 
![Screenshot 2023-01-23 at 11.24.49.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/df639806-0cfd-44bc-9ca5-ba071dbe02a0)

## Scope
- [ ] Display all voting periods in tabs all the time with future ones disabled.
In the example provided above, 3 rounds exists, so all of them need to be shown in tabs. 1/3; 2/3; 3/3

- [ ] Add tooltip on hover to the disabled tabs:
Tooltip copy: "This proposal must undergo the voting of multiple consequent councils. The result of each council vote will be displayed in the separate tabs. For the proposal to be approved, each of the councils must approve it." 

- [ ] Add info banner to the active / past rounds of voting.
"The number of councils in that must approve the proposal in a row before it has its intended effect is more than 1. The overall execution of proposal will be triggered after the last council completes the voting. Outcomes of each council voting is displayed in a separate tab."

chrlschwb:
Looks like this issue should be put on hold until there is another proposal with multiple constitutionalities.
I tried to create a test proposal for this purpose but encountered with a new issue.
#4202

dmtrjsg:
This issue should be done together with
- #4201 

Before releasing

cc @chrlschwb

dmtrjsg:
As discussed on the call, let's try and rename tab labels from 1/4; 2/4.. to Round 1, Round 2, Round 3, Round 4... For multiple constitutionality proposals.

@thesan fyi

PR #4119, opened 2023-01-26
Multiple constitutionality proposal display fix  (#4111)

chrlschwb:
fixes issue #4111


packages/ui/src/proposals/hooks/useProposal.ts
```diff
@@ -11,6 +11,8 @@ export const useProposal = (id: string): UseProposal => {
 
   const { loading, data } = useGetProposalQuery(params)
 
+  console.log(data)
+
```
chrlschwb, line 15: This query data doesn't contain all proposal data. In case of 3 constitutionals, this only displays 2 constitutionals.
This is evidently an error that is created during proposal creation.

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -15,31 +16,48 @@ export interface ProposalStagesProps extends ControlProps<number> {
...
         icon: !isDeciding && (rejected ? <CrossIcon /> : <CheckboxIcon />),
         onClick: () => onChange(decidingCount - 1),
       },
-      ...(isDormant ? [{ icon: false, onClick: undefined }] : []),
+      ...(isDormant
+        ? repeat(
+            () => ({ icon: false, onClick: undefined }),
+            totalRoundCount !== undefined ? totalRoundCount - decidingCount - 1 : 1
+          )
+        : []),
```
gyroflaw, line 40: Nice,

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -15,31 +16,48 @@ export interface ProposalStagesProps extends ControlProps<number> {
...
+              : 'The number of councils in that must approve the proposal in a row before it has its intended effect is more than one. The overall execution of proposal will be triggered after the last council completes the voting. Outcomes of each council voting is displayed in a separate tab.'
+          }
+        >
+          <TabContainer key={round} active={round === value} disabled={!onClick} onClick={onClick}>
+            {icon}
+            <span>
+              Council approvals {round + 1}/{constitutionality}
+            </span>
+          </TabContainer>
+        </Tooltip>
```
gyroflaw, line 60: LGTM

gyroflaw (approved):
LGTM

traumschule (approved):
LGTM! As next step we can add expected election and anticipated execution dates via separate PR, for example in `HISTORY` section. cc @bedeho 

const-tooltip

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -15,31 +16,48 @@ export interface ProposalStagesProps extends ControlProps<number> {
...
       {rounds.map(({ icon, onClick }, round) => (
-        <TabContainer key={round} active={round === value} disabled={!onClick} onClick={onClick}>
-          {icon}
-          <span>
-            Council approvals {round + 1}/{constitutionality}
-          </span>
-        </TabContainer>
+        <Tooltip
+          tooltipText={
+            round !== value
```
thesan, line 49: Here the "disabled" tabs are those with `typeof onClick === 'undefined'`... Which is not very readable code.

IMO the best would be to change the type of `rounds`:
from: `{ icon: boolean, onClick: undefined | () => void }` to: `'approved' | 'rejected' | 'deciding' | 'disabled'`.
So this would become something like:
```tsx
{rounds.map((status, round) => {
  const isDisabled = status === 'disabled'
  const isActive = round === value
  const onClick = isDisabled ? undefined : () => onChange(round)
  const icon = iconMap[status]

  return (
  <Tooltip key={round} tooltipText={isDisabled ? "text1..." : "text2..."}>
    <TabContainer active={isActive} disabled={isDisabled} onClick={onClick}>
      {icon}
      <spann>
        ...
```

With `iconMap` defined on the module scope as:
```ts
const iconMap = { approved: <CheckboxIcon />, rejected: <CrossIcon />, deciding: undefined, disabled: undefined }
```
or even as a function with a switch...

thesan (changes requested):
Thank you for the PR @chrlschwb. It still need some work on the behavior but I think some code improvement on the previous solution would be helpful too.

packages/ui/src/proposals/components/ProposalStages.tsx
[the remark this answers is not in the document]
chrlschwb: typeof constitutionality === "string" means if constitutionality === "-" then undefined else if constitutionality === number then constituonality.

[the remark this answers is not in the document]
thesan: I see. That was a bad solution on my end :sweat: . 
It should be:
```ts
export interface ProposalStagesProps extends ControlProps<number> {
  rounds: VotingRound[]
  status: ProposalStatus
  constitutionality?: number
}

export const ProposalStages = ({ rounds, status, constitutionality, value, onChange }: ProposalStagesProps) => {
```

and `Council approvals {round + 1}/{constitutionality ?? '-'}`

This way the return becomes:
```ts
repeat(
  () => ...,
  constitutionality ?? 1
)
```
no need for `totalRoundCount`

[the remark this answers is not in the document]
chrlschwb: image
There are errors when I try to change the code as you suggested.

packages/ui/src/proposals/components/ProposalStages.tsx
traumschule: The [type of `status` is `ProposalStatus`](https://github.com/Joystream/pioneer/blob/6851ba0f445c676586eea8260017142ddc317e5c/packages/ui/src/proposals/components/ProposalStages.tsx#L11) defined [here](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/proposals/types/proposals.ts#L14) as `String`.

packages/ui/src/proposals/components/ProposalStages.tsx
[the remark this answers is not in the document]
thesan: Yes it looks like I wrote `rounds` instead of `updates`. But still the current code should be changed since it doesn't achieve what we want on the deciding stage. Because now we want to always show all the tabs (although some of them might be disabled). The number of tabs will always be `constitutionality`. So there should be only one `repeat`.

About my second comment I was proposing to change the type of the items returned by this repeat from: `{ icon: boolean, onClick: undefined | () => void }` **to**: `'approved' | 'rejected' | 'deciding' | 'disabled'` in order to make the code more readable.

packages/ui/src/proposals/components/ProposalStages.tsx
thesan: @chrlschwb You're right about it currently being an object but this why I suggested to change it to: `'approved' | 'rejected' | 'deciding' | 'disabled'`. I guess calling the variable `status` is confusing maybe `roundStatus` would be better.

packages/ui/src/council/components/election/CandidateVote/CandidateVote.tsx
```diff
@@ -59,8 +59,9 @@ export const CandidateVote = ({
   const roundedPercentage = totalStake.gt(BN_ZERO) ? sumOfAllStakes.muln(100).divRound(totalStake).toNumber() : 0
   const userVoted = myVotes.length > 0
   const allVotesRevealed = myVotes.every((vote) => vote.voteFor)
-  const tooltipLabel = "Vote for your own membership with the same account as used for council candidacy creation counts towards the progress. Please note, this will in a voting lock applied to this account, which may only be withdrawn in the end of the council period if you win the election and released immediately if your candidacy gets outvoted by others. Voting locks are non-rivalrous."
-  const tooltipLinkURL = "https://joystream.gitbook.io/testnet-workspace/system/council#candidacy"
+  const tooltipLabel =
+    'Vote for your own membership with the same account as used for council candidacy creation counts towards the progress. Please note, this will in a voting lock applied to this account, which may only be withdrawn in the end of the council period if you win the election and released immediately if your candidacy gets outvoted by others. Voting locks are non-rivalrous.'
```
traumschule, line 63: This section is superseded by #4155 and could be merged here.

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,68 @@ import React, { useMemo } from 'react'
...
-      {rounds.map(({ icon, onClick }, round) => (
-        <TabContainer key={round} active={round === value} disabled={!onClick} onClick={onClick}>
-          {icon}
-          <span>
-            Council approvals {round + 1}/{constitutionality}
-          </span>
-        </TabContainer>
-      ))}
+      {rounds.map((status, round) => {
+        const isDisabled = status === 'dormant'
```
traumschule, line 45: Referring to the [original comment](https://github.com/Joystream/pioneer/pull/4119#discussion_r1091047618): `dormant` is the phase after approval until the next council. However if a tab `isDisabled` depends on content of `onClick`:
> Here the "disabled" tabs are in fact those with typeof onClick === 'undefined' (not round !== value), which is not very readable code.

traumschule (changes requested):
Please correct / explain.

thesan (changes requested):

traumschule:
Clicking on the tab does not have an effect. As i understand it the votes by that council should be shown.

traumschule:
This looks very good already. History shows all stages with date and it's possible to click each tab to review votes and rationales.

It is just unclear why after 3x3 CM approved the proposal is still deciding. Like what is the next step? The answer isn't obvious in either stage. Suggest to merge soon and create follow-ups.

consti

@bedeho wdyt? https://dao-git-fork-chrlschwb-fix-4111-proposalsdisplay-joystream.vercel.app/#/proposals/preview/63

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,73 @@ import React, { useMemo } from 'react'
...
-  status: ProposalStatus
+  roundStatus: ProposalStatus
   updates: ProposalStatusUpdates[]
-  constitutionality?: number | string
+  constitutionality?: number
 }
 
-export const ProposalStages = ({ status, updates, constitutionality = '-', value, onChange }: ProposalStagesProps) => {
-  const rounds = useMemo(() => {
+type RoundState = 'approved' | 'rejected' | 'deciding' | 'disabled'
```
traumschule, line 16: Maybe these should be included from `types.ts`? I checked the code and there are [multiple locations](https://github.com/search?q=org%3AJoystream+dormant&type=code) that hard-code possible stages. @thesan wdyt? For example [here](https://github.com/Joystream/pioneer/blob/bb6f63bd6455c812b8934d96f0b964bb4192d782/packages/ui/src/proposals/types/proposals.ts#L27):
> type CurrentProposalStatus = Extract<ProposalStatus, 'deciding' | 'gracing' | 'dormant'>

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,73 @@ import React, { useMemo } from 'react'
...
+        const isDisabled = roundStatus === 'disabled'
+        const isActive = round === value
+        const onClick = isDisabled ? undefined : () => { return onChange(round) }
+        const icon = iconMap[roundStatus]
+        return (
+
+          <Tooltip
+            key={round}
+            tooltipText={
+              !isDisabled
```
traumschule, line 59: minor nits:
- check for `isDisabled` (not inverse)
- define strings / test earlier to make component easier to read:
>  <Tooltip key={round} tooltipText={isDisabled ? tooltip.disabled : tooltip.active}>

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,73 @@ import React, { useMemo } from 'react'
...
+                ? 'This proposal must undergo the voting of multiple consequent councils. The result of each council vote will be displayed in the separate tabs. For the proposal to be approved, each of the councils must approve it.'
+                : 'The number of councils in that must approve the proposal in a row before it has its intended effect is more than one. The overall execution of proposal will be triggered after the last council completes the voting. Outcomes of each council voting is displayed in a separate tab.'
```
traumschule, line 61:
```suggestion
                : 'The number of councils that must approve the proposal in a row before it has its intended effect is more than one. The execution is triggered after the last council completes the voting. Votes for each council are displayed in a separate tab.'
```
Question: In above screenshot voting is complete but status is "Deciding". What am i missing?
> The execution is triggered after the last council completes the voting.

traumschule (approved):
LGTM, some room to refactor though.

bedeho:
I am a bit confused, why do I see voting like this for `2/3`

Screenshot 2023-02-19 at 22 16 51


is this a bug? seems to combine voting for differnt councils which makes 0 sense, we should show voting pattern for the council selected only.

bedeho:
Also, what is going on here with being in deciding ?

Screenshot 2023-02-19 at 22 17 54

gyroflaw:
> I am a bit confused, why do I see voting like this for `2/3`
> 
> Screenshot 2023-02-19 at 22 16 51
> 
> is this a bug? seems to combine voting for differnt councils which makes 0 sense, we should show voting pattern for the council selected only.

I think this should be handled in another PR :)

packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
```diff
@@ -160,7 +162,7 @@ export const ProposalPreview = () => {
 
           {(proposal.status === 'dormant' || votingRounds.length > 1) && (
             <ProposalStages
-              status={proposal.status}
+              roundStatus={proposal.status}
```
thesan, line 165: I don't think this should be changed. This is the status of the proposal so "status" is correct

packages/ui/src/council/components/election/CandidateVote/CandidateVote.tsx
```diff
@@ -89,6 +90,7 @@ export const CandidateVote = ({
                   tooltipLinkText="Read more"
                   tooltipLinkURL="https://handbook.joystream.org/system/council#vote"
                 >
+                  tooltipLinkURL="https://handbook.joystream.org/system/council#vote"
```
thesan, line 93: This look like a mistake

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,73 @@ import React, { useMemo } from 'react'
 import { ControlProps } from '@/common/components/forms'
 import { CheckboxIcon, CrossIcon } from '@/common/components/icons'
 import { TabContainer, TabsContainer } from '@/common/components/Tabs'
+import { Tooltip } from '@/common/components/Tooltip'
 import { last, repeat } from '@/common/utils'
 import { ProposalStatus, ProposalStatusUpdates } from '@/proposals/types'
 
 export interface ProposalStagesProps extends ControlProps<number> {
-  status: ProposalStatus
+  roundStatus: ProposalStatus
```
thesan, line 11: Again this should remain "status"

packages/ui/src/proposals/components/ProposalStages.tsx
thesan: This type is just for this file. It's the status of a round which is different from the status of the proposal this is why I think it should stay here and not be exported.

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,73 @@ import React, { useMemo } from 'react'
...
+type RoundState = 'approved' | 'rejected' | 'deciding' | 'disabled'
+const iconMap = { approved: <CheckboxIcon />, rejected: <CrossIcon />, deciding: undefined, disabled: undefined }
+
+export const ProposalStages = ({ roundStatus, updates, constitutionality, value, onChange }: ProposalStagesProps) => {
+  const rounds: RoundState[] = useMemo(() => {
     const decidingCount = updates.filter(({ status }) => status === 'deciding').length
+
     const lastUpdate = last(updates).status
-    const onGoing = lastUpdate === status
+    const onGoing = lastUpdate === roundStatus
```
thesan, line 24:
```suggestion
    const onGoing = lastUpdate === status
```

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,73 @@ import React, { useMemo } from 'react'
...
+  constitutionality?: number
 }
 
-export const ProposalStages = ({ status, updates, constitutionality = '-', value, onChange }: ProposalStagesProps) => {
-  const rounds = useMemo(() => {
+type RoundState = 'approved' | 'rejected' | 'deciding' | 'disabled'
+const iconMap = { approved: <CheckboxIcon />, rejected: <CrossIcon />, deciding: undefined, disabled: undefined }
+
+export const ProposalStages = ({ roundStatus, updates, constitutionality, value, onChange }: ProposalStagesProps) => {
+  const rounds: RoundState[] = useMemo(() => {
```
thesan, line 20: :+1:

packages/ui/src/proposals/components/ProposalStages.tsx
thesan: @traumschule it shows 3 votes but they are the votes from the first round so it looks to me like it the issue #4201

packages/ui/src/proposals/components/ProposalStages.tsx
```diff
@@ -3,44 +3,73 @@ import React, { useMemo } from 'react'
...
-      },
-      ...(isDormant ? [{ icon: false, onClick: undefined }] : []),
-    ]
-  }, [updates.length, status])
+    const isDormant = onGoing && roundStatus === 'dormant'
+
+    return repeat((round) => {
+      if (round < decidingCount) {
+        return 'approved'
+      } else if (round > decidingCount) {
```
thesan, line 32: It should be `decidingCount - 1`:
```suggestion
      if (round < decidingCount - 1) {
        return 'approved'
      } else if (round > decidingCount - 1) {
```

thesan (changes requested):

packages/ui/src/proposals/components/ProposalStages.tsx
chrlschwb: Ok. I will do.

traumschule:
```
 | Summary of all failing tests
 | FAIL test/council/pages/PastCouncil.test.tsx (13.677 s)
 |   ● UI: Past Council page › Council found › Renders
 | 
 |     Timed out in waitForElementToBeRemoved.
```

thesan (approved):
@chrlschwb I implemented some changes since this PR went through a lot of review cycle already. I also included @traumschule requests. @traumschule these were good suggestion, please don't approve PRs when you found some code quality and typos issues.

I realized there was a confusing point about the `updates` not including the initial deciding round which may have contributed to the confusion on this PR.
It also results in this minor bug #4213.

Issue #3353, opened 2022-07-13
Right side of proposal page is outside of the screen


ivanturlakov:
## Scope

- [ ] Make sure all components on a page fit to a page on smaller screens (13 inch)

⚠️ we are suggesting not to wait till the #1794 epic design is finished, and suggesting to fix this one by reducing the width and white space in the table. 

## Context

Tested on prod
MacBook 13`
Latest Chrome

Proposal page
<img width="1440" alt="Снимок экрана 2022-07-13 в 13 10 59" src="https://user-images.githubusercontent.com/46903215/178709823-22d52df9-0050-4cd1-920a-1ac189ee8277.png">

vrrayz:
I'd like to take this up.

vrrayz:
I just made a PR for this here => https://github.com/Joystream/pioneer/pull/4146

ivanturlakov:
✅ Tested on https://dao-git-dev-joystream.vercel.app/#/proposals/current

<img width="1440" alt="Снимок экрана 2023-03-05 в 15 21 18" src="https://user-images.githubusercontent.com/46903215/222963061-df9e679b-d06c-4af2-a836-8d9eb158d356.png">

PR #4146, opened 2023-02-05
Issue 3353

vrrayz:
Fixed https://github.com/Joystream/pioneer/issues/3353 by reducing the padding and column-gap on proposal list item

Screenshot (485)


chrlschwb (approved):
LGTM

packages/ui/src/proposals/constants/styles.ts
```diff
@@ -5,8 +5,8 @@ import { ListHeaders, ListHeader } from '@/common/components/List/ListHeader'
 export const ProposalColLayout = '296px 148px 156px 60px 92px 1fr'
 
 export const ProposalsListHeaders = styled(ListHeaders)`
-  padding-right: 16px;
-  grid-column-gap: 36px;
+  padding-right: 0px;
+  grid-column-gap: 21px;
```
thesan, line 9: This new layout looks even more cramped on bigger screen:
image

Instead I'm proposing to just remove some unused columns from the `ProposalColLayout`. It should be fine because the current proposal tabs has 3 columns and the past proposal past has 4 of them. But `ProposalColLayout` defines 5 fixed width columns + a flexible width one. Deleting the last `92px` wide columns should be enough (actually it should give even more space than removing the right paddings when accounting for the gap between columns) and it will keep the same layout on bigger screens:
```ts
export const ProposalColLayout = '296px 148px 156px 60px 1fr'
```

thesan (changes requested):

packages/ui/src/proposals/constants/styles.ts
vrrayz: That works as well, i have made the necessary changes,

chrlschwb (approved):

packages/ui/src/proposals/components/ProposalList/ProposalListItem.tsx
```diff
@@ -108,8 +108,8 @@ const CopyButton = styled(CopyButtonTemplate)`
 `
 
 const ProposalItem = styled(TableListItem)`
-  grid-column-gap: 36px;
-  padding-right: 16px;
+  grid-column-gap: 20px;
+  padding-right: 0px;
```
thesan, line 112: Please keep the original padding on the ProposalItem too, so they remain aligned with the headers and also so they don't look crammed.

thesan (changes requested):

thesan (approved):
:+1:

PR #4147, opened 2023-02-06
Fixed Joystream Pioneer Forum Back Link #3768

vrrayz:
Fixed forum back link issue on #3768 by modifying the Pagination component to add a page query string and using a custom hook to check for that query string and set the page based on the result

https://user-images.githubusercontent.com/33874323/217080212-f41d4142-7cf7-4487-ae1a-bc815a51da3c.mp4

### DIFF
--- a/packages/ui/src/common/components/Pagination/Pagination.tsx
+++ b/packages/ui/src/common/components/Pagination/Pagination.tsx
@@ -1,5 +1,6 @@
 import React, { FC } from 'react'
 import ReactPaginate from 'react-paginate'
+import { useHistory } from 'react-router'
 import styled from 'styled-components'
 
 import { Arrow } from '@/common/components/icons'
@@ -12,8 +13,14 @@ interface PaginationProps {
 }
 
 export const Pagination: FC<PaginationProps> = ({ pageCount = 0, handlePageChange, page }) => {
+  const history = useHistory()
+
   if (pageCount < 2) {
     return null
+  }
+  const pageChangeAction = (value: any) => {
+    history.push(`?page=${value.selected + 1}`)
+    handlePageChange(value.selected + 1)
   }
 
   return (
@@ -32,7 +39,7 @@ export const Pagination: FC<PaginationProps> = ({ pageCount = 0, handlePageChang
         nextLabel={<Arrow direction="right" />}
         nextLinkClassName="pagination__link"
         previousLinkClassName="pagination__link pagination__link--previous"
-        onPageChange={(value) => handlePageChange(value.selected + 1)}
+        onPageChange={(value) => pageChangeAction(value)}
         forcePage={page && page - 1}
       />
     </StyledPaginateContainer>
--- a/packages/ui/src/forum/components/PostList/PostList.tsx
+++ b/packages/ui/src/forum/components/PostList/PostList.tsx
@@ -35,9 +35,6 @@ export const PostList = ({ threadId, isThreadActive, isLoading, isDiscussion }: 
     (page: number) => history.replace({ pathname, search: page > 1 ? `page=${page}` : '' }),
     []
   )
-  useMemo(() => {
-    if(isReady && !posts.length) history.replace({pathname, search: ''});
-  },[isReady, posts])
 
   const postsRefs: AnyKeys = {}
   const getInsertRef = (postId: string) => (ref: RefObject<HTMLDivElement>) => (postsRefs[postId] = ref)
--- a/packages/ui/src/forum/components/PostList/PostList.tsx
+++ b/packages/ui/src/forum/components/PostList/PostList.tsx
@@ -35,6 +35,9 @@ export const PostList = ({ threadId, isThreadActive, isLoading, isDiscussion }: 
     (page: number) => history.replace({ pathname, search: page > 1 ? `page=${page}` : '' }),
     []
   )
+  useMemo(() => {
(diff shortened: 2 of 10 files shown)


packages/ui/src/forum/constant/pagination.ts
```diff
@@ -1 +1 @@
-export const THREADS_PER_PAGE = 30
+export const THREADS_PER_PAGE = 10
```
traumschule, line 1: was this just for testing?

packages/ui/src/forum/hooks/useForumCategoryThreadPage.ts
```diff
@@ -0,0 +1,10 @@
+export const useForumCategoryThreadPage = () => {
+  const locationHash = window.location.hash
+  const queryString = locationHash.substring(locationHash.indexOf('?'))
+  const urlParams = new URLSearchParams(queryString)
+  if (urlParams.has('page')) {
+    const num = Number(urlParams.get('page')) || 1
+    return num
+  }
+  return 1
```
traumschule, line 9: Does this work:
```suggestion
  return Number(urlParams.get('page')) || 1
```

packages/ui/src/common/components/Pagination/Pagination.tsx
```diff
@@ -12,9 +13,15 @@ interface PaginationProps {
...
 
 export const Pagination: FC<PaginationProps> = ({ pageCount = 0, handlePageChange, page }) => {
+  const history = useHistory()
+
   if (pageCount < 2) {
     return null
   }
+  const pageChangeAction = (value: any) => {
+    history.push(`?page=${value.selected + 1}`)
+    handlePageChange(value.selected + 1)
```
traumschule, line 23: Maybe i'm missing something: Why can page only increase?

traumschule (changes requested):
Tried to test with [this thread](https://dao-git-fork-vrrayz-issue-3768-joystream.vercel.app/#/forum/thread/125?page=2): The page is empty and has no navigation. Please check.

packages/ui/src/forum/constant/pagination.ts
vrrayz: yes

packages/ui/src/common/components/Pagination/Pagination.tsx
vrrayz: value.selected is zero index based so for page 1 value.secleted is zero

packages/ui/src/forum/constant/pagination.ts
vrrayz: I'll set that back to the default 30

packages/ui/src/forum/hooks/useForumCategoryThreadPage.ts
vrrayz: It works as well

vrrayz:
> Tried to test with [this thread](https://dao-git-fork-vrrayz-issue-3768-joystream.vercel.app/#/forum/thread/125?page=2): The page is empty and has no navigation. Please check.

Thats because the total amount of posts under this thread isn't enough to reach to a page 2 or 3. I could set it up to redirect to page 1 once the users try getting to anything greater than the pageCount. would that be okay?

traumschule:
That's a good idea. Also pagination navigation can be hidden / disabled depending on number of posts and current page.

vrrayz:
> That's a good idea. Also pagination navigation can be hidden / disabled depending on number of posts and current page.

okay.. i'll get right to it

traumschule (approved):
LGTM

oleksanderkorn (approved):
LGTM

packages/ui/src/forum/components/PostList/PostList.tsx
```diff
@@ -35,6 +35,9 @@ export const PostList = ({ threadId, isThreadActive, isLoading, isDiscussion }:
     (page: number) => history.replace({ pathname, search: page > 1 ? `page=${page}` : '' }),
     []
   )
+  useMemo(() => {
+    if(isReady && !posts.length) history.replace({pathname, search: ''});
+  },[isReady, posts])
```
thesan, line 40: I don't understand. What is this change for ?
Also if this is needed please use a `useEffect` instead of `useMemo`.
Finally it looks like the linting wasn't applied.

thesan (changes requested):
Good job :+1: adding the page to the history is a good step forward.

There are still a bunch of issues with this back button but I think they can be taken care of in a follow up PR.

Just please add response to my comment :point_down:

packages/ui/src/forum/components/PostList/PostList.tsx
vrrayz: I think i tried useEffect at first but the whole page wasn't rendering properly. and the this is just to replace the history to the first  page if there's a case where someone enters a page number that probably doesn't have any posts

thesan: @vrrayz 
I don't really understand this is a change on the PostList component which appears on forum threads, but this PR is about adding page numbers to the query params of the category page urls.

Also `posts.length` is never `0` because `posts` includes the thread initial post.

Unless I'm missing something, could you please remove this change. The rest of the PR is great.

thesan (approved):

Issue #4122, opened 2023-01-26
WASM validation fails for compressed runtime blob


mnaamani:
When building the runtime WASM blobs produced are as follows:

```
-rw-r--r--  1 mokhtar  staff   1.0M Jan 23 11:41 joystream_node_runtime.compact.compressed.wasm
-rw-r--r--  1 mokhtar  staff   5.0M Jan 23 11:41 joystream_node_runtime.compact.wasm
-rwxr-xr-x  1 mokhtar  staff   5.6M Jan 23 11:41 joystream_node_runtime.wasm
```

We use the compressed version because the size of the non-compressed blobs are too large and do not fit in a block. The runtime also limits the size of the blob to ~3MB for proposals.

Pioneer fails to validate when using the compressed version.

![runtime-upgrade-proposal](https://user-images.githubusercontent.com/1621012/214811582-934e2e7a-1860-4249-9893-3ff19a971322.png)

Validation is performed here when processing the dropped file:
https://github.com/Joystream/pioneer/blob/9d4e04c3a0c945f6c1774475a0a57c8a031c6d0f/packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx#L45

We need a step to decompress the file first before validating it, but the compressed file is the final one that should actually be used in the proposal details.

The compression algorith used is: https://github.com/facebook/zstd
Reference in substrate node where it is using `zstd` cargo crate https://crates.io/crates/zstd

Will need to find an implementation in javascript to use in pioneer. Here is one but I have not tested it.
https://www.npmjs.com/package/zstandard-wasm

dmtrjsg:
This one seems trivial, once ephesus Pioneer scope is done, Theo could take this one on to turnaround in a few hours.

cc @mnaamani @bwhm

mnaamani:
Reference implementation to handle decompressing of wasm blob and inspecting runtime version

https://github.com/Joystream/joystream/pull/4629/files#diff-0b8c337c90045d20004df9d58183118f424b6bc70383c5f22fcde8ef0b90ea3f

ivanturlakov:
Tested on https://pioneer-2-kvulka3th-joystream.vercel.app/#/proposals/current
wss://54.205.24.16.nip.io/ws-rpc

Used this blob https://pioneerapp.xyz/#/proposals/preview/175

⚠️ Validation was successful, but the button is inactive
[...]

PR #4182, opened 2023-02-16
Decompress wasm runtime before validating

mnaamani:
Fixes https://github.com/Joystream/pioneer/issues/4122

Effect on bundle size, increased by ~600KB

New bundle size:
```
 | WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
 | Entrypoints:
 |   main (14.1 MiB)
 |       runtime~main.a482ee5c0a6b9c4e7e91.js
 |       main.6466ad456ae28e6c37a6.js
```

Old bundle size:
```
 | WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
 | Entrypoints:
 |   main (13.5 MiB)
 |       runtime~main.18b2bca8852a873cb1ab.js
 |       main.0fa1610cb597cee324a3.js
```

There is a separate issue with actually submitting the proposal, the "Create Proposal" button is not enabled.

Screenshot 2023-02-24 at 10 50 12 AM


mnaamani:
I found the problem, the package I chose doesn't support "streaming decompression". (Originally selected as it is tiny package size)

Tested with https://www.npmjs.com/package/@oneidentity/zstd-js/v/1.0.3 and that works correctly, but the package is ~67Mb. Although docs show that it is possible to import submodules specifically for just decompression which should hopefully not impact the total size of packed pionner bundle.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
```diff
@@ -24,6 +25,22 @@ const MAX_FILE_SIZE = 3 * 1024 * 1024
...
+// https://github.com/paritytech/substrate/blob/master/primitives/maybe-compressed-blob/src/lib.rs
+// Look for 8 byte magic prefix in blob that indicates it is compressed with zstd algorithm
+// If is compressed, strips the prefix and decompresses remaining bytes, otherwise returns original blob
+const maybeDecompressRuntimeBlob = async (blob: ArrayBuffer): Promise<Buffer> => {
+  const ZSTD_PREFIX = Buffer.from([82, 188, 83, 118, 70, 219, 142, 5])
+  let wasm = Buffer.from(blob)
+  const prefix = wasm.slice(0, 8)
+  const isCompressed = Buffer.compare(prefix, ZSTD_PREFIX) === 0
+  if (isCompressed) {
+    const { ZstdStream } = await ZstdInit()
```
mnaamani, line 37: I suspect this is not he best place to import or load the library, but I'm not an expert in React so perhaps it can be refactored by someone else, or suggest how/where to place initializing of the library and how to pass it down to this component.

Was also looking for a way to "unload" the library from memory (as I have seem some other wasm libraries do) but couldn't find it in the documentation.

thesan (approved):

thesan:
@mnaamani thank you for the WASM validation logic :pray: I would have struggled with this !

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/RuntimeUpgrade.tsx
thesan: I almost forgot about this! You're right it makes sense to move this lib out of the main bundle.

I chose to move it to a web worker in [this commit](https://github.com/Joystream/pioneer/pull/4182/commits/d35c3c6b5f7b229560c0478bad67b843811408f9) (because it's already a common pattern in Pioneer) however I'm not sure where did the library actually went:
Screenshot from 2023-05-12 14-37-51

The main bundle lost the correct size, but I couldn't find any bundles which significantly grew. Maybe the lib was already a dependency in that chunk.

thesan (approved):

Issue #4052, opened 2023-01-10
Add error boundaries


thesan:
## Context
Exceptions not caught within the component they occurred in will crash the whole app. E.g this is what was happening in #3994. It forces users to reload the app (maybe on a different page to avoid the app crashing again), and wait for the connections to the QN/chain to re-establish.

## Scope
- [ ] Add react [error boundaries](https://reactjs.org/docs/error-boundaries.html) to catch pages and modal crashes.
- [ ] Maybe show a dialog acknowledging that the page/modal crashed and the error message.

Issue #4180, opened 2023-02-15
[USERSNAP] History page for the SWG blank


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/working-groups/storage
**Fields**:
- Houston, we have a problem.: History page for the SWG blank
- Comment: When navigating to the history tab on the SWG I get bank page. I am using a brave browser.

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/0e90871a-4f7e-4b23-84a0-004ed5e65c3d-annotated_73bc3b9a-c6d6-44ea-9288-771296b3c39b.png?etag=22960bfa69b689a45aa7c36b55371f7c)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/0e90871a-4f7e-4b23-84a0-004ed5e65c3d-annotated_73bc3b9a-c6d6-44ea-9288-771296b3c39b.png?etag=22960bfa69b689a45aa7c36b55371f7c)*

**Browser**: Chrome 110 (Windows 10)
**Screen size**: 1920x1080
**Browser size**: 1920x973
**[Open #391 in Usersnap](https://app.usersnap.com/l/feedback/55704d8f-4a2c-4aa5-a186-cf828c1f4be7?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
There's an error which is browser-agnostic: 

<img width="1435" alt="Screenshot 2023-02-16 at 13 40 06" src="https://user-images.githubusercontent.com/93924525/219327706-67d379c0-017e-4b3a-aedf-0ad61e6a92cd.png">

Issue #4034, opened 2023-01-07
Regression on WG `History` tab


traumschule:
App crash on `/#/working-groups/storage`
```
react-dom.production.min.js:216 TypeError: Cannot read properties of undefined (reading 'map')
    at useGroupActivities.ts:35:53
    at Object.So [as useMemo] (react-dom.production.min.js:168:377)
    at n.useMemo (react.production.min.js:23:113)
    at n.useGroupActivities (useGroupActivities.ts:21:22)
    at n.HistoryTabSidebar (HistoryTab.tsx:38:26)
```
Edit: only with https://dao-git-fork-mkbeefcake-fix-forumpostreplyntshow4022-joystream.vercel.app/#/working-groups/storage (#4023)

traumschule:
Can't reproduce it on the official deployment currently.

traumschule:
Happened again: https://pioneer-2-git-fork-chrlschwb-fix-4041-myrolesvalues-joystream.vercel.app/#/working-groups/builders

IlyaSmiyukha:
I can replicate the issue as well. The backend doesn't return workerRewardAmountUpdatedEvents in the API call here. It is always undefined. And that's why the page crashes. See the screenshot below. I can't find Api documentation. To check is it stil needed.
<img width="1443" alt="Screenshot 2023-02-19 at 14 52 25" src="https://user-images.githubusercontent.com/39830587/219949412-203d2c4a-9a13-4a77-a8f3-261af78e3da3.png">

chrlschwb:
@IlyaSmiyukha 
This is the query node for Joystream data.
https://query.joystream.org/graphql

thesan:
@chrlschwb please don't close issues until they are merged in production (`main`)

PR #4195, opened 2023-02-20
Add error boundary

gyroflaw:
Fix https://github.com/Joystream/pioneer/issues/4052

### DIFF
--- a/packages/ui/src/app/Providers.tsx
+++ b/packages/ui/src/app/Providers.tsx
@@ -1,4 +1,5 @@
 import React, { ReactNode } from 'react'
+import { ErrorBoundary } from 'react-error-boundary'
 import { HashRouter } from 'react-router-dom'
 
 import { AccountsContextProvider } from '@/accounts/providers/accounts/provider'
@@ -13,6 +14,7 @@ import { OnBoardingProvider } from '@/common/providers/onboarding/provider'
 import { TransactionStatusProvider } from '@/common/providers/transactionStatus/provider'
 import { MembershipContextProvider } from '@/memberships/providers/membership/provider'
 
+import ErrorFallback from './components/ErrorFallback'
 import { GlobalStyle } from './providers/GlobalStyle'
 import { RouteActions } from './RouteActions'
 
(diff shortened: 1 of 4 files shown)


chrlschwb (approved):
LGTM.
I've never encountered Pioneer app crash yet. How will QA team test this?

packages/ui/src/app/components/ErrorFallback.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+interface FallbackProps {
+  error: Error
+  resetErrorBoundary: (...args: Array<unknown>) => void
+}
+
+export default function ErrorFallback({ error, resetErrorBoundary }: FallbackProps) {
+  return (
+    <Modal modalSize="m" onClose={resetErrorBoundary}>
+      <ModalHeader icon={<FailureIcon />} title="Something went wrong" onClick={resetErrorBoundary} />
```
traumschule, line 15: This is close to the [documentation](https://reactjs.org/docs/error-boundaries.html). However in our case most likely crashes hide in pages / tabs. From my view the goal wouldn't be a global modal  "Something went wrong" but replacing the specific component with a "Crashed" message and if possible a trace or instructions how triage and report.

packages/ui/src/app/Providers.tsx
```diff
@@ -21,32 +23,39 @@ interface Props {
...
-      </ApiContextProvider>
-    </NetworkEndpointsProvider>
-  </KeyringContextProvider>
+  <ErrorBoundary
+    FallbackComponent={ErrorFallback}
+    onReset={() => {
+      // reset the state of your app so the error doesn't happen again
+    }}
+  >
+    <KeyringContextProvider>
```
thesan, line 32: Catching errors all the way up here is a good idea it will allow users to still use usersnap to report errors for crashes happening in providers.

However apart from that it still mean loosing the whole app and needing to reload the page. As described in the issue #4052 other boundaries should be added:
1. to catch page crashes. I think it should be around `<Screen />` in `App.tsx`. This way the navigation is still available and the user can simply click away from the faulty page.
2. to catch modal crashes. I think it has to be done around `<GlobalModals />` (although maybe there is a better solution). The error should show in a modal too. This way the rest of the app is still available and only the modal crashed.

packages/ui/src/app/components/ErrorFallback.tsx
thesan: Yes I completely agree

thesan (changes requested):
@chrlschwb @gyroflaw based on #4034 and #4180 `/#/working-groups/storage` is crashing ATM. So this route should help testing the error boundaries for development and later QA.

packages/ui/src/app/Providers.tsx
gyroflaw: totally agreed will update PR

traumschule:
in favor of 4439

PR #4200, opened 2023-02-24
Display blacklisted accounts in the election module (#4138)

### LINKED ISSUE #4138: Display blacklisted accounts in the election module
## Context

In Ephesus release, there will be a possibility to blacklist specific accounts, meaning it will not be possible to use them in voting for council candidates. 

Original issue:
- https://github.com/Joystream/joystream/issues/2927

So we need a way to display all blacklisted accounts, their balances and the net balance in context of elections. 

## Scope

- [ ] Display the list of blacklisted accounts, their balance and total balance of JOY tokens that cannot participate in governance. 
- [ ] Add membership information on the accounts that are added to the list
- [ ] Add a note "These accounts are excluded from participating in elections, and cannot be used for voting"

## Design Implementation Suggestion

Options that come to my mind: 

1. Widget

New widget on the elections page, similar to signal proposal additional details

..that can trigger a side panel, displaying a list of accounts in the same way how multiple accounts funding [proposal is designed](https://www.figma.com/file/YBfvoRpnduOXxUnA9kvNP5/Pioneer-Design?node-id=38%3A85550&t=PKAcJiblNhNAJS0A-1)

Preview: 

2. Tab

Add a new tab  "Blacklisted accounts" to elections to display the list of blacklisted accounts.

chrlschwb:
fixes #4138


packages/ui/src/app/config/network.ts
```diff
@@ -28,30 +28,30 @@ type PredefinedEndpoint = { [K in NetworkType]?: string }
 
 const QUERY_NODE_ENDPOINT_SUBSCRIPTION: PredefinedEndpoint = {
   mainnet: MAINNET_QUERY_NODE_SOCKET,
-  local: 'ws://localhost:8081/graphql',
+  local: 'wss://joy-testnet.joystream.org/query-node/server/graphql',
   testnet: TESTNET_QUERY_NODE_SOCKET,
-  'local-mocks': 'ws://localhost:8081/graphql',
+  'local-mocks': 'wss://joy-testnet.joystream.org/query-node/server/graphql',
```
traumschule, line 33: These changes should only happen locally, ideally via `.env`.

traumschule:
Please find or create an issue about it.

Issue #3997, opened 2022-12-29
Consider optimized design for narrow screens


traumschule:
On multiple occasions we ran into design issues when available window size is lowered below a certain limit.

We do not want to require users to have a wide screen and instead should also take mobile users into account.

For example the left sidebar could be collapsible and wide tables can be reduced to relevant fields.
Horizontal sliders can be reduced to one element with three-dots navigation on small screens.

Accounts on tiled screen
![acc-sm-scrn](https://user-images.githubusercontent.com/31551045/209950093-6cdd04e4-adb6-42aa-a3ad-5df54d42f7b2.png)
Only at 75% zoom all columns become visible.

traumschule:
Another visualization how to easily improve layout by hiding left and right columns / sidebars. Not only on mobile.

![narrow](https://user-images.githubusercontent.com/31551045/211144396-a30eecea-937e-4549-8703-6d6070c61a3b.png)

dmtrjsg:
@traumschule we have #1794 epic, where this could be addressed, but that is a huuuuge undertaking and based on the user agent analysis there is hardly any users from mobile. 

I agree that the app responsiveness should be addressed holistically by the designer at some point, but suggest park this for the time being and do work as part of the Epic.

traumschule:
Ok, will address this for JSS on a separate branch then. :handshake:

traumschule:
Reopening to single out the collapsible sidebar.

dmtrjsg:
The goal here would be to view Pioneer and not interact with it? Starting to work on this we need to be wary that this PR will be hard to merge later in the future so before we do any work on this let's make sure that collectively this can be a focus of few people. 

Also just hiding sidebar will not just do much by itself :/ 

Also you may find these helpful: 
- https://github.com/Joystream/atlas/issues/2903
-There are a bunch of apps can also enable interacting with Pioneer and not just viewing [link](https://support.polkadot.network/support/solutions/articles/65000068702-where-to-store-dot-polkadot-wallet-options)

traumschule:
> The goal here would be to view Pioneer and not interact with it?

I don't understand the question in this context. The goal is to get the sidebar out of the way because it blocks most of the screen.

dmtrjsg:
I believe we should define some sort of an MVP for mobile view, as simply making the side bar collapsible may not cut it for decent experience. There is a danger in patchy approach here, but I guess we can try and get the branch started for POC. 

[...]

PR #4208, opened 2023-02-27
Adaptive design #1

catch-a-chalk:
What has been changed when the screen is compressed:
- the buttons and labels line up in rows;
- indicators in the "working groups" go under the description;
- "working gropus" and "council"- activities go down after cells;
- the "settings" was stretched to 100%.

Suggestions: 
- black navigation panel has to be wrapped in mobile screens;
- the sizes of some columns and rows are precisely specified, which can break the grid when adding content. It needs to add paddings and remove the exact width.


traumschule:
This targets #1794 and is blocked by #3997.

Issue #4213, opened 2023-03-03
Proposals history doesn't show the initial deciding stage


thesan:
# Context

New proposals history show the virtual "Created" stage:

![Screenshot from 2023-03-03 17-23-42](https://user-images.githubusercontent.com/6571453/222775678-5241a9b6-f167-422a-b3f0-4ad1d49c72f4.png)

---

Once they are approved/rejected the next stage is shown right after "Created". E.g:

![Screenshot from 2023-03-03 17-23-55](https://user-images.githubusercontent.com/6571453/222774856-f22e9705-7f66-4223-82fd-843707460ebe.png)

---

# Solution
- Add a "Deciding" stage after "Created"
- Display correct event on each of the multi-const tabs for such proposals

Issue #4201, opened 2023-02-24
Multiple Constitutionality Proposal Error in Council Display


chrlschwb:
The original comment can be found here
https://github.com/Joystream/pioneer/pull/4119#issuecomment-1436096153

<img width="281" alt="r" src="https://user-images.githubusercontent.com/101491745/221137268-144b26b1-0a11-44cb-9c84-063f35439a8a.png">

dmtrjsg:
We need to show  correct voters (correct council corresponding the the council term when the vote happened)  on each individual tab as we are managing multiple rounds for proposals via tabs.

PR #4226, opened 2023-03-13
Multiple Constitutionality Proposal View Fix

### LINKED ISSUE #4111: Multiple constitutionality proposals display upd

chrlschwb:
fixes #4111, #4213, and #4201

### DIFF
--- a/packages/ui/src/proposals/components/ProposalHistory.tsx
+++ b/packages/ui/src/proposals/components/ProposalHistory.tsx
@@ -12,9 +12,10 @@ interface ProposalHistoryProps {
 }
 
 export const ProposalHistory = ({ proposal }: ProposalHistoryProps) => {
+ 
   const history = useMemo((): StepperStep[] => {
     const updates = proposal.proposalStatusUpdates
-    const createdStatus = { status: 'Created', createdAt: proposal.createdAt, inBlock: proposal.createdInBlock }
+    const createdStatus = { status: 'Deciding', createdAt: proposal.createdAt, inBlock: proposal.createdInBlock }
     const endStatus = !proposalActiveStatuses.includes(proposal.status) && [
       { status: proposal.status, createdAt: proposal.statusSetAtTime, inBlock: proposal.statusSetAtBlock },
     ]
--- a/packages/ui/src/proposals/components/ProposalHistory.tsx
+++ b/packages/ui/src/proposals/components/ProposalHistory.tsx
@@ -12,7 +12,6 @@ interface ProposalHistoryProps {
 }
 
 export const ProposalHistory = ({ proposal }: ProposalHistoryProps) => {
- 
   const history = useMemo((): StepperStep[] => {
     const updates = proposal.proposalStatusUpdates
     const createdStatus = { status: 'Deciding', createdAt: proposal.createdAt, inBlock: proposal.createdInBlock }
--- a/packages/ui/src/proposals/components/ProposalHistory.tsx
+++ b/packages/ui/src/proposals/components/ProposalHistory.tsx
@@ -14,11 +14,12 @@ interface ProposalHistoryProps {
 export const ProposalHistory = ({ proposal }: ProposalHistoryProps) => {
   const history = useMemo((): StepperStep[] => {
     const updates = proposal.proposalStatusUpdates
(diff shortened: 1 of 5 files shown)


packages/ui/src/proposals/hooks/useVotingRounds.ts
```diff
@@ -28,7 +28,7 @@ export const useVotingRounds = (votes: ProposalVote[] = [], updates: ProposalSta
   const councilSize = useCouncilSize()
 
   const voteRounds: (Omit<VotingRound, 'count'> & { total?: number })[] = useMemo(() => {
-    const decidingCount = updates.filter(({ status }) => status === 'deciding').length || 1
+    const decidingCount = updates.findIndex(({ status }) => status === 'deciding') + 1 || 1
```
thesan, line 31: I think this won't work.

To see how proposalStatusUpdates works check e.g:
```gql
{
  proposalByUniqueInput (where: { id: "63" }) {
    proposalStatusUpdates {
      newStatus { __typename }
    }
  }
}
```

You could use the index based on the assumption that only dormant and deciding stage alternate up to the last deciding stage. But that would be very unclear.

So I thinks this should be:
```suggestion
    const decidingCount = updates.filter(({ status }) => status === 'deciding').length + 1
```
Same as in `@/proposals/components/ProposalStages`

thesan (changes requested):
Also please check the failing CI

packages/ui/src/proposals/hooks/useVotingRounds.ts
chrlschwb: If the suggested code is implemented, `decidingCount` will be 1 forever because there is only one update with the status `deciding` at any given time. 
As soon as a new deciding stage is entered, the former `deciding` stage goes into `dormant`.

packages/ui/src/proposals/components/ProposalHistory.tsx
```diff
@@ -14,7 +14,7 @@ interface ProposalHistoryProps {
 export const ProposalHistory = ({ proposal }: ProposalHistoryProps) => {
   const history = useMemo((): StepperStep[] => {
     const updates = proposal.proposalStatusUpdates
-    const createdStatus = { status: 'Created', createdAt: proposal.createdAt, inBlock: proposal.createdInBlock }
+    const createdStatus = { status: 'Deciding', createdAt: proposal.createdAt, inBlock: proposal.createdInBlock }
```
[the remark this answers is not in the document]
chrlschwb: added a "Deciding" stage. 
The CI check is failing right after pulling from the "dev" branch. It is unrelated to this issue.

packages/ui/src/proposals/hooks/useVotingRounds.ts
thesan: No `update` keeps track of all stage changes.

Please try the query I wrote above on https://query.joystream.org/graphql

Because this proposal was voted on 3 times, it shows 2 `ProposalStatusDeciding` "newStatus" (as it doesn't include the initial deciding stage).

packages/ui/src/proposals/components/ProposalHistory.tsx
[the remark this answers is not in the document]
thesan: It looks related to me. The test suites failing are "useVotingRounds" and "ProposalPreview".

Check: https://github.com/Joystream/pioneer/actions/runs/4413195684/jobs/7733400617

thesan (approved):

Issue #2991, opened 2022-05-07
[USERSNAP] Restore Votes does not upate My Votes


dmtrjsg:
**Taken at**: https://dao.joystream.org/#/election
**Fields**:
- Houston, we have a problem.: Restore Votes does not upate My Votes
- Comment: The votes are restored from file but user has to reload the page manually.
- Feedback: l1dev

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/759150c8-afab-41b0-beaf-601fec21dc72-annotated_10120d69-d1c1-4a70-82fa-ef277bdb9f46.png?etag=7260a90a9eebc7d874b4db3082f1543b)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/759150c8-afab-41b0-beaf-601fec21dc72-annotated_10120d69-d1c1-4a70-82fa-ef277bdb9f46.png?etag=7260a90a9eebc7d874b4db3082f1543b)*

**Browser**: Chrome 100 (Linux)
**Screen size**: 1920x1080
**Browser size**: 1190x1216
**[Open #242 in Usersnap](https://app.usersnap.com/l/feedback/4f1936ef-9048-44cb-b924-bea7a523d7d4?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
Probably must be addressed in #2311. To be triaged.

ivanturlakov:
Tested on https://dao-git-carthage-joystream.vercel.app/#/election
wss://54.227.116.57.nip.io/ws-rpc

@traumschule @dmtrjsg  Looks like not fixed yet

⚠️ The votes are restored from file but user has to reload the page manually. Tested on Voting and Revealing stages.

To test this case, I voted for a few candidates then backup votes, and then deleted votes from Local Storage.

traumschule:
There's a fix in #3868 by @mkbeefcake.

Issue #2986, opened 2022-05-06
After voting `My Votes` is not updated


traumschule:
With the modal that confirms that i successfully voted make another QN request to update my votes.

Currently user has to reload manually.

dmtrjsg:
@traumschule is this for proposals? If so same as #2896 

Pls confirm

mkbeefcake:
@isonar can you please assign this issue to me?

polikosi:
tested on: https://pioneer-2-git-fork-mkbeefcake-fix-myvotesisntupdated-joystream.vercel.app/#/settings?network-config=https://54.205.24.16.nip.io/network/config.json

 The `My Votes` counter increases and the votes list is updated.
![image](https://user-images.githubusercontent.com/80429191/220093672-3db10668-5b3b-4444-abbc-993fefec8318.png)

polikosi:
tested on: https://pioneerapp.xyz/
I voted for "tomato" but the `Vote` button didn't change to `Vote again` and the `My Votes` counter is increasing but the vote list is not updating.
![image](https://user-images.githubusercontent.com/80429191/228893669-f744fd77-9f2b-408f-8344-1b54ee600d56.png)
![image](https://user-images.githubusercontent.com/80429191/228894723-02357212-2f79-44ba-98a2-a8765f1f13e7.png)

[TX](https://polkadot.js.org/apps/?rpc=wss://rpc.joystream.org:9944#/explorer/query/0xff0c2fd2f0d15769dc616c25d9e254d4575e200837de607440a3e7ef28afe72c)

dmtrjsg:
This probably needs to get a revival..

PR #4262, opened 2023-03-23
Update MyVotes (#2991, #2986)

chrlschwb:
fixes issues #2991, #2986

### DIFF
--- a/packages/ui/src/common/hooks/useLocalStorage.ts
+++ b/packages/ui/src/common/hooks/useLocalStorage.ts
@@ -42,7 +42,7 @@ export const useLocalStorage = <T>(key?: string) => {
   }, [key])
 
   useEffect(() => {
-    const handleEventOnce = (event: any) => {
+    const handleEventOnce = () => {
       setState(getItem(key))
     }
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/common/hooks/useLocalStorage.ts
```diff
@@ -41,11 +41,22 @@ export const useLocalStorage = <T>(key?: string) => {
...
+      document.removeEventListener(`storage_event_${key}`, handleEventOnce)
+    }
+  }, [])
+
   const dispatch = useCallback(
     (setStateAction: T | ((prevState?: T) => T)) => {
       const value = isFunction(setStateAction) ? setStateAction(getItem(key)) : setStateAction
-      setState(value)
       setItem(key, value)
+      document.dispatchEvent(new CustomEvent(`storage_event_${key}`, {}))
```
traumschule, line 59: From the code it's not obvious what we are saving and why. Maybe add a comment?

traumschule (changes requested):
1. #2991: I don't see an automatic update after restoring. After switching to another page and back to election restored votes are shown correctly.
2. #2986: This seems to work on https://pioneerapp.xyz already?

PR #4270, opened 2023-03-27
fixes #3968, fixes #4129 fix autoscroll to issue

### LINKED ISSUE #3968: Proposal discussion auto-scroll distraction

### LINKED ISSUE #4129: [USERSNAP] Forum post auto-scrolling bug
**Taken at**: https://pioneerapp.xyz/#/proposals/preview/81?post=69
**Fields**:
- Houston, we have a problem.: Forum post auto-scrolling bug
- Comment: When I follow a link to a specific post, like https://pioneerapp.xyz/#/proposals/preview/81?post=69, then if I try to scroll to another part of the page I will be auto scrolled back to the post after a few seceonds.

SCREENSHOT
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/0f8b195b-3502-4893-9a24-4265c5e149be-annotated_8aff3c26-0b7f-4043-a40d-ddc523022ed8.png?etag=d102f9b1395b2f82df877f42c764d492)*

**Browser**: Firefox 109 (macOS Catalina)
**Screen size**: 5120x1440
**Browser size**: 1858x1171
**[Open #384 in Usersnap](https://app.usersnap.com/l/feedback/302c1b22-0355-47af-94c0-5ccbddfeb1c1?utm_medium=integration&utm_campaign=integration&utm_content=open)**

## Technical guidance:

The `` is happening [here](https://github.com/Joystream/pioneer/blob/0d308b26de798e8d41b8a7680b18b395ec42f03e/packages/ui/src/forum/components/PostList/PostListItem.tsx#L71):
```ts
useEffect(() => {
  if (ref.current && isSelected) {
    ref.current.scrollIntoView({ behavior: 'smooth', inline: 'start' })
  }
}, [location, ref.current])
```
I'm not 100% sure which of the 2 `useEffect` dependencies is at fault but I don't think `ref.current` should be here.

IlyaSmiyukha:
#3968 and #4129 same issue
@chrlschwb check pls

### DIFF
--- a/packages/ui/src/forum/components/PostList/PostListItem.tsx
+++ b/packages/ui/src/forum/components/PostList/PostListItem.tsx
@@ -67,7 +67,7 @@ export const PostListItem = ({
     if (ref.current && isSelected) {
       ref.current.scrollIntoView({ behavior: 'smooth', inline: 'start' })
     }
-  }, [ref.current])
+  }, [isSelected])
 
   const [editing, setEditing] = useState(false)
   const editionTime = useMemo(() => {
--- a/packages/ui/src/forum/components/PostList/PostListItem.tsx
+++ b/packages/ui/src/forum/components/PostList/PostListItem.tsx
@@ -14,7 +14,6 @@ import { ArrowReplyIcon, LinkIcon, ReplyIcon } from '@/common/components/icons'
 import { MarkdownPreview } from '@/common/components/MarkdownPreview'
 import { Badge } from '@/common/components/typography'
 import { BorderRad, Colors, Fonts, Shadows } from '@/common/constants'
-import { useLocation } from '@/common/hooks/useLocation'
 import { useModal } from '@/common/hooks/useModal'
 import { relativeIfRecent } from '@/common/model/relativeIfRecent'
 import { PostHistoryModalCall } from '@/forum/modals/PostHistoryModal'
@@ -58,7 +57,6 @@ export const PostListItem = ({
   const { active } = useMyMemberships()
   const { createdAtBlock, lastEditedAt, author, text, repliesTo } = post
   const [postLastEditedAt, setPostLastEditedAt] = useState<string | undefined>(lastEditedAt)
-  const location = useLocation()
   const { showModal } = useModal()
   const ref = useRef<HTMLDivElement>(null)
   useEffect(() => {
@@ -69,7 +67,7 @@ export const PostListItem = ({
     if (ref.current && isSelected) {
       ref.current.scrollIntoView({ behavior: 'smooth', inline: 'start' })
     }
-  }, [location, ref.current])
+  }, [ref.current])
 
   const [editing, setEditing] = useState(false)
   const editionTime = useMemo(() => {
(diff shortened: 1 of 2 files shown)


chrlschwb (approved):
ok

packages/ui/src/forum/components/PostList/PostListItem.tsx
```diff
@@ -66,8 +67,9 @@ export const PostListItem = ({
   }, [ref.current])
 
   useEffect(() => {
-    if (ref.current && isSelected) {
+    if (ref.current && isSelected && firstLoad) {
       ref.current.scrollIntoView({ behavior: 'smooth', inline: 'start' })
+      setFirstLoad(false)
     }
   }, [location, ref.current])
```
thesan, line 74: This solution would work but it add some unnecessary complexity.

Instead AFAICS both `location` and `ref.current` are the wrong dependencies `useEffect`.
1. `location` is actually the source of the problem because the `useLocation` hook returns a new JS object (same values different reference) on every rerender.
     Instead it should be:
    ```ts
    return useMemo(() => ({ pathname, search, hash, origin }), [pathname, search, hash])
    ```
     But in this case we should only care about `isSelected` (which itself depends on the location).

3. `ref.current` is assigned to the `ForumPostBlock` `div` before any of the `useEffect` get called, so it's useless to add it in the dependencies.
```suggestion
    if (ref.current && isSelected) {
      ref.current.scrollIntoView({ behavior: 'smooth', inline: 'start' })
    }
  }, [isSelected])
```
**Improtant:**
Please fix the `useLocation` hook (as described above) because it might cause unnecessary re-renders in other places and maybe some other bugs.

packages/ui/src/forum/components/PostList/PostListItem.tsx
```diff
@@ -58,6 +58,7 @@ export const PostListItem = ({
   const { active } = useMyMemberships()
   const { createdAtBlock, lastEditedAt, author, text, repliesTo } = post
   const [postLastEditedAt, setPostLastEditedAt] = useState<string | undefined>(lastEditedAt)
+  const [firstLoad, setFirstLoad] = useState<boolean>(true)
   const location = useLocation()
```
thesan, line 62:
```suggestion
```

thesan (changes requested):

packages/ui/src/forum/components/PostList/PostListItem.tsx
IlyaSmiyukha: You right. Resolved

IlyaSmiyukha: Done. Sorry misunderstood this part.

thesan: Np!

thesan (approved):
:100: 

Thanks again !

PR #4318, opened 2023-04-03
fixes #4108 Forum returns 404 for correct links

### LINKED ISSUE #4108: Forum returns 404 for correct links
Sometimes this happens and one has to open the same link again which then works.

STR
1. open a forum thread
2. edit the thread number in the address bar
3. hit enter
=> 404

IlyaSmiyukha:
fixes #4108

### DIFF
--- a/packages/ui/src/app/pages/Forum/ForumThread.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumThread.tsx
@@ -31,7 +31,7 @@ import { ForumPageLayout } from './components/ForumPageLayout'
 
 export const ForumThread = () => {
   const { id } = useParams<{ id: string }>()
-  const { isLoading: isLoadingThread, thread } = useForumThread(id)
+  const { isLoading: isLoadingThread, thread, hasError } = useForumThread(id)
   const isRefetched = useRefetchQueries({
     interval: MILLISECONDS_PER_BLOCK,
     include: ['GetForumThreads'],
@@ -60,7 +60,7 @@ export const ForumThread = () => {
   }
 
   useEffect(() => {
-    if (!isLoading && !thread) {
+    if (!isLoading && !thread && !hasError) {
       history.replace('/404')
     }
   }, [isLoading, thread])
--- a/packages/ui/src/app/pages/Forum/ForumThread.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumThread.tsx
@@ -14,9 +14,7 @@ import { PinIcon } from '@/common/components/icons/PinIcon'
 import { MainPanel, RowGapBlock } from '@/common/components/page/PageContent'
 import { PreviousPage } from '@/common/components/page/PreviousPage'
 import { Colors } from '@/common/constants'
-import { useRefetchQueries } from '@/common/hooks/useRefetchQueries'
 import { createType } from '@/common/model/createType'
-import { MILLISECONDS_PER_BLOCK } from '@/common/model/formatters'
 import { metadataToBytes } from '@/common/model/JoystreamNode'
 import { getUrl } from '@/common/utils/getUrl'
 import { PostList } from '@/forum/components/PostList/PostList'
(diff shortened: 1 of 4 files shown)


chrlschwb (approved):

thesan (changes requested):
@IlyaSmiyukha to QA this and simulate the QN going down without needing to run a local network I used FoxyProxy.

But the page still redirect to a 404 instead of showing an unending loading screen or an error message. Which would both be an improvement on the current behavior:

[Peek 2023-04-07 16-59.webm](https://user-images.githubusercontent.com/6571453/230630773-0265edf0-4e24-4467-b5b8-af85e337d3a8.webm)

IlyaSmiyukha:
Will check it

IlyaSmiyukha:
@thesan 
Screenshot 2023-04-11 at 13 08 45

I made a fix for handling errors for threads. Right now it shows an infinity loader. We can change it to an error message like this.  But I need a text for the message.

packages/ui/src/common/hooks/useRefetchQueries.ts
```diff
@@ -38,6 +38,7 @@ export const useRefetchQueries = (
         }, interval)
 
         return () => {
+          isRefetched.current = false
```
thesan, line 41: @IlyaSmiyukha I don't think this change is right. AFAIK `isRefetched` is just supposed to indicate that the data was already fetched once. So once it has been set to `true` it should remain this way. Or am I missing something ?

thesan (changes requested):
The loader would be a good option if Apollo was re-trying on error which doesn't seem to be the case (maybe there is a setting for it).

Alternatively could we have something like:
"Something went wrong fetching this thread: {Error message returned by the QN}"

WDYT @dmtrjsg ?

dmtrjsg:
@thesan sounds good 👍

packages/ui/src/common/hooks/useRefetchQueries.ts
IlyaSmiyukha: For the current post you are right, But when the user opens another post `isRefetched` will return true. And in [this case ](https://github.com/Joystream/pioneer/blob/e26105af6e27c36142c3ef748c00cd85ea057dc6/packages/ui/src/app/pages/Forum/ForumThread.tsx#L38) he will be redirected to 404 page. So when the URL was changed we should update isRefetched as well

thesan (commented):
@IlyaSmiyukha

packages/ui/src/app/pages/Forum/ForumThread.tsx
```diff
@@ -108,8 +109,14 @@ export const ForumThread = () => {
 
   const displayMain = () => (
     <ThreadPanel ref={sideNeighborRef}>
-      <PostList threadId={id} isThreadActive={isThreadActive} isLoading={isLoading} />
-      {thread && isThreadActive && <NewThreadPost ref={newPostRef} getTransaction={getTransaction} />}
+      {hasError ? (
+        <EmptyPagePlaceholder title="Something went wrong fetching this thread" copy="" button={null} />
```
thesan, line 113: Small detail:
```suggestion
        <EmptyPagePlaceholder title="Something went wrong fetching this thread." copy="" button={null} />
```

IlyaSmiyukha:
> @IlyaSmiyukha

I will recheck it one more time

IlyaSmiyukha:
https://user-images.githubusercontent.com/39830587/232200776-2a1c02fd-5345-43ea-be4b-6d32a08e9fdb.mov

@thesan 
Pay attention pls to the console log. I removed the fix and print what return refetched on the page change. It should return false, but the hook still returns true. If be honest not sure why.

thesan (approved):
Hey @IlyaSmiyukha you were right there was a lot going wrong with the query refetching: First the `isRefetched` was set to `true` before it got fetched in the first place, second there is no `GetForumThreads` query on this page and finally the posts should be polled not the thread itself. Also the `useRefetchQueries` is called almost everywhere without dependencies, which works but is bad for performances, but then I think that's the intended behavior in some places, so it shouldn't be changed everywhere blindly.

Since there was a lot to fix I committed the fixes directly on this PR. Hopefully it doesn't break other component using the same hook :crossed_fingers: 

Good job catching the bug here :raised_hands: 

@chrlschwb could someone give this PR one last review before I merge it please? Just in case I missed something in the code I wrote.

IlyaSmiyukha:
It's great! Thanks @thesan

chrlschwb (approved):
All the stuff you mentioned are in the code.

Issue #4277, opened 2023-03-28
Add backend API tests


thesan:
Will be handled by JSG

PR #4337, opened 2023-04-09
📐 Test the API integration

thesan:
- #4277
Depends on #4322

[Readme update](https://github.com/Joystream/pioneer/blob/feature/backend-api-tests/packages/server/README.md)


packages/server/test/setup.ts
```diff
@@ -1,7 +1,14 @@
-import log from 'npmlog'
+import { prisma } from '@/common/prisma'
 
-import '@/common/config'
-log.level = 'silent'
+process.env.APP_SECRET_KEY = 'foo'
```
traumschule, line 3: Maybe worth a README hint to change this for improved security.

packages/server/.env.dev
```diff
@@ -33,12 +31,12 @@ APP_LOG_LEVEL: "verbose"
 # Email (only one of the api provider should be defined)
 #
 
-EMAIL_SENDER: "test@example.com" # This is only needed when SENDGRID_API_KEY is defined
+# EMAIL_SENDER: "test@example.com" # This is only needed when SENDGRID_API_KEY is defined
 
-SENDGRID_API_KEY: "SG.1234567890qwerty"
+# SENDGRID_API_KEY: "SG.1234567890qwerty"
```
traumschule, line 36: Are there other options than sendgrid, like local postfix or custom smpt credentials?

traumschule (approved):
Code looks ok, without checking functionality.

packages/server/test/setup.ts
thesan: @traumschule thank you for the review.

This value will only be used for running the integration tests. I might be missing something but I don't see how it can be a security risk. However if you still think it's the case it could be randomized and imported in the tests where it's needed.

packages/server/.env.dev
thesan: We decided to only focus on sendgrid for the MVP (because of the big free tier and it's probably what JSG will end up using) but there is already an issue to add more options soon after: #4285 (which I just updated).

Issue #4348, opened 2023-04-12
Proposal with Constitutionality 2 is not showing multiple tabs


dmtrjsg:
## Context
https://discord.com/channels/811216481340751934/1035125653722365993/1095732012914069664

## Details
![Screenshot 2023-04-12 at 19.32.00.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/4d159f5d-739c-4380-a73d-5a2dd3d5ea78)
![Screenshot 2023-04-12 at 19.32.08.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/9c1e9337-e88b-4342-ace9-7ef9b50f5115)

thesan:
It looks like we forgot to update the condition to display the tabs in [`1.3.1`](https://github.com/Joystream/pioneer/releases/tag/v1.3.1).

It's still `proposal.status === 'dormant' || votingRounds.length > 1` so the tabs don't show during the first deciding stage.

dmtrjsg:
Oh nice catch, thanks for triaging this so quickly! Let's try to patch it when we can 👍

chrlschwb:
I will fix it in a moment.

ivanturlakov:
Tested on https://dao-git-dev-joystream.vercel.app/#/proposals/current
wss://54.205.24.16.nip.io/ws-rpc

@chrlschwb do I use a correct env?

Proposal created
<img width="1153" alt="Screenshot 2023-05-01 at 15 48 50" src="https://user-images.githubusercontent.com/46903215/235453771-b86d0ece-5ba6-4520-9ece-6884cc248881.png">

404 when trying to open a proposal page
<img width="1440" alt="Screenshot 2023-05-01 at 15 48 27" src="https://user-images.githubusercontent.com/46903215/235453840-8001b075-259f-4b49-b5a4-5c9b50742168.png">

chrlschwb:
What is this endpoint? wss://54.205.24.16.nip.io/ws-rpc
The error shows that "UpdateChannelPayoutProposal" which was introduced in Ephesus isn't recognized. I think the QN you are using is outdated.

ivanturlakov:
Tested on 
https://pioneer-2-git-fork-chrlschwb-fix-4348-proposal-a7b1be-joystream.vercel.app/#/proposals/preview/289
https://dao-git-dev-joystream.vercel.app/#/proposals/preview/289
Mainnet

Error when trying to open a proposal page
[...]

PR #4350, opened 2023-04-13
Fix/4348 proposal constitutionaly2

chrlschwb:
Fixes issue #4348


chrlschwb:
"proposal.status === 'dormant' || votingRounds.length >= 1" is have ci-test error. The error is 
image

packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
```diff
@@ -158,7 +158,7 @@ export const ProposalPreview = () => {
             </BadgeAndTime>
           </RowGapBlock>
 
-          {(proposal.status === 'dormant' || votingRounds.length > 1) && (
+          {(proposal.status === 'dormant' || votingRounds.length >= 1) && (
```
thesan: This is not the right way.

Now the tabs should be shown for any proposal with a `constitutionality > 1`  so `proposal.status` and `votingRounds.length` is irrelevant.

thesan (changes requested):

chrlschwb:
@thesan I've changed the code to "constitutionality > 1" but the result isn't as we expect.

chrlschwb:
constants?.constitutionality = undefined

thesan (approved):
LGTM!

Issue #4236, opened 2023-03-14
Email notifications: Backend unavailable errors


dmtrjsg:
## Context
- #4230 

## Scope
- [ ] Implement modal to display error when BE is unavailable

Before sending conf email
After following the conf email
When "save" settings is clicked on the settings page after adding any modifications to the page

## Design

[Link to designs> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11947%3A504166&t=kbnLvo8CmtA4rR1I-1)

Preview: 
<img width="608" alt="Screenshot 2023-03-30 at 18 46 47" src="https://user-images.githubusercontent.com/93924525/228874511-94790637-251c-45f8-bef1-aa31e9398b54.png">

AnttoLion:
Hello. I have created an error modal component. Could someone please guide me on how to capture the event described in this issue? For instance, I discovered that the ExtensionNotification is triggered by the code "const { error } = useMyAccounts()" in the ExtensionWarning.tsx file.

thesan:
Hello @AnttoLion the goal of this issue is only to have the modal available for later. It will be displayed for example when on the last step of the signup flow (described here: #4235) the mutation fails due to connection issues.

PR #4358, opened 2023-04-20
Fix/4236 email noti be unavailable errors

AnttoLion:
#4236 email notification modal - backend unavailable errors


chrlschwb (approved):
ok

Issue #4238, opened 2023-03-14
Resubscribe to notifications when BE token expired


dmtrjsg:
## Context
- #4230 

## Scope
- [ ] Add Resubscribe to notifications flow when BE is expired case

## Design
[Link to user story when BE token expires > ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11958%3A544281&t=kbnLvo8CmtA4rR1I-1)

Preview: 
<img width="833" alt="Screenshot 2023-03-30 at 18 56 42" src="https://user-images.githubusercontent.com/93924525/228877917-91a5a01a-0291-41f7-891b-237269e67183.png">

PR #4361, opened 2023-04-21
Fix/4238 resubscribe notification token expired

RAISEMEUP007:
Resubscribe to notifications when BE token expired #4238


chrlschwb (approved):
ok

thesan (changes requested):
ATM I don't really understand why 2 components are needed to show this notification. Could you add a story (for at least one of the components) ?

Also please make sure the `WarnedIconLine` shows here: https://pioneer-2-storybook-git-fork-anttolion-fix-423-18260a-joystream.vercel.app/?path=/story/common-icon--default

AnttoLion:
Hello @thesan, I apologize for disturbing you. Could you let me know more details about it? I always agree with you, but I am having difficulty understanding the comment you left.

AnttoLion:
To follow @thesan's comment, it was necessary to modify the SettingsInformation component. I hope my decision is correct.

thesan:
@AnttoLion sorry I completely missed your question from 3 weeks ago. 

> modify the SettingsInformation component. I hope my decision is correct.

Yes this solution is good IMO.

But also please use a named export for the `AuthenticationInfo` component:
```tsx
export const AuthenticationInfo: ...
```
Because it's the project convention.

Finally activities are these for example: 
image

So the `WarnedIconLine` is not really an activity icon. Instead please re-use `@/common/components/icons/WarningIcon` and:
1. replace the `fill` attribute by "currentColor"
2. export this icon in `@/common/components/icons/index.ts` (it's missing for some reason)
3. remove the `typeof Component === 'function'` from the `if` in `@/common/components/icons/icons.stories`
4. Finaly in `@/common/modals/OnBoardingModal/OnBoardingModal` `OnBoardingTextFooterWrapper` should have a:
   ```css
   ${WarningIcon} {
     color: #FFAA02;
   }
   ```

PR #4364, opened 2023-04-21
🛰️ Backend deployments

thesan:
A PR to add different needed deployments for Pioneer backend.

### Development docker image usage
```shell
docker run --rm -it -e POSTGRES_PASSWORD="my-password123" --name api -p 3000:3000 thesan/pioneer-backend
```
### Production blueprint
[Deploy to Render](https://render.com/deploy?repo=https://github.com/Joystream/pioneer/tree/feature/backend-deployments)

The notify script can be run with: `docker exec api notify`
> **Warning**
> A `STARTING_BLOCK` with a recent block height value should be passed to avoid processing the whole chain QN events when running `notify`.

### TODO
- [x] Create a development image
- [x] Create a production image

### Out of scope
- [ ] #4282

### To finalize later
- [ ] Set-up a JSG deployment


chrlschwb (approved):

backend.dev.Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM node:alpine3.17 AS builder
```
kdembler, line 1: I think we should specify Node version we want this to run on

backend.dev.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+WORKDIR /app
+
+COPY packages/server/package.json ./
+COPY yarn.lock ./
+RUN yarn remove node-cron @joystream/types
```
kdembler, line 6: Why is this needed?

backend.dev.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+COPY packages/server/prisma ./
+COPY packages/server/codegen.ts ./
+COPY packages/server/src ./src
+RUN yarn build:local
+
+RUN rm -rf node_modules
+ENV NODE_ENV=production
+RUN yarn --prod --immutable
+RUN yarn prisma generate
```
kdembler, line 21: Why is this double `yarn` install necessary?

backend.dev.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+FROM postgres:alpine3.17
```
kdembler, line 23: As I understand, this single image is supposed to operate both DB and the API for development. IMO this is a bit unintuitive and personally I'd prefer them separate and managed with `docker-compose`, like you would do for production setup.

backend.prod.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN rm -rf node_modules
+ENV NODE_ENV=production
+RUN yarn --prod --immutable
+RUN yarn prisma generate
+
+FROM alpine:3.17
+WORKDIR /app
+RUN set -eux; \
+    adduser -u $(getent group www-data | cut -d: -f3) -S -D -G www-data www-data; \
+    apk add --no-cache nodejs
```
kdembler, line 27: Why build own nodejs image when there are prebuilt available?

backend.dev.Dockerfile
thesan: I think the build would fail with these included but I don't remember what was the exact issue.

backend.dev.Dockerfile
thesan: Basically to reduce the final image size:
Some packages are huge due to their dependencies especially prisma however due to prisma it's really hard to bundle the code without the `node_modules` in the final image. So the first install create a `node_modules` with all the dev dependencies which also includes what's needed by the [`yarn build:local`](https://github.com/Joystream/pioneer/blob/backend/packages/server/package.json#L29C6-L29C11) script, then a much smaller `node_modules` is installed with only what is needed to run the code and the migrate db. Because the first install is done in a build stage it doesn't increase the image size.

I think the size reduction from this is pretty big the first version of the image was close to 1GB.

backend.dev.Dockerfile
thesan: It makes sense but unless I forgot about something specific to the `:prod` image I think we could have a `docker-compose` file for development that uses this `:prod` image. But what I appreciate with the all bundled dev image is that it allows less experience developers to get started with a one liner (`docker run --rm -it -e POSTGRES_PASSWORD="my-password123" --name api -p 3000:3000 thesan/pioneer-backend`) regardless of the state of their local repository. This was really an issue back when I opened this PR (with only the development image).

But again I think it would make sense to add the `docker-compose` file add switch the `:latest` tag to the production image once we move it to `joystream/pioneer-backend`. WDYT ?

backend.prod.Dockerfile
thesan: I think to reduce the size again (the `node:alpine` images are bigger than manually installing nodejs) but TBH I don't remember how significant that was I might have gone overboard here !

backend.dev.Dockerfile
thesan: Yeah I agree... `node:18-alpine3.17` ?

chrlschwb: node-v20.8.1

thesan: @chrlschwb it would be nice but v20 hasn't been released as a LTS yet it should happen soon but rn v18 is still the most recent LTS

backend.dev.Dockerfile
thesan: It didn't fail anymore ! I removed it now.

backend.dev.Dockerfile
thesan: I added the `docker-compose.yml` and made the `:prod` image the latest.

backend.prod.Dockerfile
thesan: The gain was `118MB` but I think it's still worth using prebuilt image to make sure the node version is the same as the one it builds from.

packages/server/docker-compose.yml
```diff
@@ -1,14 +1,27 @@
...
+      context: ../..
+      dockerfile: backend.Dockerfile
+    ports: ['${PORT}:${PORT}']
+    depends_on: [pioneer-db]
+    env_file: [.env] # ensure `.env` exist
+    environment:
+      DATABASE_URL: postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@pioneer-db:5432/${POSTGRES_DB}
+
+  pioneer-db:
     image: postgres:12
```
kdembler, line 24: Haven't noticed earlier but I think we could use a newer version? 12 is 4 years old already, 16 was released recently, wondering if it would work?

packages/server/docker-compose.yml
```diff
@@ -1,14 +1,27 @@
...
-    ports:
-      - '127.0.0.1:${DB_PORT}:5432'
-    volumes:
-      - db-data:/var/lib/postgresql/data
-    env_file:
-      - .env
+    ports: ['127.0.0.1:${DB_PORT}:5432']
```
kdembler, line 25: Does the DB need to be exposed outside of docker-compose?

packages/server/docker-compose.yml
```diff
@@ -1,14 +1,27 @@
...
+    external: true
 
 services:
-  db:
+  pioneer-api:
+    image: thesan/pioneer-backend # TODO change this to joystream/...
+    build:
+      context: ../..
+      dockerfile: backend.Dockerfile
+    ports: ['${PORT}:${PORT}']
```
kdembler, line 17: Haven't seen this syntax used in docker-compose yet, always the regular list
```
ports:
  - '${PORT};${PORT}'
```
Super small, don't bother to change but I find the above syntax easier to parse automatically :D

packages/server/docker-compose.yml
thesan: Right it should I'll set it to 16

packages/server/docker-compose.yml
thesan: I think it's better as `yarn workspace server docker:up` can be used to quickly work develop on the frontend (using a local backend). But to actually work on the backend this [Run locally](https://github.com/Joystream/pioneer/blob/feature/backend-deployments/packages/server/README.md#run-locally) guide is better (and still very simple) but it runs the api on a local network but the db with docker.

Which makes me realized I forgot to change the service name to `pioneer-db` on all these scripts !

packages/server/docker-compose.yml
thesan: Yeah I tend to favor terser syntax to a fault !

Issue #4230, opened 2023-03-14
Email Notifications Dev Epic


dmtrjsg:
# Context
- #2880

## Design
Pls see set of all related user stories below the section where the link takes you to.

👉 🖼️ [Link to Figma> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=3600%3A73&t=LsnFZyLXE94cDQkb-1)

## Back End 
👉 🛠️  Link to BE work done already: 
- https://github.com/Joystream/pioneer/pull/4210

👉 📓  Link to BE readme: 
- [Link > ](https://github.com/Joystream/pioneer/blob/feature/backend-poc/packages/server/README.md)

## Scope

⚠️ Important to address the issues in the exact order they are laid out here. FE issues are linked with the pre-requisite BE issues. Pls pay attention to the "blocking" and "blocked by"  sections in the description of each issue. 
___________
### MVP 🎯 
Sign up to notifications from new memberships creation flows
Sign up to notifications from settings module
Notifications are not customaseable
Simple Text Email Template Used
Only email notifications supported
Notification events are _not customisable_ - forum direct mentions and council election stages update only

👉 🖼️ [Designs for settings page for MVP > ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=12215%3A461709&t=LsnFZyLXE94cDQkb-1)

#### BE

- [x] Forum direct mentions
- [x] #4273
- [x] #4274
- [x] #4275
- [x] #4278

#### FE

- [x] #4232
- [x] #4231
- [x] #4236
- [x] #4235
- [x] #4237
- [x] #4238
- [ ] #4309
- [ ] #4310

___________

#### BE

- [ ] #4279
- [ ] #4280
- [ ] #4281

#### FE

- [ ] #4234

___________
#### BE
- [x] #4282
- [ ] #4283
- [ ] #4284
- [ ] #4285

#### FE
- [ ] #4304

___________
#### BE

**Role based**
- [ ] #4286
- [ ] #4287
- [ ] #4288
- [ ] #4289

Last day before event x
- [ ] #4290
- [ ] #4291
- [ ] #4292
- [ ] #4293

Only voters for revealing stage notiffs
- [ ] #4294

#### FE

- [ ] #4303
- [ ] #4302
- [ ] #4233

#### Will be handled by JSG
- [x] #4276
- [x] #4277
- [x] #4252

[...]

PR #4365, opened 2023-04-24
🔌 Add an Apollo client instance for the backend

thesan:
Add a `useBackend` hook to connect to the backend. See the usage example: [`useIsMemberRegistered`](https://github.com/Joystream/pioneer/blob/feature/backend-connection/packages/ui/src/common/hooks/useIsMemberRegistered.ts)

This might not be the best/final solution but it should be good enough to get started implementing #4230.

> **Warning**
> The connection endpoint should be set in the `package/ui/.env` file. When using the development backend deployment from #4364:
> ```
> docker run --rm -it -e POSTGRES_PASSWORD="my-password123" --name api -p 3000:3000 thesan/pioneer-backend
> ```
> The mainnet backend endpoint should: `REACT_APP_MAINNET_BACKEND=http://localhost:3000`


chrlschwb (approved):
ok

Issue #4309, opened 2023-03-30
Watch forum categories/ threads/ proposals/ working groups


dmtrjsg:
## Context
- #4230 

## Design

⚠️ The design of the email templates is already implemented here:
- https://github.com/Joystream/pioneer/issues/4274

[Watching proposals> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11947%3A504136&t=kbnLvo8CmtA4rR1I-1)
[Watching Forum categories> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11947%3A504157&t=kbnLvo8CmtA4rR1I-1)
[Watching forum threads> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11947%3A504160&t=kbnLvo8CmtA4rR1I-1)
[Watching Working groups > ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11947%3A504154&t=kbnLvo8CmtA4rR1I-1)

⚠️ NB: there's dependency between watching and muting entities in terms of interactions, which is[ covered here >](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=12275%3A467690&t=kbnLvo8CmtA4rR1I-1) 

## Scope
- [ ] Implement "watching" 
-- Forum categories
-- Forum threads
-- Proposals
-- Working groups

When user watches entity they get all updates in the activities of this entity

1️⃣ Forum categories: 
new threads created

Email template:

`Hi {Name},
New thread created in the category you are watching: {name_of_forum_category}.'

CTA = View on Forum

2️⃣ Forum threads:
new posts and replies

Email template: 

`Hi {Name},
New post created in the forum thread you are watching: {forum_Thread_name}`

CTA = View on Forum

3️⃣ Proposals:
proposals votes by council

`Hi {Name},
New vote casted on proposal you are watching: {proposal_name; Proposal_ID}. Voter: {member_handle}. Vote: {type_of_vote_casted}`

CTA = View Proposal

proposals changed state

`Hi {Name},
Proposal you are watching: {proposal_name; Proposal_ID} changed state to {new_state_name}.`

CTA = View Proposal

Proposal discussion post

`Hi {Name},
[...]

PR #4376, opened 2023-05-04
Fix/4310 from fix/4309 mute forum categories threads

RAISEMEUP007:
It is not necessary to handle issue/4309 separately. In order to complete issue/4310, #4309 must be completed and the hooks required for both issues are different from what is needed for the independent case.


chrlschwb (commented):
Looks good, but it cannot be completed before https://github.com/Joystream/pioneer/pull/4337 is ready.

Issue #2804, opened 2022-04-06
Update the scripts to populate data on the chain


dmtrjsg:
Add these to this ticket.

PR #4383, opened 2023-05-17
🗝️ Fix `opening:create` on the liberated chain

thesan:
Related to https://github.com/Joystream/pioneer/issues/2804


traumschule (approved):
Can't test right now but looks useful :)

PR #4385, opened 2023-05-18
Fix/4032 ckedittable fix update

### LINKED ISSUE #4032: Forum: Editing Tables
CkEditor isn't able to help producing tables. However it can be used to save a a post including a correctly formatted table.

On editing that post however CkEditor breaks that table while it should just keep its raw format as it is.

table
editing-table
Thread ID: 2

chrlschwb:
fixes #4032


packages/ui/src/common/components/MarkdownPreview/MarkdownPreview.tsx
```diff
@@ -75,10 +76,11 @@ export const MarkdownPreview = ({ markdown, append, ...styleProps }: MarkdownPre
   }, [markdown, append])
 
   const stripBackslashes = (text: string) => text.replace(/\\(.)/gm, '$1')
+
   return (
     <div className="markdown-preview">
       <MarkdownPreviewStyles {...styleProps} />
-      <ReactMarkdown rehypePlugins={rehypePlugins} remarkPlugins={[remarkGfm]} components={components} rawSourcePos>
+      <ReactMarkdown rehypePlugins={[rehypeRaw]} remarkPlugins={[remarkGfm]} components={components} rawSourcePos>
```
thesan, line 83: @chrlschwb for security using `rehype-raw` is not an acceptable solution.

Remember that posts (for example) don't need to come from Pioneer, they can be freely written directly on chain. This way (almost) anything can be returned from the QN.

I was able to inject `<style>` tags to completely change the look of the page and any other kind of HTML. An attacker could easily use this to bring Pioneer users to a website they control, e.g using CSS to hide some navigation links and replace them by links from the post (with some clever absolute positioning). The only thing I wasn't able to do is run scripts, but even this might be doable somehow.

Also in case you didn't notice this PR is based on backend work.

thesan (changes requested):

traumschule:
Parent is done. Also `for security using rehype-raw is not an acceptable solution.`

PR #4406, opened 2023-06-02
⬆️ Upgrade to Storybook 7

thesan:
> **Note**
> `node >= 16` is now required to run Storybook

- #4402


packages/markdown-editor/package.json
```diff
@@ -44,7 +44,7 @@
     "ts-loader": "^9.2.6",
     "ts-node": "^10.4.0",
     "typescript": "4.4.3",
-    "webpack": "^5.64.1",
```
traumschule, line 47: Just curious, was there are reason to downgrade webpack?

traumschule (approved):
This needs to be added to the Changelog / release notes:
> 7.0 breaking changes
> Dropped support for Node 15 and below
> Storybook 7.0 requires Node 16 or above. If you are using an older version of Node, you will need to upgrade or keep using Storybook 6 in the meantime.
[Storybook 7](https://storybook.js.org/docs/react/migration-guide) [migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-65x-to-700)

packages/markdown-editor/package.json
thesan: Actually webpack was upgraded:
image

It's just that there was initially duplicated versions of webpack after upgrading storybook. `yarn dedupe` would have taken care of that but IMO it's just as good to just force the major version of webpack (unless we realized later that some particular minor version breaks the build).

thesan:
> This needs to be added to the Changelog / release notes:
> 
> > 7.0 breaking changes
> > Dropped support for Node 15 and below
> > Storybook 7.0 requires Node 16 or above. If you are using an older version of Node, you will need to upgrade or keep using Storybook 6 in the meantime.
> > [Storybook 7](https://storybook.js.org/docs/react/migration-guide) [migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-65x-to-700)

@traumschule nice catch I should remember to mention it in the release notes when this gets released.

At the same time I don't think `engines.node` should be updated yet because the version 16+ is only required to run Storybook which is more of a dev dependency and upgrading to `engines.node` means a breaking change. Therefore I will rather upgrade `engines.node` on the `backend` branch which will be the next major release.

Issue #4274, opened 2023-03-28
[backend] Write nicer html emails


thesan:
## Context

Since we want the solution to be email distribution platform agnostic, we want the HTML templates for emails to be passed as HTML to the email platform used by BE service, rather than configuring them on each platform individually. 

## Layout Design

[LInk to Figma>](https://github.com/Joystream/pioneer/issues/4252)  

## Scope
- [ ] Create HTM version of the template for the different notifications, using layout designs attached. 

1️⃣ Verify email

`Dear {Name},
Follow this link {ilnk back to the app} to verify your email for receiving transactional notifications from the Pioneer App.`

CTA = Verify email. 

Points back to Pioneer.

2️⃣  2.1 Council election stage updates: Election stage

`Hi {Name},
Election stage has now started. Follow the link to create council election candidacy.`

CTA = Create Candidacy

2️⃣  2.2 Council election stage updates: Voting stage

`Hi {Name},
Council elections voting stage has now started. Follow the link to vote for the council candidates.`

CTA = View Candidates

2️⃣  2.3 Council election stage updates: Revealing stage

```
Hi {Name},
Council elections revealing stage has now started. In case you voted, follow the link to reveal your votes.
```

CTA = Reveal votes

3️⃣  Forum/Direct mentions 

```
Hi {Name},
You've been mentioned in Forum thread: {Forum_thread_Name} by {handle_of_member_who_mentioned}. 

{Preview of message}
```

CTA = Reply on Forum

Points to the exact post where user was mentioned

4️⃣ Forum/Threads/EveryNewThreadCreated

`Hi {Name},
New forum thread was created in {Name_of_forum_category} Category.`

CTA = View on Forum

Points to thread

5️⃣  Forum/Threads/AllPostsInThreadIcreated

`Hi {Name},
New post was added to the {Name_of_thread} thread you created`

CTA = View on Forum

Points to thread

6️⃣ Forum/Threads/AllPostsAndRepliesInThreadsIposted

`Hi {Name},
New post was added to the thread where you posted: {Name_of_thread}.`

CTA = View on Forum

Points to post

## Technical suggestions

For emails, styles should be inline.

In my experience writing emails which look good regardless of the email client can be painful. So if it's not overkill [mjml](https://github.com/mjmlio/mjml) looks like a nice solution.

ignazio-bovo:
After a call with @thesan I tried locally a combination of handlebars and mjml
Steps to reproduce:
``` sh
npm i mjml
[...]

PR #4410, opened 2023-06-05
 Fix/4274 write nicer html emails #4274

RAISEMEUP007:


packages/ui/src/app/pages/Proposals/EmailNotification.tsx
```diff
@@ -0,0 +1,382 @@
...
+        <ModalOverlay>
+          <ModalContainer>
+            <ModalHeader>
+              <Link>
+                <img src={NotificationLogo} alt="Mark" />
+              </Link>
+            </ModalHeader>
+            <ModalContent>
+              <ContentTitle>Hi {'{name}'},</ContentTitle>
+              <ContentBody>New proposal {'{proposal name}'} is created... there are 2 days left to vote...</ContentBody>
```
chrlschwb, line 34: According to the issue https://github.com/Joystream/pioneer/issues/4274
you must create six variations of notifications. Please create the text of the contentbody and button as variables.

chrlschwb (changes requested):

chrlschwb (commented):
You must define "EmailNotifications.tsx" inside "src/common"

thesan (changes requested):
@AnttoLion I don't know if this PR is still in progress, but in case it is:

Using React to generate html emails looks like a bad idea. Remember that the emails are generated on the backend so there shouldn't be any changes in `packages/ui/*`. Also you can't run js in emails and even though React can be used for ssr it's really not the right tool when the frontend can't run React.

kdembler:
@thesan This is superseded by my PR, can be closed

thesan:
Done in #4511

Issue #4403, opened 2023-06-01
Change app title based on current thread / proposal


traumschule:
Rationale

Working with Pioneer requires to have lots of tabs with Pioneer instances. Modern browsers have an option to show the list of open tabs. Currently Pioneer just shows `Joystream Governance App` which does not help to select the right tab.

New (examples):
- `Working Groups - Joystream Governance App`
- `Proposal "[Title]"  - Joystream Governance App`
- `Thread "Development Priorities Q2" - Joystream Governance App`

Tasks:
1. Each route needs a new title property (defined per page routes).
2. Title can contain variables like thread, proposal or opening title.

dmtrjsg:
Like the idea of naming the tabs in accordance to the page the person is viewing, but suggest to drop `Joystream Governance App` from title and rely on the icon instead.

traumschule:
We can keep the static title in index.html and replace it with `document.title` or [Helmet](https://github.com/nfl/react-helmet).

PR #4424, opened 2023-06-11
Drop app title (#4403)

traumschule:
#4403

### DIFF
--- a/packages/ui/src/index.html
+++ b/packages/ui/src/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/svg+xml" sizes="any" href="/favicon.svg" />
-    <title>Joystream Governance App</title>
+    <title></title>
   </head>
 
   <body>


packages/ui/src/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/svg+xml" sizes="any" href="/favicon.svg" />
-    <title>Joystream Governance App</title>
+    <title></title>
```
chrlschwb, line 7: This will only remove the current title. But the issue requirement is to make the title change according to the page the viewer is in.

chrlschwb (changes requested):

packages/ui/src/index.html
traumschule: True it's in response to https://github.com/Joystream/pioneer/issues/4403#issuecomment-1584362104 although probably not the best solution. Possibly misunderstood also:
We can keep the static title in index.html and replace it later on with `document.title` or [Helmet](https://github.com/nfl/react-helmet). wdyt @dmtrjsg

PR #4428, opened 2023-06-12
4330 create validator profile

### LINKED ISSUE #4330: Create validator profile
## Context

- https://github.com/Joystream/pioneer/issues/4321

## Design

🆕 [Link to Figma> ](https://www.figma.com/file/QnjZuWYtGimTEkQlN6MNlH/Validators?type=design&node-id=1701%3A216629&mode=design&t=nWRhVkKaUSy1Vu7k-1)

Design Preview

Screenshot 2023-10-26 at 16 23 29

## Scope
Add section to create membership modal for validators. 

- [ ] Toggle: I am validator. By default set to "no". When set to Yes adds more fields to the form
- [ ] Stash account(s) or Validator Controller account(s) binding
 ⚠️ These fields will go to membership metaprocol and we need to add these fields  to QN
- [ ]Add  tooltips
- [ ] Status will be "Unverified by default"

⚠️ When validator account added, the membership creation tx has additional step - binding account to membership. This happens before Membership is created/ updated (as a first step in the tx)

Screenshot 2023-10-26 at 16 23 36

⚠️ Pls make sure that all form inputs validation are done in accordance to the full scnearios of user story 

- Binding multiple accs at once
- Insufficient balance
- Wrong acc address pasted
- Signing the binding tx with correct account (stash or validator controller)

ℹ️  Verification is covered in this story:
- https://github.com/Joystream/pioneer/issues/4334

eshark9312:
- #4330 

Add section to create membership modal for validator's membership.
Add additional step to create membership transaction modal for binding validator's account to the membership.

[> Design](https://www.figma.com/file/QnjZuWYtGimTEkQlN6MNlH/Validators?type=design&node-id=1717-236794&mode=design&t=z1LVePeYx80EU2R5-0)
[> Story](https://pioneer-2-storybook-git-fork-eshark9312-4330-c-db4933-joystream.vercel.app/?path=/story/app--default)


chrlschwb (approved):

thesan (changes requested):
Please rebase https://github.com/Joystream/pioneer/pull/4428 on the latest `dev` so you can add a test case for the validator profile creation in the [App stories](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/app/App.stories.tsx)

thesan (changes requested):
@eshark9312 please remove the last commit ([fix storybook](https://github.com/Joystream/pioneer/pull/4428/commits/489d7f6b662c53de4cd043f29a906bf20738d129)). All you need to do to fix the interaction tests is `git pull upstream dev` (with upstream being `Joystream/pioneer`).

thesan:
@eshark9312 sorry was wrong in the previous comment:

1. Actually you are correct on this change `expect(await modal.findByText('Failure'))` on `packages/ui/src/app/App.stories.tsx`
But please remove the change on `packages/ui/src/app/pages/Proposals/CurrentProposals.stories.tsx`.
2. Pulling `dev` should solve the interaction tests. But also has I wrote [here](https://github.com/Joystream/pioneer/pull/4428#pullrequestreview-1548101212): please add some tests to `packages/ui/src/app/App.stories.tsx`.

eshark9312:
> @eshark9312 sorry was wrong in the previous comment:
> 
> 1. Actually you are correct on this change `expect(await modal.findByText('Failure'))` on `packages/ui/src/app/App.stories.tsx`
>    But please remove the change on `packages/ui/src/app/pages/Proposals/CurrentProposals.stories.tsx`.
> 2. Pulling `dev` should solve the interaction tests. But also has I wrote [here](https://github.com/Joystream/pioneer/pull/4428#pullrequestreview-1548101212): please add some tests to `packages/ui/src/app/App.stories.tsx`.

eshark9312:
@thesan
I removed the change on `packages/ui/src/app/pages/Proposals/CurrentProposals.stories.tsx` and added some tests.

packages/ui/src/memberships/modals/BuyMembershipModal/machine.ts
```diff
@@ -32,21 +36,31 @@ export type BuyMembershipEvent =
...
   initial: 'prepare',
   states: {
     prepare: {
       on: {
         DONE: {
-          target: 'transaction',
+          target: 'buyMembershipTx',
+          actions: assign({ form: (_, event) => event.form }),
+        },
+        DONEWITHVAL: {
```
thesan, line 57: Please rename this:
```suggestion
        DONE_IS_VALIDATOR {
```

packages/ui/src/memberships/modals/UpdateMembershipModal/UpdateMembershipFormModal.tsx
```diff
@@ -81,7 +94,15 @@ export const UpdateMembershipFormModal = ({ onClose, onSubmit, member }: Props)
     mode: 'onChange',
   })
 
-  const [controllerAccount, rootAccount, handle] = form.watch(['controllerAccount', 'rootAccount', 'handle'])
+  const [controllerAccount, rootAccount, handle, isValidator, validatorAccountCandidate] = form.watch([
```
thesan, line 97: I think the changes to this file (and the `types.ts` file) should be reverted and added in a followup PR because this flow might be more complex. E.g the validator accounts might be already partially bound to the existing membership so it's difficult to predict how many transactions will be needed. Also it will have to be tested so a new set of stories (I think "My Membership") have to be added.

WDYT ?

thesan (changes requested):

packages/ui/src/app/App.stories.tsx
```diff
@@ -606,10 +612,91 @@ export const BuyMembershipWithValidatorAccountHappyAndBindHappy: Story = {
...
+        handle: NEW_MEMBER_DATA.handle,
+        metadata: metadataToBytes(MembershipMetadata, {
+          name: NEW_MEMBER_DATA.metadata.name,
+          about: NEW_MEMBER_DATA.metadata.about,
+          avatarUri: NEW_MEMBER_DATA.metadata.avatar.avatarUri,
+        }),
+        invitingMemberId: undefined,
+        referrerId: undefined,
+      })
+      expect(args.onAddStakingAccount).toHaveBeenCalledTimes(4) // means 2 times, due to the React hook 'useMomo'
```
thesan, line 698: I don't think that's correct. `onSend` is only called when the mock transaction `signAndSend` method is called. So this should only happen once per transaction (so 2 times) regardless of the `useMemo`.

Am I missing something ? If not could you look into why this is called 4 times right now please ?

packages/ui/src/app/App.stories.tsx
```diff
@@ -606,10 +612,91 @@ export const BuyMembershipWithValidatorAccountHappyAndBindHappy: Story = {
...
+      expect(args.batchTx).toHaveBeenCalledTimes(2)
```
thesan, line 699: Same here AFAICS this should be called only once

thesan (changes requested):

packages/ui/src/app/App.stories.tsx
eshark9312: Actually this extrinsic has called two times for each sign, first time is to get the fee of the tx, and next time for signature.

https://github.com/Joystream/pioneer/blob/5d255aed7ea1f8ffd2db41074995c43cef59ff37/packages/ui/src/common/hooks/useProcessTransaction.ts#L108
https://github.com/Joystream/pioneer/blob/5d255aed7ea1f8ffd2db41074995c43cef59ff37/packages/ui/src/common/hooks/useProcessTransaction.ts#L127

thesan: Here's the mocked transaction code:

https://github.com/Joystream/pioneer/blob/4b9a450e210c0e14d005eca5fb7c75c94ea26594/packages/ui/src/mocks/helpers/transactions.ts#L30-L40

As you can see the `onSend` is only called in `signAndSend` not in `paymentInfo`. There must be a different reason why these transactions get sent multiple times.

thesan:
@eshark9312 I figured it out. It's related to this:
image
I completely missed in my earlier reviews :facepalm:

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -152,6 +160,18 @@ export const BuyMembershipForm = ({
   ])
 
   const [validatorAccounts, setValidatorAccounts] = useState<Account[]>([])
+  const isValidValidatorAccount = useMemo(
```
thesan, line 163: To use the `filter` it might be nice to expose the list of addresses:

```suggestion
  const validatorAddresses = useMemo(() => {
    if (!validatorAccountCandidate || !allValidatorsWithCtrlAcc || !allValidators) return

    return [...allValidatorsWithCtrlAcc, ...allValidators.map(({ address }) => address]
      .filter((element) => !!element) // Is this filter needed ?
      .map(encodeAddress)
  }, [allValidators, allValidatorsWithCtrlAcc, validatorAccountCandidate])

  const isValidValidatorAccount = useMemo(
```

thesan (changes requested):

thesan (approved):
Nice work once again :raised_hands: 

I had forgotten to suggest skipping the validator query for until someone claims to be a validator on my last review so I added it in a small commit (especially since I'm not so available rn so going through another review cycle would waste time).

Also I'm merging this on a new `validators` branch so feature can still be released while the validator dashboard is getting tested.

So from the top of my head what's left is:
- [ ] #4706
- [ ] ~~Adding the dashboard to the app routes and the navigation~~ Done in #4706 !
- [ ] The update membership modal

Issue #4275, opened 2023-03-28
[backend] Allow retrying sending email when it fails and replace `isSent` by `emailStatus` enum which accounts for failure


thesan:
## Context

Currently the notification entity has only 2 statuses regarding the email sending: `isSent: true` or `isSent: false`. When an email fails to be sent it `isSent` is set to `true` to avoid infinite retry.

## Solution

- [ ] There should be a third state something like `FAILED_TO_SEND`. So the `isSent` boolean should be renamed and changed into an enum.
- [ ] There should also be a counter to only retry sending the email a few times. These attempts shouldn't happen within the same job call. E.g if an email failed to be send during the cron job of 9:00 it will not be retried until the next cron job at maybe 9:30.
- [ ] This behavior should be included in the integration tests (to be created).

PR #4432, opened 2023-06-14
Fix/4275 allowing retrying sending email

RAISEMEUP007:
- #4275


thesan (changes requested):
@AnttoLion the logic looks correct but the [prisma schema](https://github.com/Joystream/pioneer/blob/backend/packages/server/prisma/schema.prisma#L47) needs to be updated (and a new migration is also needed) for this to work.

Also please add a test in [the 'Notifier' test suite](https://github.com/Joystream/pioneer/blob/backend/packages/server/test/notifier.test.ts#L9) for this.

PR #4439, opened 2023-06-15
Error Boundary

### LINKED ISSUE #4052: Add error boundaries

traumschule:
rebase #4195 , fixes #4052

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -67,6 +67,7 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-dropzone": "^11.4.2",
+    "react-error-boundary": "^3.1.4",
     "react-hook-form": "^7.29.0",
     "react-i18next": "^11.15.1",
     "react-intersection-observer": "^8.32.1",
--- a/packages/ui/src/app/App.tsx
+++ b/packages/ui/src/app/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { ErrorBoundary } from 'react-error-boundary'
 import { Redirect, Route, Switch } from 'react-router-dom'
 
 import '@/services/i18n'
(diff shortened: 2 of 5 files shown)


chrlschwb (approved):
thanks, look good

chrlschwb:
but why are the CI tests not running?

thesan (changes requested):
@traumschule the code looks good but please add a test case in [`@/app/App.stories.tsx`](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/app/App.stories.tsx).

I think we should avoid adding error boundary in any other stories for the sake of smoke tests. So the story defining the test should redefine the render method with the error boundary and a component which will throw something. I'm thinking something like this:
```tsx
const MockError: FC<{ error: () => void }> = ({ error }) => {
  useEffect(error, [])
  return <div>Error component</div>
}

export const ErrorBoundary: Story = {
  render: () => {
    const renderCount = useRef(0)
    return (
      <Page>
        <NoPaddingStyle />
        <SideBar />
        <ErrorBoundary FallbackComponent={ErrorFallback}>
          <Screen>
            <OnBoardingOverlay />
            <MockError error={() => ++renderCount.current === 1 && (undefined as any)()} />
          </Screen>
        </ErrorBoundary>
      </Page>
    )
  },

  play: () => {
    /* Test what's rendered */
  },
}
```

I wish it could be tested on modals too because I'm a bit worried about how it interacts with `ReactDOM.createPortal` but that would require a bit too much work IMO. This should be good enough for now, if some issues get reported later on regarding modals, further tests could be added then.

PR #4446, opened 2023-06-19
🧪 POC Move the `ProposalPreview` tests to Storybook

thesan:
Part of:
- #3559


traumschule (approved):
good idea, `RecursivePartial` code is hard to review though.

thesan:
> good idea, `RecursivePartial` code is hard to review though.

Yeah these recursive types can be painful to write. TBH I just got this one online. But it's mostly here to make it a bit easier to write QN mocks (without needing to constantly jump through schema) by providing type based auto completion. Sometime it will spot errors but the `as` keyword is not really the right tool for that so it shouldn't be seen as a way to make the mocks type safe.

chrlschwb (approved):

PR #4452, opened 2023-06-20
🔧 Fix the storybook CI

thesan:
This PR fixes the new CI for PRs made from forks, which basically will be all PRs made to Pioneer very soon.

E.g this is why the CI failed on #4453.

The previous version required secrets to use the Vercel CLI, but secrets are not available on workflows triggered from a fork. So now the CI generates the Storybook deployment URL following [the Vercel doc](https://vercel.com/docs/concepts/deployments/generated-urls#url-with-git-branch) and relies on one of the few public endpoints from the Vercel REST API to wait for the deployment to complete.

I couldn't find a way to rely on the `deployment_status` trigger [like in the Storybbok doc example](https://storybook.js.org/docs/react/writing-tests/test-runner#run-against-deployed-storybooks-via-github-actions-deployment) because of the 2 other Vercel preview deployments which are almost always slower and because GH only looks at the latest run of a given job (even if the job was skipped) and ignore the previous ones.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -106,16 +106,16 @@ jobs:
       - name: Get the Storybook preview deployment url
         run: |
           if [ "$PREFIX" == 'git-fork' ]; then
-            BRANCH=$(echo -n "$LABEL" | sed 's/:/-/')
+            BRANCH=$(echo -n '${{ env.LABEL }}' | sed 's/:/-/')
           else
-            BRANCH=$(echo -n "$LABEL" | cut -d ':' -f2-)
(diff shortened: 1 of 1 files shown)


traumschule (approved):
needs testing

thesan:
> needs testing

Yeah I had to do [**a lot**](https://github.com/Joystream/pioneer/actions?query=branch%3Afix%2Fstorybook-ci-with-a-long-branch-name+branch%3Afix%2Fstorybook-ci) of testing for this:
Screenshot from 2023-06-21 21-14-41

Using this PR and #4454. I also tested the push trigger which get called after PRs are merged. I think it should be all good now :crossed_fingers: writing GH actions is the worst !

PR #4458, opened 2023-06-24
🚚 Move the vote for proposal modal tests to Storybook 

thesan:
Part of:
- #3559


packages/ui/src/app/GlobalModals.tsx
```diff
@@ -205,6 +205,12 @@ export const GlobalModals = () => {
   const { status } = useTransactionStatus()
   const Modal = useMemo(() => (modal && modal in modals ? memo(() => modals[modal as ModalNames]) : null), [modal])
 
+  const [container, setContainer] = useState(document.body)
+  useEffect(() => {
+    const container = document.getElementById('modal-container')
+    if (container) setContainer(container)
```
traumschule, line 211: It's clear that the `useEffect` context defines its own `container` to set and overwrite the outer one to the selected element in the document tree.

traumschule (approved):
Vercel didn't load via Tor but should be fine.

thesan:
@traumschule I can't load any domain ending in `.vercel.app` with Tor not even https://dao-git-main-joystream.vercel.app even though https://pioneerapp.xyz loads. I'm assuming Vercel blocked Tor connections on their free sub domains.

PR #4474, opened 2023-07-11
signature for email modal

### LINKED ISSUE #4232: Email notifications: signer modal component and transaction
## Context
- https://github.com/Joystream/pioneer/issues/4230

## Backend
⚠️ [Link to BE API implementation >](https://github.com/Joystream/pioneer/blob/backend/packages/server/README.md)

## Dependencies

Blocks: 
- #4231 
- #4237 

Blocked by:
- #4274 

## Scope
- [ ] Create component (Signer modal for connecting tx email with membership.) [Link to Design > ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11947%3A511977&t=kbnLvo8CmtA4rR1I-1)
-- email input
-- membership controller account displayed which is used to sign
-- Ctas: "sign and authorize email"; "Not now"
-- state when email is not provided -> CTA  "sign and authorise" is disabled unless email is provided
- [ ] When email is provided and modal signed, confirmation email must be sent.. 
- [ ] Confirmation email template for email auth. HTML template is required to be used  see
- #4274 point 1.
- [ ] Following the link user is brought back to the App with confirmation modal shown. [Link to design> ](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?node-id=11947%3A508551&t=kbnLvo8CmtA4rR1I-1)

## Design
Links are added to acceptance criteria
Preview: 
Screenshot 2023-03-30 at 18 08 42

IlyaSmiyukha:
closes #4232


thesan (changes requested):
@IlyaSmiyukha technically it works but it's not an acceptable solution because creating api instances is resource intensive, this is why Atlas and Pioneer ended up moving the api instance on web workers (the page would freeze for users with lower resource computer). In this case the api instance is garbage collected and it's created again if the modal is reopened. I made sure of this using my browser profiler and the main thread does always spike shortly after opening the modal :
Screenshot from 2023-07-11 18-18-06
(the 1st row shows app screenshots, I couldn't figure out how to make it bigger).

Instead what I'm suggesting is to:
1. *On the client:* Go to [`@/proxyApi/client/ProxyApi.ts`](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/proxyApi/client/ProxyApi.ts) and add a `sign(account: string, message: any)` method on the class. This method should call `postMessage` with a new message type and the parameters, and return `firstValueFrom(messages.pipe(...))` by also filtering from a new message type and mapping it to the signature value returned be the worker.
2. Then *on the web worker:* ([`@/proxyApi/worker/index.ts`](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/proxyApi/worker/index.ts)) the `api.sign` method can be called with the correct value and when it resolves the signature should be passed back to the client with `postMessage` (similarly to how it was done 1).

I haven't tried it but it seems the sign method is also called from the web worker on Atlas so it should work. You can check the other asynchronous messages passed to and from the web worker for reference and don't hesitate to ask me if you get stuck.

thesan (commented):
Nice work ! It's great the wallet instance was usable for this instead of having to mess with the web worker.

Now please use `send` to update the machine context with the `signature` and `timestamp`.

Also I just merged the new [`App`](https://pioneer-2-storybook-git-backend-joystream.vercel.app/?path=/story/app--default) tests from `dev` into `backend`. So in [`packages/ui/src/app/App.stories.tsx`](https://github.com/Joystream/pioneer/blob/backend/packages/ui/src/app/App.stories.tsx) could you please as part of this PR:

1. Define the member emails for the tests/stories. This can be done easily by adding this to the mocks:
   ```ts
   localStorage: {
     membersEmail: { alice: 'alice@example.com', bob: 'bob@example.com' }
   }
   ```
2. Start adding a test for this flow. For this you will need a story where the emails above are not defined. To do that I usually use storybook `args` so that instead of just defining: `membersEmail: { alice: 'alice@ex...` I would define something like that: `membersEmail: args.hasRegisteredEmail ? { alice: 'alice@ex... } : {}`. Now by creating a story with `{ args: { hasRegisteredEmail: false } }` the modal will show and the test flow can start.

packages/ui/src/memberships/modals/EmailSubscriptionModal/EmailSubscriptionModal.tsx
```diff
@@ -13,18 +13,29 @@ import { EmailSubscriptionMachine } from './machine'
...
+  const signEmail = async () => {
+    const timestamp = Date.now()
+    const signature = await wallet?.signer.signPayload({
+      address: member.controllerAccount,
+      data: `${member.id}:${timestamp}`,
+    })
+    if (signature) {
+      send('SIGNED', { signature: signature.signature, timestamp })
+    }
+    return signature
```
thesan, line 33: There is no need to return the signature. However if the `signature` is empty it's probably that the user cancelled the process, thus state the machine should be updated to the cancel state

chrlschwb (approved):

packages/ui/src/app/App.stories.tsx
```diff
@@ -408,3 +413,44 @@ export const BuyMembershipTxFailure: Story = {
...
+    isLoggedIn: true,
+    hasMemberships: true,
+    hasAccounts: true,
+    hasFunds: true,
+    hasWallet: true,
+    isRPCNodeConnected: true,
+    hasRegisteredEmail: false,
+  },
+  play: async ({ canvasElement }) => {
+    const screen = within(canvasElement)
```
thesan, line 431: For a slight performance improvement but also improving the test accuracy, whenever looking for an element in a modal restrict the search to the modal container:
```suggestion
    const modal = withinModal(canvasElement)
```

thesan (changes requested):

IlyaSmiyukha:
@thesan 
Should be fine now

packages/ui/src/app/App.stories.tsx
```diff
@@ -408,3 +413,65 @@ export const BuyMembershipTxFailure: Story = {
...
+    hasRegisteredEmail: false,
+  },
+  play: async ({ canvasElement }) => {
+    const modal = withinModal(canvasElement)
+    const button = modal.getByText(/^Sign and Authorize Email/i)
+    expect(button.closest('button')).toBeDisabled()
+    await userEvent.type(modal.getByPlaceholderText('Add email for notifications here'), 'test@email.com')
+    await waitFor(() => expect(button.closest('button')).toBeEnabled())
+    await userEvent.click(button)
+    expect(modal.getByText('Transaction was canceled.'))
```
thesan, line 475: At this point the machine gets in the cancel state because `wallet? signer.signPayload` is `undefined`, the code tries to call it, and all errors are assumed to be the user cancelling the signature. So this is not correct. Instead please change the accounts mock so you can pass it a [storybook action](https://storybook.js.org/docs/react/essentials/actions) called "sign" which will be added to the mock wallet. Currently [`WALLET` is a constant in the module scope](https://github.com/Joystream/pioneer/blob/backend/packages/ui/src/mocks/providers/accounts.tsx#L149) so this should be changed.

thesan (changes requested):

thesan (approved):
Nice job!

Issue #4319, opened 2023-04-03
Add Grace Period to Proposal Preview Page


dmtrjsg:
## Context

For time-sensitive proposals its important to show grace period for operations coordination. Currently grace period only shows to creator of proposal in the Create Proposal flow. 

## Scope
Add grace period to proposal details view
- [ ] Show grace period duration in time/ block
- [ ] Show Exact Block Number at which proposal executes

thesan:
QA: https://github.com/Joystream/pioneer/pull/4478#issuecomment-1742762682

@ivanturlakov @chrlschwb thank you for taking care of the QA. But whenever possible (like here were the PR has already been merged on the `dev` branch) please test the issues on [pioneer-2.vercel.app](https://pioneer-2.vercel.app). (But obviously no need to re-test this one).

PR #4478, opened 2023-07-16
Added Proposal Grace Period

vrrayz:
According to this [issue](https://github.com/Joystream/pioneer/issues/4319)

- [x] Show grace period duration in time/ block
- [x]  Show Exact Block Number at which proposal executes

Screenshot (629)

The second task **Show Exact Block Number at which proposal executes** 
I thought i could achieve that with `proposal.exactExecutionBlock`  but its always undefined for the proposals i've checked out so far


chrlschwb (approved):
ok

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
```diff
@@ -72,6 +74,24 @@ export const ProposalDetails = ({ proposalDetails }: Props) => {
...
+    }
+    if (exactExecutionBlock) {
+      return [
+        {
+          renderType: 'NumberOfBlocks',
+          label: 'Exact Execution Block',
+          value: exactExecutionBlock,
+        },
+      ] as RenderNode[]
+    }
```
thesan, line 94: Here `exactExecutionBlock` is not actually a number of blocks but a block height so `NumberOfBlocks` is not the right component. Instead it should show something like the [BlockTime](https://pioneer-2-storybook-joystream.vercel.app/?path=/story/common-blocktime--block-time-view) component.

thesan (changes requested):
That's the correct data but the current proposal statistics components are not good enough to show these information in a helpful way.

Also we should not show both gracing period and execution block at the same time. AFAIK when `exactExecutionBlock` is set the gracing period doesn't matter. So please show only the `exactExecutionBlock` when it's set and only the gracing period otherwise.

chrlschwb (changes requested):
please put the changes into a new type file

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
```diff
@@ -72,6 +75,24 @@ export const ProposalDetails = ({ proposalDetails }: Props) => {
   const detailsRenderStructure = useMemo(() => getDetailsRenderStructure(proposalDetails), [proposalDetails])
 
   const additionalDetails = useMemo(() => {
+    if (gracePeriod) {
```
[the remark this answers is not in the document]
vrrayz: Thanks for the heads up. I tried it and i think adding type `setGracePeriod` doesn't work because `gracePeriod` originally isn't a type of proposal just like `fundingRequest, cancelWorkingGroupLeadOpening...` and the other types.

In `pioneer/packages/ui/src/proposals/types/constants.ts` line 6. its already a type property of `ProposalConstants` so it is different from the `proposalDetails?.type`

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
```diff
@@ -72,6 +75,24 @@ export const ProposalDetails = ({ proposalDetails }: Props) => {
...
+      return [
+        {
+          renderType: 'BlockDurationStatistics',
+          title: 'Gracing Period',
+          value: gracePeriod,
+        },
+      ] as unknown as RenderNode[]
+    }
+    if (exactExecutionBlock) {
```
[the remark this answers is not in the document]
vrrayz: Just like my comment earlier i think the same goes for the 'exactExecutionBlock', in the `packages/ui/src/app/pages/Proposals/ProposalPreview.tsx` file line 186. its a property of `Proposal` in the `packages/ui/src/proposals/types/proposals.ts` file. Thats why i had to add them as properties seperately in the `ProposalDetails` component

chrlschwb (approved):

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
```diff
@@ -51,9 +53,10 @@ const renderTypeMapper: Partial<Record<RenderType, ProposalDetailContent>> = {
   Percentage: Percentage,
   Hash: Hash,
   DestinationsPreview: DestinationsPreview,
+  BlockDurationStatistics: BlockDurationStatistics,
```
thesan, line 56: It shows the right thing but I think this should be the only way we show a number of blocks. So instead of adding:
```ts
BlockDurationStatistics: BlockDurationStatistics,
```
Please replace `NumberOfBlocks` by:
```ts
NumberOfBlocks: BlockDurationStatistics,
```
and remove `NumberOfBlocks.ts`

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
```diff
@@ -72,6 +75,24 @@ export const ProposalDetails = ({ proposalDetails }: Props) => {
   const detailsRenderStructure = useMemo(() => getDetailsRenderStructure(proposalDetails), [proposalDetails])
 
   const additionalDetails = useMemo(() => {
+    if (gracePeriod) {
+      return [
```
thesan, line 79: I missed that on the first review but if this logic is in this `useMemo`: `setReferralCut`, `fundingRequest`, and `updateWorkingGroupBudget` will be missing there "Amount" statistics. E.g [`setReferralCut`](https://pioneer-2-storybook-git-fork-vrrayz-issue-4319-joystream.vercel.app/?path=/story/pages-proposals-proposalpreview--set-referral-cut).

Instead either "Gracing Period" or "Exact Execution Block" should be append to the statistics (just like `additionalDetails` already is).

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
```diff
@@ -72,6 +75,24 @@ export const ProposalDetails = ({ proposalDetails }: Props) => {
...
+          renderType: 'BlockDurationStatistics',
+          title: 'Gracing Period',
+          value: gracePeriod,
+        },
+      ] as unknown as RenderNode[]
+    }
+    if (exactExecutionBlock) {
+      return [
+        {
+          renderType: 'BlockDurationStatistics',
```
thesan, line 90: Finally this https://github.com/Joystream/pioneer/pull/4478#discussion_r1327571671 still stands. `exactExecutionBlock` is a date not a duration so it shouldn't be shown has a `BlockDurationStatistics` but has something which looks like [BlockTime](https://pioneer-2-storybook-joystream.vercel.app/?path=/story/common-blocktime--block-time-view).

thesan (changes requested):

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
thesan: BTW I just added `exactExecutionBlock` in the stories so you can test it there. Just set it to a non zero value in the controls.

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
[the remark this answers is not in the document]
chrlschwb: image
change in packages/ui/src/proposals/types/proposals.ts

and add follow code in /home/create/Work/pioneer/packages/ui/src/proposals/types/ProposalDetails.ts

export type GracePeriodDetail = {
  value: string
}

export type GracePeriodDetails = ProposalDetailsNew<'GracePeriodDetails', GracePeriodDetail>

image

I don't think this is exactly right, but I think you can do it like this.

chrlschwb (approved):

thesan (changes requested):

packages/ui/src/proposals/components/ProposalDetails/ProposalDetails.tsx
```diff
@@ -32,6 +35,7 @@ interface Props {
   proposalDetails?: ProposalWithDetails['details']
   gracePeriod?: number
   exactExecutionBlock?: number
+  createdInBlock: Block
   createdAt?: string
   updates?: ProposalStatusUpdates[]
```
thesan, line 40: Please remove these 2:
```suggestion
```

thesan (changes requested):
Last things:

thesan (approved):
LGTM!

ivanturlakov:
Tested on https://dao-21pwiyi8j-joystream.vercel.app/#/proposals/current
Mainnet

✅ Show grace period duration in time/ block(ex block not specified)
Screenshot 2023-10-02 at 13 13 51

✅ Show Exact Block Number at which proposal executes(ex block 4 307 120)
Screenshot 2023-10-02 at 13 04 13

Issue #4484, opened 2023-07-31
Too many queries on the forum category page 


thesan:
On category pages with large amounts of threads the amount of QN request quickly reaches the hundreds and keeps on rising. The QN is pretty responsive at the moment but after opening several category pages it starts to throttle and navigating the forum becomes painful.

IMO there shouldn't be that many queries it looks like a bug.

![Screenshot from 2023-07-31 09-51-32](https://github.com/Joystream/pioneer/assets/6571453/74439685-2acb-47af-be0a-2fde2c1985b0)

thesan:
Suggestion:

The "last post" sections might be the most expensive part and "last activity" the close second. IMO instead of sending a lot of queries we could:
1. Sending a big query to get e.g the 500 latest post as well as their tread and direct category id.
2. Caching this query for a long time (maybe even a day).
3. The first occurrence of each subcategories is looked for in the 500 result then either the resulting post is shown or the section is left empty.
4. Do the same for threads for last activity.

Also showing 10 thread instead of 30 would help too.

Alternatively addressing https://github.com/Joystream/joystream/issues/2599 would solve this long term but it would take more work.

mkbeefcake:
@thesan 
I have checked this issue 4484, 
By checking the source code, there were period call for [GetForumThreads, GetForumThreadsCount] graph query per 6 seconds. 
For example, in this attached screenshot, It contains 4 sub categories and 1 main category, so there happened almost (4 + 1) * 3 requests per each 6 seconds by apolloclient.refetchQueries to update the cache from the QN. 
Also per each thread, it calls ForumPost query to get the author of thread from QN, these increases the requests.
So these many requests happened.

thesan:
@mkbeefcake thank you for checking it now makes sense why the number of queries keeps on growing.

However that's part of the bug:
1. Refetching `GetForumThreads` is just meant to keep the current category thread list up to date. But currently it also refetches all the `GetForumThreads` used to get the subcategory latest activity and the popular thread. The solution I proposed above would solve part of this problem too by not sending all of these `GetForumThreads` to begin with.
2. For the `GetForumThreadsCount` IMO it's also not worth keeping the subcategory thread count constantly up to date. Instead the only the current category's `GetForumThreadsCount` should be refetched every 6 seconds.

[...]

PR #4508, opened 2023-08-30
Fix: issue 4485, check 4484

### LINKED ISSUE #4485: Forum: Category pages refreshing multiple times before loading
# Problem
Example: https://pioneerapp.xyz/#/forum/category/8

Page will load and refresh multiple times. This happens on different browsers and with or without wallet extension present.

mkbeefcake:
Fixed #4485 and  checked #4484

### DIFF
--- a/packages/ui/src/app/pages/Forum/ForumCategory.tsx
+++ b/packages/ui/src/app/pages/Forum/ForumCategory.tsx
@@ -13,9 +13,7 @@ import { PageTitle } from '@/common/components/page/PageTitle'
 import { PreviousPage } from '@/common/components/page/PreviousPage'
 import { Label, TextMedium } from '@/common/components/typography'
 import { useModal } from '@/common/hooks/useModal'
-import { useRefetchQueries } from '@/common/hooks/useRefetchQueries'
 import { useSort } from '@/common/hooks/useSort'
-import { MILLISECONDS_PER_BLOCK } from '@/common/model/formatters'
 import { ForumCategoryList } from '@/forum/components/category/ForumCategoryList'
 import { ForumPageHeader } from '@/forum/components/ForumPageHeader'
 import { ThreadFilters } from '@/forum/components/threads/ThreadFilters'
@@ -48,16 +46,8 @@ export const ForumCategory = () => {
     },
     { perPage: THREADS_PER_PAGE, page }
   )
-  const isRefetched = useRefetchQueries({
-    interval: MILLISECONDS_PER_BLOCK,
-    include: ['GetForumThreads', 'GetForumThreadsCount'],
-  })
 
   const { showModal } = useModal()
-
-  if (isLoadingThreads && !isRefetched) {
-    return <Loading />
-  }
 
   if (isLoadingCategory) {
     return <Loading />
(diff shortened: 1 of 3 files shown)


packages/ui/src/app/pages/Forum/ForumCategory.tsx
```diff
@@ -48,17 +48,13 @@ export const ForumCategory = () => {
...
     interval: MILLISECONDS_PER_BLOCK,
     include: ['GetForumThreads', 'GetForumThreadsCount'],
   })
 
   const { showModal } = useModal()
 
-  if (isLoadingThreads && !isRefetched) {
-    return <Loading />
-  }
-
```
thesan, line 61: Great work on catching this ! It wasn't an easy one to debug :exploding_head: 

But I think the fix could be improved because first it now makes empty categories show a loader every 6 seconds e.g: https://dao-git-fork-mkbeefcake-fix-4485-4484-joystream.vercel.app/#/forum/category/17?network-config=https://34.230.5.182.nip.io/network/config.json. But more importantly it still refetch way to much data (as you pointed out on #4484). Also I realized we've struggled with this particular `useRefetch` for a long time (see #4318) so I really think the best is to remove this one.

Instead we could replace this logic by a `pollInterval` of 6 second in `packages/ui/src/forum/hooks/useForumCategoryThreads.ts` but only when `isArchive` is `false`. Also `isLoading` should become `!threadsData || !countData` this way `isLoading` is only `true` on the initial load.

thesan (changes requested):

packages/ui/src/app/pages/Forum/ForumCategory.tsx
mkbeefcake: image
Thanks for your guide. also I figured out "posts" field in forumThreads query, If i choose this mechanism, i can reduce the query count to 1, but it could take all posts info per each thread, so could take much data and a bit more loading time.
I hope to get your opinion about this.

thesan: Yes `forumThreads { posts { ... } }` will return all the posts for all the threads shown one the category page so I think 1 request per thread is a better trade off.

mkbeefcake: @thesan I have added posts field for ForumThread interface and fetched it in GetForumThreads Query, so reduce the queries in Forum Category page, please check the updated code.

packages/ui/src/forum/queries/forum.graphql
```diff
@@ -122,6 +122,26 @@ fragment ForumThreadFields on ForumThread {
...
+      __typename
+    }
+    threadId
+    thread {
+      categoryId
+    }
+    author {
+      ...MemberFields
+    }
+  }
```
thesan, line 144: I don't think we should go with this solution:
ATM on the Joystream General category it ~6x the size of the thread requests from `137kb` to `851kb` but the 30 posts request saved by this solution were between `4kb` to `10kb` so already the page queries a lot more data to start with. And then it repeats the thread request every 6 seconds. On top of that the thread request go all the way to several megabytes on some pages.

A bigger request can often be better than a lot of smaller one but here I really don't think this is worth it.

thesan (changes requested):

packages/ui/src/forum/queries/forum.graphql
mkbeefcake: I have removed "posts" from thread and updated the code. please review and give me feedback

Issue #4252, opened 2023-03-20
Email Templates Design for Pioneer


dmtrjsg:
We need a very simple html template for the transactional emails

Header
Footer with links back to pioneer app and joystream.org

[Link to Pioneer Logo> ](https://www.figma.com/file/3hnTi8QWyHDOtkZENhjxjB/Pioneer-brand?node-id=22-9469&t=xxwhdMgYZ4BV1ZPc-0)

KubaMikolajczyk:
The design for templates for emails are done ✅
https://www.figma.com/file/xpddKGxayX5O2orVAeUJUR/Pioneer%3A-Email-Templates?node-id=1%3A2&t=Yjb46paFCEFZWj81-1

PR #4511, opened 2023-08-31
[backend] add new email template

kdembler:
Added new notification email template according to the design done here https://github.com/Joystream/pioneer/issues/4252

For emails, I've used https://react.email, it allows writing email templates in React, so we use the same UI framework as the main app. The React template gets rendered to raw HTML when the email is sent.

You can do live preview of emails using `yarn workspace server dev:emails`. Here's an example: 

<details>
<summary>Details</summary>

CleanShot 2023-08-31 at 16 50 15@2x

</details>

I've also did some small stylistic code improvements to make it easier to read. When I first started looking at `server` package I was a bit confused by all the HOFs and passing functions around.

Closes https://github.com/Joystream/pioneer/issues/4274


kdembler:
@thesan any idea why the backend integration tests are failing?

kdembler:
nvm, found the issue!

packages/server/src/notifier/model/event/forum.ts
```diff
@@ -16,6 +16,9 @@ export const fromPostAddedEvent: NotifEventFromQNEvent<'PostAddedEvent'> = async
 
   const eventData = pick(postAddedEvent, 'inBlock', 'id')
 
+  // TODO: add FORUM_POST_REPLY support
+  // TODO: somehow restrict number of notifs per event to avoid spamming
```
thesan, line 20: This is done [here](https://github.com/Joystream/pioneer/blob/backend/packages/server/src/notifier/model/notifications.ts#L67) but it's at the end of the whole aggregation process in order to prioritize based on the user preferences.

packages/server/src/notifier/model/email/utils/index.ts
```diff
@@ -1,16 +1,62 @@
...
+
+    cachedForumThreads[id] = {
+      author: thread.author.handle,
+      title: thread.title,
+      text: thread.initialPost?.text,
+    }
+  }
+
+  return cachedForumThreads[id]
+}
```
thesan, line 62: Since there will be a lot of these queries based on entity ids I think it would be nice to move these two functions to a new `./forum.ts` file (in this folder) exported by this `./index.ts`

packages/server/src/common/email-templates/pioneer-email.tsx
```diff
@@ -0,0 +1,195 @@
...
+export const PioneerEmailTemplate = ({
+  memberHandle = 'bob',
+  text = 'New council election has just started. Follow the link below to announce your candidacy.',
+  summary,
+  // button default value only for email development preview
+  button = {
+    href: `${PIONEER_URL}/#/election`,
+    label: 'See on Pioneer',
+  },
+}: PioneerEmailTemplateProps) => (
```
thesan, line 35: That's a small detail but why not:
```suggestion
}: PioneerEmailTemplateProps) => render(
```
?

thesan (changes requested):
Great work! I didn't know `react-email` nor `ts-pattern` these are really good :star_struck: 

I added a few CRs but more importantly the jest config should be fixed to parse the jsx. Also you will need to mock `createEmailProvider` in `packages/server/test/setup.ts` and replace `text` by `html` in the notifier test.

I'm not sure why the interaction tests fail... Might be a Vercel issue or more likely I forgot to merge some CI fixes on this branch. If it happens again I'll take a better look at it.

packages/server/src/notifier/model/event/forum.ts
kdembler: thanks, missed that

packages/server/src/notifier/model/email/utils/index.ts
kdembler: moved 👍

packages/server/src/common/email-templates/pioneer-email.tsx
kdembler: good call, added exported `renderPioneerEmail` function

kdembler:
@thesan Fixed tests, but interaction tests are still failing

thesan (approved):
Perfect :raised_hands: 

I'll merge it right after #4521

packages/server/src/common/email-templates/pioneer-email.tsx
```diff
@@ -0,0 +1,196 @@
...
+  <Html>
+    <Head />
+    <Preview>{summary}</Preview>
+    <Body style={bodyStyle}>
+      <Container style={containerStyle}>
+        <Section style={logoSectionStyle}>
+          <Img
+            src="https://github.com/Joystream/design/blob/master/logo/pioneer/horizontal/monochromatic/light/cropped/png/pioneer-logo@1x.png?raw=true"
```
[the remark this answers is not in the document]
kdembler: I like the idea but apparently the support for SVGs in emails is low, especially for embedded ones:
https://www.caniemail.com/features/image-svg/
https://www.caniemail.com/features/html-svg/
It seems it's the same case for base64 encoded PNGs:

> Some bad news about base64 encoded images:
> 
> They are totally blocked by Outlook.
> They are not displayed by most webmail services (especially if you use more than one inside the message).
>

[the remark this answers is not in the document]
thesan: Thanks for having a look. And I am bookmarking caniemail !

PR #4512, opened 2023-09-01
🚀 Release `1.6.0`

thesan:
### Added
- Multiple recipient funding request proposal creation.

### Fixed
- Forum category page reloading multiple time on opening.
- Broken links to members on forum posts.
- Reduce the amount of queries from the forum category page.

### Changed
- Display amount staked by voting for a council candidate instead of their candidacy stake when applicable.


traumschule:
Sorry for the delay.
- I see no problem in this case but it shouldn't become a habit to release without review, even though individual issues underwent proper QA.
- Issues in release notes should be mentioned with issue number.
- This is the `staging` branch although description doesn't mention validators.
- Might be better not to hard-code validator addresses inline, even in stories.

PR #4522, opened 2023-09-04
[backend] add email send retries

### LINKED ISSUE #4275: [backend] Allow retrying sending email when it fails and replace `isSent` by `emailStatus` enum which accounts for failure

kdembler:
~Based on #4518, will rebase after it's merged~

Adds retries to emails that failed to be sent.

Resolves #4275


packages/server/test/notifier.test.ts
```diff
@@ -459,14 +468,171 @@ describe('Notifier', () => {
...
+            }),
+          ],
+        })
+        .mockReturnValue({
+          events: [],
+          forumPostByUniqueInput: {
+            author: { handle: 'author:handle' },
+            thread: { id: 'thread:id', title: 'thread:title' },
+            text: 'Lorem Ipsum',
+          },
```
thesan, line 582: Is this a mistake ?
```suggestion
```

packages/server/src/notifier/index.ts
```diff
@@ -1,7 +1,7 @@
 import { createNotifications } from './createNotifications'
-import { sendNotifications } from './sendNotifications'
+import { processNotifications } from './sendNotifications'
```
thesan, line 2: Just for consistency could you either rename the function `sendNotifications` or the file `processNotifications` ?

thesan (commented):
Looks great, just 2 things:

packages/server/test/notifier.test.ts
kdembler: Not sure what you mean, but that data is needed to create the notification for `postAddedEvent`

packages/server/src/notifier/index.ts
kdembler: Done 👍

packages/server/test/notifier.test.ts
thesan: Right! It's been a while... I forgot this mocks the result of all the gql requests

PR #3868, opened 2022-11-21
Fixed restore_votes not update MyVotes #2991, #2986

mkbeefcake:
Added customEvent mechanism inside the pioneer and refresh the Election page after Restore Votes clicked.
Fixed #2991,  #2986

Issue #2599, opened 2022-03-21
Update Working Group Descriptions


dmtrjsg:
All roles

https://joystream.gitbook.io/testnet-workspace/system/working-groups

## Scope

Add descriptions to working groups cards: 

- [x] Storage

Broadly responsible for ensuring storage infrastructure uptime, namely running complete and up-to-date copy of the content directory and accept inbound uploads from end users.

- [x] Content

Monitor publishing of the new content into the content directory, respond to the reported publications and adjudicate possible dispute processes.

- [x] Builders

A diverse set of contributors, such as Developers, Designers and Product Managers, responsible for development of infrastructure and user facing applications.

- [x] Distribution

Run and maintain distributor nodes that deliver large volumes of upstream data to a large number of simultaneous end users.

- [x] HR 

Human Resources working group is responsible for integrating new members greeting, onboarding, catalyzing and nurturing, as well as managing bounties.

- [x] Marketing

Marketing group is responsible for increasing the outreach, sharing the content from the platform with the world and new members acquisition.

- [x] Membership

Membership group is responsible for new memberships invitations, referral rewards for existing members and overall process of adding more members via referral scheme.

**Out of scope**

- [ ] links to gitbook

[Storage>](https://joystream.gitbook.io/testnet-workspace/system/storage)
[Content >](https://joystream.gitbook.io/testnet-workspace/system/content-directory)
[Builders>](https://joystream.gitbook.io/testnet-workspace/system/builders)
[Distribution>](https://joystream.gitbook.io/testnet-workspace/system/storage#distributor)
[Marketing>](https://joystream.gitbook.io/testnet-workspace/system/marketers)
[Human Resources](https://joystream.gitbook.io/testnet-workspace/testnet/council-period-scoring/human-resources-score)
[Membership>](https://joystream.gitbook.io/testnet-workspace/testnet/council-period-scoring/human-resources-score)

PR #4523, opened 2023-09-04
Reduce `ForumPost` requests on category pages

mkbeefcake:
- #4484

integrated persist cache with local storage.
need to discuss more with @thesan about the point when can fetch 500 posts from server.

### DIFF
--- a/packages/ui/src/forum/hooks/useCategoryLatestPost.ts
+++ b/packages/ui/src/forum/hooks/useCategoryLatestPost.ts
@@ -1,7 +1,8 @@
+import * as Apollo from '@apollo/client'
 import { useEffect } from 'react'
 
 import { ForumPostOrderByInput, ForumThreadOrderByInput } from '@/common/api/queries'
-import { useGetForumPostsLazyQuery, useGetForumThreadsQuery } from '@/forum/queries'
+import { GetForumPostsDocument, useGetForumThreadsQuery } from '@/forum/queries'
 import { asForumPost, asForumThread } from '@/forum/types'
 
 export const useCategoryLatestPost = (category_eq: string) => {
@@ -25,7 +26,9 @@ export const useCategoryLatestPost = (category_eq: string) => {
       })
   }, [threadData])
 
-  const [fetchPost, { data: postData, loading: loadingPosts }] = useGetForumPostsLazyQuery()
+  const [fetchPost, { data: postData, loading: loadingPosts }] = Apollo.useLazyQuery(GetForumPostsDocument, {
+    fetchPolicy: 'cache-first',
+  })
   const rawPost = postData?.forumPosts[0]
 
   return {
--- a/packages/ui/src/forum/hooks/useForumCategoryThreadCount.ts
+++ b/packages/ui/src/forum/hooks/useForumCategoryThreadCount.ts
@@ -11,6 +11,7 @@ export const useForumCategoryThreadCount = (category_eq: string, isArchive?: boo
   }
   const { data } = useGetForumThreadsCountQuery({
     variables: { where: { category: { id_eq: category_eq }, status_json } },
+    fetchPolicy: 'cache-first',
   })
   return { threadCount: data?.forumThreadsConnection.totalCount }
 }
--- a/packages/ui/src/forum/hooks/useForumPopularThreads.ts
+++ b/packages/ui/src/forum/hooks/useForumPopularThreads.ts
@@ -21,6 +21,7 @@ export const useForumPopularThreads = ({ categoryId, page = 1, threadsPerPage = 
       offset: (page - 1) * threadsPerPage,
       limit: threadsPerPage,
     },
+    fetchPolicy: 'cache-first',
   })
 
   return {
--- a/packages/ui/src/forum/hooks/useThreadLatestPost.ts
+++ b/packages/ui/src/forum/hooks/useThreadLatestPost.ts
@@ -8,7 +8,7 @@ const orderBy = [ForumPostOrderByInput.UpdatedAtDesc]
 
 export const useThreadLatestPost = (threadId: string) => {
   const where = useMemo((): ForumPostWhereInput => ({ thread: { id_eq: threadId }, isVisible_eq: true }), [threadId])
-  const { data } = useGetForumPostsQuery({ variables: { where, orderBy, limit: 1 } })
+  const { data } = useGetForumPostsQuery({ variables: { where, orderBy, limit: 1 }, fetchPolicy: 'cache-first' })
   const rawPost = data?.forumPosts[0]
   return { post: rawPost && asForumPost(rawPost) }
 }


mkbeefcake:
Hello @thesan  I have tried to use local storage as persist cache for Apollo client. and then updated the code to use fetchPolicy as 'cache-only' for posts on Forum page. But i hope to know when and where will be the best to fetch 500 posts from the server and hope to get your opinion. 
Regards

mkbeefcake:
@thesan I have tried to reduce Graph queries by using memory cache by updating some calls.  Also I integrated to fetch 500 latest posts per one day, but it looks like not good effect for this. Please review and give me feedback.

chrlschwb:
fixes #4484

thesan (changes requested):
Hey @mkbeefcake, sorry I wasn't checking my GH notifications lately so I completely missed your 3 messages.

Again the idea of relying on Apollo fetch policies is great but sadly I haven't had a very good experience with the Apollo cache either.

I detailed a solution closer to what I had in mind in the comments. It's not as good in theory because it assumes that enough of the currently showing threads and sub-categories have a post which is one of the 500 last updated ones else it shows a "-".

Finally I just thought of this but it would probably be safer to fetch the 500 last **created** post rather than updated.

packages/ui/src/forum/hooks/useForumLatestPosts.ts
```diff
@@ -1,48 +1,29 @@
...
-  return fetchPolicy
+export type ForumPostsCache = {
+  expires: number
+  posts: ForumPost[]
 }
+const orderBy = [ForumPostOrderByInput.UpdatedAtDesc]
 
 export const useForumLatestPosts = () => {
-  const fetchPolicy = useCacheWithExpiration('last_LatestPosts')
+  const [cache, setCache] = useLocalStorage<ForumPostsCache>()
```
thesan, line 17: Sorry I forgot the key in my suggestion (so nothing is not working now). Something like `'latestForumPosts'`

packages/ui/src/forum/hooks/useForumLatestPosts.ts
```diff
@@ -1,48 +1,29 @@
...
-    setFetchPolicy('cache-and-network')
-    setLastCacheTime(currentDate.toDateString())
-  }, [lastCacheTime])
-
-  return fetchPolicy
+export type ForumPostsCache = {
+  expires: number
+  posts: ForumPost[]
 }
+const orderBy = [ForumPostOrderByInput.UpdatedAtDesc]
```
thesan, line 14: Since `orderBy` is already here let's put the whole `variables` here too:
```suggestion
const variables = { where: { isVisible_eq: true }, orderBy: [ForumPostOrderByInput.UpdatedAtDesc], limit: 500 }
```

packages/ui/src/forum/hooks/useCategoryLatestPost.ts
```diff
@@ -15,28 +17,29 @@ export const useCategoryLatestPost = (category_eq: string) => {
...
-    isLoading: loadingPosts || loadingThreads,
-    post: rawPost && asForumPost(rawPost),
-    thread: threadData?.forumThreads.length ? asForumThread(threadData?.forumThreads[0]) : undefined,
-  }
+  const [postsCache] = useLocalStorage<ForumPostsCache>()
+  return useMemo(
+    () => ({
+      isLoading: /*loadingPosts || */ loadingThreads,
+      post: postsCache?.posts.find((post) => post.categoryId === category_eq),
+      thread: threadData?.forumThreads.length ? asForumThread(threadData?.forumThreads[0]) : undefined,
```
thesan, line 41: I think this is not needed:
```suggestion
```
just use this in `@/forum/components/category/LatestPost.tsx`:
```ts
      <LatestPostLink to={`${generatePath(ForumRoutes.thread, { id: post.threadId })}?post=${post.id}`}>
```

thesan (changes requested):

thesan (commented):
Hey @mkbeefcake, thanks to you I finally got to test this idea. Here's the result on the "Joystream General" category:



1. It works fine for the sub categories 4 out 4 here.
2. It doesn't works well on the threads 5 out of 30.

This result is consistent on most categories. Sometime a few sub categories are missing their latest post but overall most sub categories have a latest post. However most thread latest posts are missing - except for the very active categories like "Pre-Proposals" which has a 30 out of 30 on the first page !

The solution could probably be tweaked a bit e.g:
- Increasing the amount cached posts from 500 to something a bit bigger.
- Decreasing the amount cached posts but filtering by category instead (for the threads last activity).
- Relying on the Apollo cache instead (like the way you started this PR) so that the missing still posts still get queried.
- Or a combination of these :point_up: 

However [when I proposed this caching solution](https://github.com/Joystream/pioneer/issues/4484#issuecomment-1691401969) the query node were in a really bad shape which resulted in Pioneer being unusable for most people, so I felt that we needed to act fast even if it meant sacrificing some of the functionalities in the short term. Back then maybe this solution would have been acceptable but IMO now it's not. As a result, I don't think we should be spending time trying to tweak this short term solution now that the problem is not urgent anymore. 

Instead I think we should focus on the long term solution which is https://github.com/Joystream/joystream/issues/2599

What do you think @mkbeefcake ?

chrlschwb:
@thesan can we complete this PR and merge it as a solution to the original issue?
#2599 requires a query node modification which might cost a lot of dev time and most likely this can only be attempted much later due to a large workload

thesan:
> @thesan can we complete this PR and merge it as a solution to the original issue? #2599 requires a query node modification which might cost a lot of dev time and most likely this can only be attempted much later due to a large workload

@chrlschwb even if this solution was viable this PR would need a bit more work, also #3868 would need to be finalized first (but I think this part should be done anyway). But even once its implementation is complete it turns out that the original idea I proposed isn't very good (see the screenshot I posted of the "Joystream General" category :point_up:). And since the QN are now working I don't think it's worth spending a lot of effort improving it because I think it will never be great, it will always be a short term trade-off until the QN issue can be addressed.

However [this change](https://github.com/Joystream/pioneer/pull/4523/files/54922262a769207710255052b97f4d1d4af7d571#diff-25e5b793b6bbdaa377c103b84dad7c5856d8a336eb6a7a35a7577b5a9ac6ff43) could be merged and maybe it could be applied to `useCategoryLatestPost`, `useForumLatestPosts`, `useThreadLatestPost`, and `useForumPopularThreads` (after all other changes on these files are reversed).

cc @mkbeefcake

ivanturlakov:
Tested on https://pioneer-2.vercel.app/#/forum/category/2
Mainnet

The number of requests when the page first loaded remained the same, but with continued interaction with the forum, the number of requests decreased significantly. Data display is OK 👍

|  | pioneerapp.xyz | pioneer-2.vercel.app|
|--------|--------|--------|
| First load | ~50 requests | ~50 requests |
| 2 | +50 | +10 |
| 3 | +50 | +10 |
| 4 | +50 | +10 |
| Total | 200 | 80 |

Issue #4507, opened 2023-08-27
Settings: Allow for custom query node to be used


mochet:
## Problem
Currently, the values for query node and RPC node on the settings page are hardcoded into Pioneer.
![image](https://github.com/Joystream/pioneer/assets/50830161/76261e46-bf66-41c6-8b5c-748a68e4d5f8)

Recently we had two issues:
* All query nodes were broken due to a bug (that has since been fixed)
* Slow query node performance (which has still not been resolved)

Due to this, users may want to use alternative query nodes. While it is possible to spin up alternative instances of Pioneer with different QNs set, this is problematic since the opportunity for fraud/scamming is much higher.

## Solution
* Add a `custom` option that allows a user to input any Query Node that they want to.
* Include a warning about the potential for fraud by using 3rd party QNs

dmtrjsg:
From my pov: 
1. Add an option to the drop down called "Custom"
2. When selected, render 3 separate input fields for URLs with labels: "Faucet", "RPC Node", "Query Node".  
3. The inputs require validation after submitted. So if URL does not resolve to the right place it would be smth like "Connection error". For individual endpoints.
4. When the custom option is selected, we'd need to add an info banner "Please beware that using 3rd party QN may result in fraud. Only use trusted providers."

msmadeline:
@dmtrjsg @chrlschwb The designs for this issue are ready to be reviewed. ✔

📁Link to Figma setting changelog: https://www.figma.com/file/ZwWMTt8ZvEPmFekDHCBWe1/Custom-query-node?type=design&node-id=5-2720&mode=design&t=YcFFRuL2UVaisG61-4

PR #4536, opened 2023-09-13
Allow for custom query node to be used

mkbeefcake:
- #4507

Updated the setting screen

### DIFF
--- a/packages/ui/src/app/pages/Settings/Settings.tsx
+++ b/packages/ui/src/app/pages/Settings/Settings.tsx
@@ -73,9 +73,9 @@ export const Settings = () => {
     try {
       const faucetStatusEndpoint = customFaucetEndpoint.replace(new RegExp('register$'), 'status')
       const response = await fetch(faucetStatusEndpoint)
-      await response.json()
-      setIsValidFaucetEndpoint(true)
-      return true
+      const succeeded = response.status < 400
+      setIsValidFaucetEndpoint(succeeded)
+      return succeeded
     } catch {
       setIsValidFaucetEndpoint(false)
       return false
@@ -84,44 +84,30 @@ export const Settings = () => {
 
   const checkRpcEndpoint = async () => {
     // check RPC endpoint
-    try {
+    return await new Promise<boolean>((resolve) => {
       const ws = new WebSocket(customRpcEndpoint)
-      ws.onopen = function () {
-        setIsValidRpcEndpoint(true)
-        return true
+      const willResolveTo = (succeeded: boolean, timeout?: any) => () => {
+        if (timeout)
+          clearTimeout(timeout)
+        
+        ws.close()
+        setIsValidRpcEndpoint(succeeded)
+        resolve(succeeded)
       }
-      ws.onerror = function () {
-        setIsValidRpcEndpoint(false)
-        return false
-      }
-    } catch {
-      setIsValidRpcEndpoint(false)
-      return false
-    }
 
-    // timeout
-    await new Promise((res) => setTimeout(res, 3000))
+      const timeout = setTimeout(willResolveTo(false), 3000)
+      ws.onopen = willResolveTo(true, timeout)
+      ws.onerror = willResolveTo(false, timeout)
+    })
   }
 
   const checkQueryEndpoint = async () => {
     // check GraphQL endpoint
     try {
-      const query = `
-        query {
-          electedCouncils{
-            id
-          }
-        }`
-      const response = await fetch(customQueryEndpoint, {
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-        },
-        body: JSON.stringify({ query }),
-      })
-      await response.json()
-      setIsValidQueryEndpoint(true)
-      return true
+      const response = await fetch(customQueryEndpoint + '?query=%7B__typename%7D')
+      const succeeded = response.status < 400 && JSON.parse(await response.json()).data['__typename'] === 'Query'
+      setIsValidQueryEndpoint(succeeded)
+      return succeeded
     } catch {
       setIsValidQueryEndpoint(false)
       return false
@@ -223,7 +209,7 @@ export const Settings = () => {
                           /^(ws|wss):\/\//i.test(customRpcEndpoint)
                             ? isValidRpcEndpoint
                               ? undefined
-                              : 'Connection Error, Sometimes it failed by network speed, Please try to check once more'
+                              : 'Connection Error. Sometimes it fails due to network speed. Please try to check once more'
                             : 'This RPC endpoint must start with ws or wss'
                         }
                       >
--- a/packages/ui/src/app/pages/Settings/Settings.tsx
+++ b/packages/ui/src/app/pages/Settings/Settings.tsx
@@ -1,4 +1,3 @@
-import { WsProvider } from '@polkadot/api'
 import React, { useState, useEffect } from 'react'
 import { useForm, FormProvider } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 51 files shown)


chrlschwb (approved):

chrlschwb:
fixes #4507

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -65,6 +99,68 @@ export const Settings = () => {
...
+                      <SettingsWarningInformation icon={<WarnedIcon />} title="Attention">
+                        <ColumnGapBlock gap={5}>
+                          <TextMedium lighter>
+                            Please beware that using 3rd party QN may result in fraud. Only use trusted providers
+                          </TextMedium>
+                        </ColumnGapBlock>
+                      </SettingsWarningInformation>
+                      <InputComponent
+                        label={t('customFaucet')}
+                        validation={/^(http|https):\/\//i.test(customFaucetEndpoint) ? undefined : 'invalid'}
```
thesan, line 113: [The design](https://www.figma.com/file/ZwWMTt8ZvEPmFekDHCBWe1/Settings?type=design&node-id=4-8214&mode=design&t=5XjKn64QWU2WMMVo-4) implies that the validation should check if the endpoints are reachable:
image

The rest of the app is now using [React Hook Form](https://react-hook-form.com/) I think it should be used here too.

thesan (changes requested):
About `packages/ui/src/app/pages/WorkingGroups/WorkingGroup/WorkingGroup.stories.tsx` and `packages/ui/src/app/pages/WorkingGroups/WorkingGroupsOpenings.stories.tsx` you're right that the interaction test were failing because these were missing, however instead of adding them manually you should let git get them from the `dev` branch by running:
```shell
git pull upstream dev
```
(with `upstream` being `git@github.com:Joystream/pioneer.git`).

Although you already added these manually, please still run this command to fix the diff in GH.

packages/ui/src/app/pages/Settings/Settings.tsx
mkbeefcake: Saying about checking connection, When will be the best point to do it? I guess while the user enters the url, it could be a bit complicated to check connection,  also I thought it could be not good point when the user click "save setting" button for checking connection. so i just implemented the validation check for url format..

thesan: The url validation is good we should keep this. But currently if someone makes a mistake when entering one of the endpoint they might not realize it for a while. So I think we should somehow check the 3 endpoints:
- For the `rpc` not it would be just opening a websocket and has soon as it's opened closing it.
- For the QN there should be a quick way to check that this is a valid gql server :crossed_fingers: .
- For the faucet there is a status endpoint at `../status` (so for `faucet.joystream.org/member-faucet/register` it's `faucet.joystream.org/member-faucet/status`).

You're right about avoiding validating while the users are typing. I think the validation should happen on input blur but after clicking on the button we should show a loader until the validation is completed (then it either update the local storage and reload the page if all endpoints were reached or show the "Connection error" if not).

mkbeefcake: @thesan I have updated the Endpoint validation code.

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -36,13 +41,108 @@ export const Settings = () => {
...
+        setIsValidQueryEndpoint(true)
+      } catch (err) {
+        setIsValidQueryEndpoint(false)
+        needReload = false
+      }
+
+      // check RPC endpoint
+      try {
+        const wsProvider = new WsProvider(customRpcEndpoint, false, undefined, 4000)
+        await wsProvider.connect()
```
thesan, line 122: This doesn't work for me I can save with any url starting with `wss://...` even if it doesn't exist.
Could you check if one of these works please?
```suggestion
        const wsProvider = new WsProvider(customRpcEndpoint)
        await wsProvider.isConnected
```
```suggestion
        const wsProvider = new WsProvider(customRpcEndpoint)
        await new Promise((resolve, reject) => {
          wsProvider.on('connected', resolve)
          wsProvider.on('error', reject)
        })
```
Also I think 4 seconds is way too long especially since the validation is `true` in the meantime.

thesan (changes requested):

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -36,13 +41,108 @@ export const Settings = () => {
...
+          queryNodeEndpoint: customQueryEndpoint,
+          membershipFaucetEndpoint: customFaucetEndpoint,
+          queryNodeEndpointSubscription: customQueryEndpoint.replace(/^http?/, 'ws'),
+          configEndpoint: undefined,
+        })
+        window.location.reload()
+      }
+    }
+
+    _update()
```
thesan, line 143: Also why is the function calling itself again ?

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -69,7 +68,85 @@ export const Settings = () => {
...
+          }
+        }`
+      const response = await fetch(customQueryEndpoint, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({ query }),
+      })
+      await response.json()
```
thesan, line 122: There is a simpler way to check if it's a gql endpoint: https://www.apollographql.com/docs/apollo-server/monitoring/health-checks/#graphql-level-health-checks

So here it would be something like:
```suggestion
      const response = await fetch(customQueryEndpoint + '?query=%7B__typename%7D')
      const succeeded = response.status < 400 && JSON.parse(await response.json()).data['__typename'] === 'Query'
```
This should be faster and also lighter on the QN

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -69,7 +68,85 @@ export const Settings = () => {
     }
   }, [network, endpoints])
 
-  const saveSettings = () => {
+  const checkFaucetEndpoint = async () => {
+    // check faucet endpoint
+    try {
+      const faucetStatusEndpoint = customFaucetEndpoint.replace(new RegExp('register$'), 'status')
+      const response = await fetch(faucetStatusEndpoint)
+      await response.json()
```
thesan, line 76: I think you could just check the status here:
```suggestion
      const succeeded = response.status < 400
```

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -204,7 +223,7 @@ export const Settings = () => {
                           /^(ws|wss):\/\//i.test(customRpcEndpoint)
                             ? isValidRpcEndpoint
                               ? undefined
-                              : 'Connection Error'
+                              : 'Connection Error, Sometimes it failed by network speed, Please try to check once more'
```
thesan, line 226:
```suggestion
                              : 'Connection Error. Sometimes it fails due to network speed. Please try to check once more'
```

thesan (changes requested):

packages/ui/src/app/pages/Settings/Settings.tsx
mkbeefcake: Thanks for your detailed suggestion. I learned a lot.

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -84,44 +84,29 @@ export const Settings = () => {
...
-        headers: {
-          'Content-Type': 'application/json',
-        },
-        body: JSON.stringify({ query }),
-      })
-      await response.json()
-      setIsValidQueryEndpoint(true)
-      return true
+      const response = await fetch(customQueryEndpoint + '?query=%7B__typename%7D')
+      const succeeded = response.status < 400 && JSON.parse(await response.json()).data['__typename'] === 'Query'
```
thesan, line 107: Hey, I think this line is failing because `await response.json()` actually returns a js object and not a string, my bad!
```suggestion
      const succeeded = response.status < 400 && (await response.json()).data['__typename'] === 'Query'
```
Since this should be the last change needed, I'll just accept the suggestion myself. Hopefully the formatting is correct :crossed_fingers:

packages/ui/src/app/pages/Settings/Settings.tsx
thesan: No problem!

packages/ui/src/app/pages/Settings/Settings.tsx
```diff
@@ -36,13 +40,113 @@ export const Settings = () => {
...
+
+        ws.close()
+        setIsValidRpcEndpoint(succeeded)
+        resolve(succeeded)
+      }
+
+      const timeout = setTimeout(willResolveTo(false), 3000)
+      ws.onopen = willResolveTo(true, timeout)
+      ws.onerror = willResolveTo(false, timeout)
+    })
```
thesan, line 100: Sorry I was wrong [here](https://github.com/Joystream/pioneer/pull/4536#discussion_r1350014987) the PR needs one more change:
In this https://github.com/Joystream/pioneer/pull/4536#discussion_r1348503309 I meant that you should keep the `try { ... } catch { ... }` (this is why there is an `await` after the `return`).

```suggestion
    try {
      return await new Promise<boolean>((resolve) => {
        const ws = new WebSocket(customRpcEndpoint)
        const willResolveTo = (succeeded: boolean, timeout?: any) => () => {
          if (timeout) clearTimeout(timeout)

          ws.close()
          setIsValidRpcEndpoint(succeeded)
          resolve(succeeded)
        }

        const timeout = setTimeout(willResolveTo(false), 3000)
        ws.onopen = willResolveTo(true, timeout)
        ws.onerror = willResolveTo(false, timeout)
      })
    } catch {
      setIsValidRpcEndpoint(false)
      return false
    }
```

thesan (approved):

polikosi:
Tested on: https://pioneer-2.vercel.app/#/settings

✅Add a custom option that allows a user to input any Query Node that they want to.
✅Include a warning about the potential for fraud by using 3rd party QNs
image

✅showing error when I put incorrect data 
image

Issue #3225, opened 2022-06-12
[USERSNAP] Past Councils > Working Groups: all data is wrong


dmtrjsg:
**Taken at**: https://dao.joystream.org/#/council/past-councils/0000000a
**Fields**:
- Houston, we have a problem.: Past Councils &gt; Working Groups: all data is wrong
- Comment: SSIA
- Feedback: tomato#9464

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/efd1c1b2-6984-46ab-b230-80575b8c673d-annotated_12a2b2ea-dd96-46b6-a9b8-fc2842411f45.png?etag=26089cfb927f671bae2cd670193f6f78)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/efd1c1b2-6984-46ab-b230-80575b8c673d-annotated_12a2b2ea-dd96-46b6-a9b8-fc2842411f45.png?etag=26089cfb927f671bae2cd670193f6f78)*

**Browser**: Chrome 102 (Windows 10)
**Screen size**: 3440x1440
**Browser size**: 1556x1206
**[Open #276 in Usersnap](https://app.usersnap.com/l/feedback/ace8b081-914a-4365-9a8b-0b9579887c24?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

chrlschwb:
Is this issue still standing?

ivanturlakov:
Tested on https://pioneerapp.xyz/#/council/past-councils/00000003
Mainnet

@chrlschwb Looks like still standing
Pls check this screen(BWG stats)
Looks wrong to me
BTW the last two columns always = 0

<img width="1211" alt="Снимок экрана 2023-02-02 в 14 12 54" src="https://user-images.githubusercontent.com/46903215/216321973-8af8cfc3-c14c-4760-9c05-e2a049132564.png">

PR #4553, opened 2023-09-26
Fix-3225: used BudgetUpdatedEvent instead of BudgetSetEvent

mkbeefcake:
Fixed the issue #3225 

image

### DIFF
--- a/packages/ui/src/council/queries/council.graphql
+++ b/packages/ui/src/council/queries/council.graphql
@@ -21,6 +21,7 @@ fragment PastCouncilProposalsFields on ProposalVotedEvent {
 fragment PastCouncilWorkingGroupFields on WorkingGroup {
   id
   name
+  budget
 }
 
 fragment PastCouncilBudgetSetEventFields on BudgetSetEvent {
@@ -37,6 +38,11 @@ fragment PastCouncilNewMissedRewardLevelReachedEventFields on NewMissedRewardLev
   groupId
   workerId
   newMissedRewardAmount
+}
+
+fragment PastCouncilBudgetUpdatedEventFields on BudgetUpdatedEvent {
+  groupId
+  budgetChangeAmount
 }
 
 fragment ElectedCouncilFields on ElectedCouncil {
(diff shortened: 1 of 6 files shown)


chrlschwb (approved):

packages/ui/src/council/queries/council.graphql
```diff
@@ -252,6 +258,9 @@ query GetPastCouncilWorkingGroups($fromBlock: Int!, $toBlock: Int!) {
   newMissedRewardLevelReachedEvents(where: { inBlock_lte: $toBlock }, orderBy: [inBlock_DESC]) {
```
thesan, line 258: Shouldn't `newMissedRewardLevelReachedEvents` also be restricted to the council term:
```suggestion
  newMissedRewardLevelReachedEvents(where: { inBlock_gte: $fromBlock, inBlock_lte: $toBlock }) {
```
Or am I missing something ?

Also I think `budgetSetEvents` should be removed, or is this query is used somewhere else ?

mkbeefcake: I checked budgetSetEvents is not used in the source code. saying about newMissedRewardLevelReachedEvents, that was the original code, so i will follow you.

mkbeefcake: I updated the #3225

thesan (approved):
Thanks

ivanturlakov:
Tested on https://pioneer-2.vercel.app/#/council/past-councils/0000000j
Mainnet


✅ The budget columns look good
✅ The final budget matches the figure from the consuls' report

Screenshot 2023-10-03 at 15 23 24


Screenshot 2023-10-03 at 14 24 53

ivanturlakov:
I haven't figured out how to check if the data in the first two columns is correct yet. Checked the reports of the consuls and group leaders for the corresponding term, but unfortunately, there are no such figures there

ivanturlakov:
The page still needs correction, the data marked does not correspond to the data in the reports of consuls and working group leaders

Screenshot 2023-10-11 at 14 16 11

chrlschwb:
@ivanturlakov the `total spent` and `total spent on proposals` fields are going to be corrected in another PR
https://github.com/Joystream/pioneer/pull/4554
So, this PR should only fix the below fields

PR #4532, opened 2023-09-08
Rwd mvp

eshark9312:
resolves #1794

PR #4556, opened 2023-09-28
Add an election progress bar

### LINKED ISSUE #4006: Enhance election lifecycle clarity

mkbeefcake:
Fixed #4006 : implemented new Progressbar for Election

[> Design](https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?type=design&node-id=11664-445586&mode=design&t=jyjAv6fOUt9c3JKK-0)
[> RWD Designs](https://www.figma.com/file/EaObQEN95O0IWjSVTUQ5NN/Pioneer-RWD?type=design&node-id=593-91321&mode=design&t=txKVA01bjXMQvgHV-0)
[> Election page](https://dao-git-fork-mkbeefcake-fix-4006-joystream.vercel.app/#/election)
[> Story](https://pioneer-2-storybook-git-fork-mkbeefcake-fix-4006-joystream.vercel.app/?path=/story/pages-election-election--announcing)


chrlschwb (changes requested):
I see two problems:
1. When hovered over, the bar's thickness increases. It's visually unappealing.
2. It doesn't support mobile view

mkbeefcake:
> I see two problems:
> 
> 1. When hovered over, the bar's thickness increases. It's visually unappealing.
https://www.figma.com/file/GlgN8uBRtvtMJtiOsdtDF7/Pioneer-design?type=design&node-id=11664-445586&mode=design
I followed this figma design, and there is the increasement of bar when hovered

> 2. It doesn't support mobile view
There was no mobile design, and also I guess this pioneer website is not ready for mobile responsive yet... so I only focused about desktop

chrlschwb:
RWD 
https://github.com/Joystream/pioneer/issues/4006#issuecomment-1752624649

chrlschwb:
RWD will be implemented in https://github.com/Joystream/pioneer/pull/4532

chrlschwb (approved):

polikosi:
Checked on: https://pioneer-2-git-fork-mkbeefcake-fix-4006-joystream.vercel.app/#/election

Current stage - Revealing:
image

Announcing stage:
image

Voting stage:
image

Idle stage:
image

thesan (changes requested):
@mkbeefcake really nice work the component looks pretty good. However some tests are needed.

To do this please start by creating a story at `@/app/pages/Election/Election.stories.tsx`. You can use: `@/app/pages/Proposals/ProposalPreview.stories.tsx` or `@/app/pages/WorkingGroups/WorkingGroup/WorkingGroup.stories.tsx` as examples. No need to mock all the data of the page, but the period length and the current stage (and maybe block) should be controlled by the story `args` so there will be a bit of logic in the `parameters.mocks`. Once this is done actually writing the tests should be straight forward.

Apart from this, I noticed that when I mouse hover the bars all elements are pushed up a bit. Could make the bar grow on hover without moving other elements please ?

thesan:
> Idle stage: image

This one looks wrong actually. According to the design the copy should be something like "Ide stage begins in 05:07:05:59"

chrlschwb:
@mkbeefcake 
you haven't fixed this yet.
https://github.com/Joystream/pioneer/pull/4556#issuecomment-1761323745

image
Revealing stage and Idle stage

chrlschwb:
@thesan @dmtrjsg @msmadeline @mkbeefcake 

In blockchain the idle stage precedes the announcing stage. I think this should be visualized here as well.
So idle stage should be first instead of last.

dmtrjsg:
@chrlschwb at the moment there's 1 block of idle stage so its not too practical to display it.

dmtrjsg:
We believe that this issue should be addressed in this PR as well
- https://github.com/Joystream/pioneer/issues/4571

mkbeefcake:
@dmtrjsg @chrlschwb @thesan @polikosi Sorry for delayed work. I have removed idle period since its length is 1, and then have integreated stories.tsx, also fixed the issue.

chrlschwb:
@mkbeefcake 
- Idle period should be put as the first period in the progress bar. Idle period is 14 days in Nara runtime, so it cannot be ignored.
- https://github.com/Joystream/pioneer/pull/4556#issuecomment-1761323745
this isn't fixed yet. Once a period is over, the tooltip should display "XXX stage begins in DD:HH:MM:SS"
- https://github.com/Joystream/pioneer/pull/4556#issuecomment-1772508880
The tooltip should not disappear after clicking on the bar.
- There is a small spelling mistake.
image

mkbeefcake:
@chrlschwb I have moved the idle period to the first, and then fixed spell issue, disappearing tooltip at mouse click and etc. 
Saying about "Once a period is over, the tooltip should display "XXX stage begins in DD:HH:MM:SS", I followed the figma design. I have implemented this "begins in" text, when the mouse is over on the period. Please let me know the detail about this issue.

chrlschwb (approved):

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -0,0 +1,623 @@
...
+    announcingDays = Math.floor(constants?.announcingPeriod / constants?.budgetRefillPeriod)
+    votingDays = Math.floor(constants?.election.votingPeriod / constants?.budgetRefillPeriod)
+    revealingDays = Math.floor(constants?.election.revealingPeriod / constants?.budgetRefillPeriod)
+    inactiveDays = Math.floor(constants?.idlePeriod / constants?.budgetRefillPeriod)
+
+    progressBarAttr = `${inactiveDays > 0 ? inactiveDays : 1}fr ${announcingDays > 0 ? announcingDays : 1}fr ${
+      votingDays > 0 ? votingDays : 1
+    }fr ${revealingDays > 0 ? revealingDays : 1}fr`
+
+    if (props.electionStage === 'inactive') {
```
thesan, line 113: In this long `if { } else if { ...` the same code is repeated with just different stages please could you try to refactor this a function which doesn't repeat itself so much. As an example you can check the code of `packages/ui/src/council/hooks/useCouncilRemainingPeriod.ts`:

Line 23: an index is being assigned to each stage (in this case inactive should probably be first instead of last).
Line 17: duration are put in an array in order to get them by index

I think that by using similar a method this long file can be made a lot shorter and maintainable

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -0,0 +1,623 @@
...
+      } else {
+        setStageDescription(defaultStaging)
+        setVerbIndicator(defaultVerbIndicator)
+        setDaysIndicator(defaultDaysIndicator)
+      }
+      setSelectedToolbarStage('')
+      return
+    }
+
+    if (props.electionStage === 'inactive') {
```
thesan, line 291: The same thing can be done here

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -0,0 +1,623 @@
...
+          <StatisticBigLabel>{daysIndicator}</StatisticBigLabel>
+        </StatisticBigLabel>
+        <StatisticLabel>{remainDays} days until next round</StatisticLabel>
+      </StatisticItemSpacedContent>
+
+      <ProgressBarLayout layout={progressBarAttr}>
+        <TooltipProgressBar
+          start={0}
+          end={inactiveProgress / 100}
+          color={inactiveColor}
```
thesan, line 362: Please change this prop to a `isCurrent` boolean, then inside the `TooltipProgressBar`: `Colors.Blue[500]` is passed if the stage is current and `Colors.Blue[200]` if not

thesan (changes requested):

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
mkbeefcake: Hello Thesan. Thanks for your detailed guideline and update. I will follow you and update this

mkbeefcake: I have followed your suggestion except the refactoring it.  Please give me some more time to do this

thesan (changes requested):
@mkbeefcake please check the new comments. The main goal is to reduce the complexity of `ElectionProgressBar`. The changes you made already reduced the code repetition a bit, but I think this component could still be much smaller and easier to read.

packages/ui/src/council/hooks/useCouncilRemainingPeriod.ts
```diff
@@ -34,3 +34,34 @@ export const useCouncilRemainingPeriod = (until: 'stageEnd' | 'electionEnd' = 's
...
+        const durations = [
+          api.consts.council.announcingPeriodDuration,
+          api.consts.referendum.voteStageDuration,
+          api.consts.referendum.revealStageDuration,
+          api.consts.council.idlePeriodDuration,
+        ].map(Number)
+        const position = ['announcing', 'voting', 'revealing', 'inactive'].indexOf(stage)
+        const offset = until === 'stageEnd' ? 1 : durations.length
+
+        return start + sum(durations.slice(position, position + offset))
```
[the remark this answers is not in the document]
mkbeefcake: Hello @thesan  I have tried to change useCouncilRemainingPeriods function, but there are many references for this (from 5 places), so if I updated the returned data stucture, i guess i have to update these 5 places which is not associated with Progressbar at this moment. I guess this could cause some confuse at this moment, so hope to update this later in other issue. For example, in 5 places, 3 places uses until to 'electionEnd' and 2 places uses 'stageEnd' so we have to investigate this stuff before update this function.

packages/ui/src/app/pages/Election/Election.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+              },
+            },
+          ],
+        },
+      }
+    },
+  },
+} satisfies Meta<Args>
+
+export const Default: Story = {}
```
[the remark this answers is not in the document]
mkbeefcake: @thesan I have updated the code following your suggestion except useCouncilRemainPeriod, Please review my patches. Thanks

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -0,0 +1,459 @@
...
+      currentBlock +
+      [duration + sum(gaps.slice(0, 2)), duration + sum(gaps.slice(1, 2)), duration, duration - sum(gaps.slice(2, 3))][
+        currentPosition
+      ]
+
+    revealingEndBlock =
+      currentBlock +
+      [duration + sum(gaps.slice(0, 3)), duration + sum(gaps.slice(1, 3)), duration + sum(gaps.slice(2, 3)), duration][
+        currentPosition
+      ]
```
thesan, line 177: Hi @mkbeefcake the change I suggested here: https://github.com/Joystream/pioneer/pull/4556#discussion_r1409691009 is not about `useCouncilRemainingPeriods`. It's fine to have the `useCouncilPeriodInformation` hook (maybe it should be in it's own file but that's a detail). What I suggested was to change the `useCouncilPeriodInformation` some more so it can simplify this code.

The goal of this suggestion is to change this part of the code to something like this:

```suggestion
    const [inactiveEndBlock, announcingEndBlock, votingEndBlock, revealingEndBlock] =
      periodInformation?.periodEnds ?? []

    const endDates = periodInformation?.periodEnds.map((block) => blockToDate(block, currentBlock))
    const [inactiveEndDay, announcingEndDay, votingEndDay, revealingEndBlock] =
      endDates ?? []
      
    const progresses = periodInformation && periodInformation.periodEnds.map((end, index) => {
      const start = periodInformation.periodStarts[index]
      return clamp((currentBlock - start) * 100 / (end - start), 0, 100)
    })
    const [inactiveProgress, announcingProgress, votingProgress, revealingProgress] =
      progresses ?? []

    const remainDays = periodInformation && blockDurationToDays(periodInformation.remainingPeriod)
```
Here I put `const ...` because this could be moved out of the `if` (so there would be no need for all the `let ...`).

Also `periodInformation` is the output of the change I suggested for `useCouncilPeriodInformation`. I think this hook can be called in this component directly (no need to call it in `packages/ui/src/app/pages/Election/Election.tsx`).

Finally `blockToDate` is the version I suggested which converts a block height into a date (and not a block duration to a date), and [`clamp`](https://lodash.com/docs/4.17.15#clamp) from Lodash.

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -0,0 +1,459 @@
...
+
+  const constants = useCouncilConstants()
+
+  const blockDurationToMs = (blockDuration: number) => blockDuration * MILLISECONDS_PER_BLOCK
+  const blockToDate = (duration: number) => {
+    const now = Date.now()
+    const msDuration = blockDurationToMs(duration)
+    return new Date(now + msDuration).toLocaleString('en-gb', { timeZone: 'Europe/Paris' })
+  }
+  const blockDurationToDays = (blockDuration: number) => Math.floor(blockDurationToMs(blockDuration) / A_DAY)
```
thesan, line 92: Please move this functions out of the component context (`blockDurationToMs`, `blockToDate`, `blockDurationToDays`). They can just be in the module context directly since they don't rely on some values from the component context.

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -0,0 +1,459 @@
...
+
+    const formatDurationDate = (duration: number): [string | number, string][] => {
+      if (duration * MILLISECONDS_PER_BLOCK < A_MINUTE) {
+        return [['< 1', 'm']]
+      }
+      return format(duration * MILLISECONDS_PER_BLOCK)
+    }
+
+    return <DurationValue value={formatDurationDate(duration)} />
+  }
```
thesan, line 47: Please also move `convertDurationToTimeString` out of the component context but also please make it into a component:
```tsx
const Duration = ({ duration }: { duration: number }) => {
```

packages/ui/src/app/pages/Election/Election.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+          ],
+        },
+      }
+    },
+  },
+} satisfies Meta<Args>
+
+export const Inactive: Story = {
+  args: { electionStage: 'inactive' },
+}
```
thesan, line 156: I just remembered that the election page is not displayed during the idle stage so no need to test it actually:
```suggestion
```

thesan (changes requested):
Last few changes

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -75,22 +72,19 @@ export const ElectionProgressBar = (props: ElectionProgressBarProps) => {
...
+  const progresses = periodInformation?.periodEnds.map((end, index) => {
+    const start = periodInformation.periodStarts[index]
+    return clamp(((currentBlock - start) * 100) / (end - start), 0, 100)
+  })
   const [inactiveProgress, announcingProgress, votingProgress, revealingProgress] = progresses ?? []
   const remainDays = periodInformation && blockDurationToDays(revealingEndBlock - currentBlock)
 
   if (
-    !isNaN(duration) &&
+    !isNaN(remainingPeriod) &&
```
thesan, line 87: Small change I'll accept myself:
```suggestion
```

thesan (approved):

thesan (changes requested):
@mkbeefcake thank you for the extra effort of implementing the full design but there are some 2 issues:

First: this one is not new but I didn't notice it with the desktop designs (also it doesn't happens on Firefox). [Watch the "ends on" time](https://github.com/Joystream/pioneer/assets/6571453/1272a0c4-cd33-4042-a01e-c981e4795926). It jumps back and forth from 18:22:06 to 18:22:11 also the value is slowly drifting. I think we should remove the seconds as we can't reliably predict this level of accuracy especially over several days. And this copies shouldn't be changed on every rerender, it should be changed only when the current stage changes.

 Second: the items are visibly misaligned vertically:
Screenshot from 2023-12-15 12-10-47

mkbeefcake:
@thesan I have fixed the issues by adding timestamp for GENESIS BLOCK and adjusted the padding for responsive. Please review this again. Thanks

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -27,11 +27,12 @@ interface ElectionProgressBarProps extends StatisticItemProps {
   electionStage: string
 }
 
+const GENESIS_BLOCK_TIMESTAMP = 1670693046000
```
thesan, line 30: It would be nice to be able to convert block/dates without fetching anything from the network, but Pioneer should avoid hardcoding these kind of data because it should potentially work on other networks too. Also basing the future block date based on a date which is that old is that old makes it less accurate.

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -81,8 +82,9 @@ export const ElectionProgressBar = (props: ElectionProgressBarProps) => {
   const constants = useCouncilConstants()
 
   const [inactiveEndBlock, announcingEndBlock, votingEndBlock, revealingEndBlock] = periodInformation?.periodEnds ?? []
-  const endDates = periodInformation?.periodEnds.map((block) => blockToDate(block - currentBlock))
+  const endDates = periodInformation?.periodEnds.map((block) => blockToDate(block))
```
thesan, line 85: Instead I think this is good enough:
```suggestion
  const endDates = useMemo(() => periodInformation?.periodEnds.map((block) => blockToDate(block - currentBlock)), [props.electionStage])
```
When the page is reloaded the time won't be exactly the same but it will actually be more accurate.

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -27,11 +27,12 @@ interface ElectionProgressBarProps extends StatisticItemProps {
...
+const GENESIS_BLOCK_TIMESTAMP = 1670693046000
+
 const blockDurationToMs = (blockDuration: number) => blockDuration * MILLISECONDS_PER_BLOCK
-const blockToDate = (duration: number) => {
-  const now = Date.now()
-  const msDuration = blockDurationToMs(duration)
-  return new Date(now + msDuration).toLocaleString('en-gb', { timeZone: 'Europe/Paris' })
+const blockToDate = (block: number) => {
+  const msDuration = blockDurationToMs(block)
+  return new Date(GENESIS_BLOCK_TIMESTAMP + msDuration).toLocaleString('en-gb', { timeZone: 'Europe/Paris' })
```
thesan, line 35: Finally here's how to remove the seconds:
```suggestion
  const date = new Date(now + msDuration)
  return `${date.toLocaleString('en-gb', { timeZone: 'Europe/Paris', dateStyle: "short", timeStyle: "short" })} CET` 
```
It's closer to the original design and it's a lot better since the time estimation can't be accurate to the second.

thesan (changes requested):

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
mkbeefcake: I have tried to use props.electionStage, at that time sometimes the endDate is calculated as undefined, so I changed it to periodInformation.currentStage.

packages/ui/src/app/pages/Election/components/ElectionProgressBar.tsx
```diff
@@ -82,7 +82,10 @@ export const ElectionProgressBar = (props: ElectionProgressBarProps) => {
   const constants = useCouncilConstants()
 
   const [inactiveEndBlock, announcingEndBlock, votingEndBlock, revealingEndBlock] = periodInformation?.periodEnds ?? []
-  const endDates = periodInformation?.periodEnds.map((block) => blockToDate(block))
+  const endDates = useMemo(
+    () => periodInformation?.periodEnds.map((block) => blockToDate(block - currentBlock)),
+    [periodInformation?.currentStage]
```
thesan, line 87: Good call! `periodInformation?.currentStage` is better since it will be defined when `currentBlock` will be defined.

thesan (approved):
Great work :raised_hands:

PR #4561, opened 2023-10-04
[backend] update schema

kdembler:
1. Adds `receiveEmails` field to member that indicates whether notifications should be sent as emails
2. Adds `unverifiedEmail` field to member that holds an email that wasn't yet verified
3. Renames `status` field on notification to `emailStatus`

### DIFF
--- a/packages/server/src/auth/api/member.ts
+++ b/packages/server/src/auth/api/member.ts
@@ -1,5 +1,5 @@
 import { Member } from '@prisma/client'
-import { intArg, mutationField, nonNull, objectType, queryField, stringArg } from 'nexus'
+import { booleanArg, intArg, mutationField, nonNull, objectType, queryField, stringArg } from 'nexus'
 import { Member as NexusMember } from 'nexus-prisma'
 import { error } from 'npmlog'
 import * as Yup from 'yup'
@@ -16,6 +16,8 @@ export const MemberFields = objectType({
     t.field(NexusMember.id)
     t.field(NexusMember.name)
     t.field(NexusMember.email)
+    t.field(NexusMember.unverifiedEmail)
+    t.field(NexusMember.receiveEmails)
   },
 })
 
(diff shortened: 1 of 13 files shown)


thesan (changes requested):
Great PR again! Thank you for maintaining the test suites :muscle: 

I just have 2 small suggestions

packages/server/src/auth/api/member.ts
```diff
@@ -63,28 +65,37 @@ export const verifyEmail = mutationField('verifyEmail', {
...
+      }
+
+      await sendVerificationEmail({ email, memberId: member.id, name: member.name, referer: req?.headers?.referer })
+      updatedMember = await prisma.member.update({ where: { id: member.id }, data: { unverifiedEmail: email } })
     }
 
-    await sendVerificationEmail({ email, memberId: member.id, name: member.name, referer: req?.headers?.referer })
+    if (receiveEmails != null) {
+      updatedMember = await prisma.member.update({ where: { id: member.id }, data: { receiveEmails } })
+    }
```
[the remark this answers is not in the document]
kdembler: Adapted your suggestion, please check

packages/server/src/notifier/model/notifications.ts
```diff
@@ -17,11 +17,9 @@ export const notificationsFromEvent =
       (acc, { id, receiveEmails }) => ({ ...acc, [id]: receiveEmails }),
       {}
     )
+    const allMemberIds = Object.keys(membersEmailPreferences).map(Number)
     const notifsByMembers = event.potentialNotifications.flatMap<PotentialNotifByMember>(
-      getEventsByMember(
-        subscriptions,
-        allMembers.map(({ id }) => id)
-      )
+      getEventsByMember(subscriptions, allMemberIds)
```
thesan, line 22: The performance issue I mentioned is that the partially applied function (`notificationsFromEvent(subscription)`) is called by the `flatmap` once per events, and on each call it recomputes both `membersEmailPreferences` and `allMemberIds` (which depends on the total number of users in the db).

So I was suggesting to take `membersEmailPreferences` and `allMemberIds` out of the inner function that runs once per events, into the outer function's scope which runs just once. Alternatively you could also compute these in the `createNotifications` function and pass them to `notificationsFromEvent` as additional parameters.

thesan (changes requested):

packages/server/src/notifier/model/notifications.ts
kdembler: Ahh I see now, I had trouble understanding what you suggested

thesan (approved):

PR #4258, opened 2023-03-23
Fix Election Blocktime Tooltip (#4156)

chrlschwb:
fixes issue #4156

PR #4614, opened 2023-10-31
Fix the block link opening past elections

chrlschwb:
fix wrong links in "past elections"

https://github.com/Joystream/pioneer/pull/4258#issuecomment-1783199572


packages/ui/src/council/components/election/pastElection/PastElectionsList/PastElectionsListRow.tsx
```diff
@@ -22,6 +22,7 @@ export const PastElectionsListRow = ({ election }: PastElectionsListRowProps) =>
       $colLayout={PastElectionsColLayout}
       as={GhostRouterLink}
       to={generatePath(ElectionRoutes.pastElection, { id: election.id })}
+      target="_blank"
```
thesan, line 25: This work but opening all links in a new tab isn't an ideal solution (it's inconsistent with the rest of the app). Instead could you try something like this on the `BlockLink` in the `BlockInfo` info component:
```jsx
onClick={evt => evt.stopPropagation()}
```

thesan (changes requested):

thesan (approved):
:+1:

PR #4624, opened 2023-11-09
Backend Docker GitHub workflow

mnaamani:
- #4282


thesan (approved):
@mnaamani I made a bunch of changes, it would be great if you could take a look at them to check if it's all right.

I'll make another PR documenting the deployment process. Hopefully we can do that this week.

Issue #4557, opened 2023-09-29
Proposal type


dmtrjsg:
## Context
In Nara release we need to create a new type of propsal "CRT feature management"

## Scope

- [ ] Creating proposal: allows toggling frozen status of, currently, only project_token pallet.
- [ ] Proposal preview

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205620250036159) by [Unito](https://www.unito.io)

thesan:
@dmtrjsg here's the updated feature description:
- #4897

msmadeline:
Hello! I implemented the feedback on the CRT proposal. 
➡Here's the link: https://www.figma.com/file/oD3UTElt3yc4JJfPJv4dn3/Proposals?type=design&node-id=12-12032&mode=design&t=TzOBxhex0Sk3bNjj-4

PR #4625, opened 2023-11-10
Freeze pallet proposal creation

vrrayz:
Create proposal for freeze pallet #4557 

[> Designs](https://www.figma.com/file/oD3UTElt3yc4JJfPJv4dn3/Proposals?type=design&node-id=1-2&mode=design&t=fga9AiQKHcfBrT5b-0)
[> Story](https://pioneer-2-storybook-git-fork-vrrayz-issue-4557-f67df9-joystream.vercel.app/?path=/story/pages-proposals-proposallist-current--default)
[> Preview](https://dao-git-fork-vrrayz-issue-4557-proposal-type-joystream.vercel.app/#/proposals/current)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205948614728979) by [Unito](https://www.unito.io)


packages/ui/src/proposals/model/proposalDescriptions.ts
```diff
@@ -33,6 +33,7 @@ export const proposalDescriptions: ProposalDescriptions = {
     'Membership Workgroup Lead is automatically assigned invitations upon taking the role. This proposals is aimed on managing this invitations count.',
   setReferralCut:
     'Referrals, same as new membership invitations are incentivised by the platform. Referral cut entails the reward to the originator of invitation links which resulted in new memberships.',
+  updatePalletFrozenStatus: 'You have the flexibility to enable or disable CRT feature.',
```
thesan, line 36: Based on: https://www.figma.com/file/oD3UTElt3yc4JJfPJv4dn3?node-id=3:17686&mode=design#610787926
```suggestion
  updatePalletFrozenStatus: 'Enabling and disabling various features on the blockchain level.',
```

thesan, line 36: Also please move this after `runtimeUpgrade` so that "Update Channel Payouts" remains (20) and "Runtime Upgrade" (21)

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
```diff
@@ -402,5 +409,8 @@ export const schemaFactory = (api?: Api) => {
         .required(),
       channelCashoutsEnabled: Yup.boolean(),
     }),
+    updatePalletFrozenStatus: Yup.object().shape({
+      frozen: Yup.boolean().required('Field is required'),
```
thesan, line 413: This is a bit tricky but it's really important that the proposal clarify things here by having an "execution" constraint. It means that if the `ProjectToken` pallet is currently enabled and someone try to enable it, they should get a warning warning telling them that the pallet is already enabled and that the proposal would fail if executed right now.

To do that you can pass "execution" as the validation name. There are already a few examples in this file.

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/UpdatePalletFrozenStatus/SelectPalletFrozenStatus.tsx
```diff
@@ -0,0 +1,58 @@
...
+      return 'Disabled'
+    }
+    return selectedStatus
+  }, [selectedStatus])
+  const change = (selected: PalletFrozenStatus, close: () => void) => {
+    onChange(selected)
+    close()
+  }
+  return (
+    <Select
```
thesan, line 30: @dmtrjsg suggested to use a toggle instead [here](https://www.figma.com/file/oD3UTElt3yc4JJfPJv4dn3?node-id=4:20973&mode=design#610789666)

thesan (changes requested):

packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/UpdatePalletFrozenStatus/UpdatePalletFrozenStatus.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { SelectPalletFrozenStatus } from './SelectPalletFrozenStatus'
+
+export type PalletFrozenStatus = 'Enabled' | 'Disabled'
+
+export const UpdatePalletFrozenStatus = () => {
+  const { watch, setValue } = useFormContext()
+
+  const setPalletFrozenStatus = (selected: PalletFrozenStatus) => {
+    if (selected === 'Enabled') {
+      setValue('updatePalletFrozenStatus.frozen', true, { shouldValidate: true })
```
thesan, line 18: So based on [this](https://github.com/Joystream/joystream/blob/nara/tests/network-tests/src/flows/proposals/index.ts#L161):
image

"Enabled" should mean frozen to `false` and "Disabled" frozen to `true`

Since a toggle should used instead of a select (see: https://github.com/Joystream/pioneer/pull/4625#discussion_r1389241248) this isn't so relevant anymore but keep this in mind when labeling the toggle. It's kinda counterintuitive.

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
```diff
@@ -402,5 +409,8 @@ export const schemaFactory = (api?: Api) => {
         .required(),
       channelCashoutsEnabled: Yup.boolean(),
     }),
+    updatePalletFrozenStatus: Yup.object().shape({
+      frozen: Yup.boolean().test(equalValueContext('palletFrozenStatus', 'execution')).required('Field is required'),
```
thesan, line 413: Please add a message field to `equalValueContext` so it actually display the error (like on `minContext` and `maxContext`). For example on Set Referral Cut:
image

You can see the warning and the box to go further but also there is a message: "Input must be equal or less than 50% for proposal to execute" which explains why the value is wrong.

Here it would be something like:
"The ProjectToken pallet is currently disabled, so presently this proposal would fail due to execution constraints."
Or
"The ProjectToken pallet is currently enabled, so presently this proposal would fail due to execution constraints."

thesan, line 413: Also this should test whether the value is different from the context not equal to it. So I think this function should be renamed:
```suggestion
      frozen: Yup.boolean().test(differentFromContext('palletFrozenStatus', 'execution')).required('Field is required'),
```

packages/ui/src/proposals/modals/AddNewProposal/AddNewProposalModal.tsx
```diff
@@ -70,6 +71,7 @@ export const AddNewProposalModal = () => {
   const maximumReferralCut = api?.consts.members.referralCutMaximumPercent
   const minCashoutAllowed = api?.consts.content.minimumCashoutAllowedLimit
   const maxCashoutAllowed = api?.consts.content.maximumCashoutAllowedLimit
+  const palletFrozenStatus = useObservable(() => api?.query.projectToken.palletFrozen(), [api?.isConnected])
```
thesan, line 74: Because this subscription doesn't need to stay opened please use `useFirstObservableValue` instead

packages/ui/src/mocks/data/proposals.ts
```diff
@@ -257,6 +261,7 @@ export const proposalsPagesChain = (
 
         proposalsEngine: { activeProposalCount },
         staking: { minimumValidatorCount },
+        projectToken: { palletFrozen: false },
```
thesan, line 264: Please make this an "args" of the story

packages/ui/src/proposals/modals/AddNewProposal/AddNewProposalModal.tsx
thesan, line 74: Also on this proposal the value of `updatePalletFrozenStatus.frozen` be set to `!palletFrozenStatus` by default. Otherwise when the form appears the value might already be wrong so the next button is disabled but it's not clear why:
image

packages/ui/src/app/pages/Proposals/CurrentProposals.stories.tsx
```diff
@@ -1405,7 +1405,24 @@ export const SpecificParametersSetMembershipPrice: Story = {
     })
   }),
 }
+export const SpecificParametersUpdatePalletFrozenStatus: Story = {
+  play: specificParametersTest('Update Pallet Frozen Status', async ({ args, createProposal, modal, step }) => {
+    await createProposal(async () => {
+      const nextButton = getButtonByText(modal, 'Create proposal')
+      expect(nextButton).toBeDisabled()
 
+      await userEvent.click(modal.getByTestId('crt-feature-select'))
```
thesan, line 1414: After you have addressed the other comments, this test should make sure the warning and validation error appear, then it can click again to make the form valid and continue

thesan (changes requested):

thesan (changes requested):
Last thing:

packages/ui/src/proposals/modals/AddNewProposal/helpers.ts
```diff
@@ -50,6 +50,9 @@ export const defaultProposalValues = {
   updateChannelPayouts: {
     cashoutEnabled: true,
   },
+  updatePalletFrozenStatus: {
+    pallet: 'ProjectToken',
+  },
```
thesan, line 55: I'm not 100% that would fix the failing test but could you try to set a default value for freeze to `false`

thesan (changes requested):

thesan (approved):

PR #4641, opened 2023-11-22
🧽 Notifications MVP last fixes

thesan:


chrlschwb (approved):

kdembler (approved):
Only thing I've noticed: if I don't have wallet connected and go to notification settings, the Subscribe button triggers membership select modal and not wallet connect modal. I think it worked correctly at some point. Otherwise, works great, so stoked to see this (almost) live!

thesan:
@kdembler thanks a lot for having a look.

> the Subscribe button triggers membership select modal and not wallet connect modal

Yes that's a problem everywhere in Pioneer (the same thing happens for create proposal, announce candidacy, etc...). I'll create an issue so this can get fixed globally at some point.

Issue #4595, opened 2023-10-26
Validator List - updated columns


dmtrjsg:
## Context
- https://github.com/Joystream/pioneer/issues/4321

## Designs
🆕 [Link to Figma >](https://www.figma.com/file/QnjZuWYtGimTEkQlN6MNlH/Validators?type=design&node-id=1601%3A44644&mode=design&t=nWRhVkKaUSy1Vu7k-1) 

## Scope
Update the columns in the validators table view
- [ ] Two variations of the validator card - bound to membership and unbound
Columns 🆕 :
- [ ] Verification status
- [ ] State (active/ waiting)
- [ ] Own staked
- [ ] Total staked
- [ ] Expected Nom APR (APR for nominators)
- [ ] Comission

traumschule:
Done

PR #4643, opened 2023-11-23
Update validator list columns

eshark9312:
- #4595 

[> Design](https://www.figma.com/file/QnjZuWYtGimTEkQlN6MNlH/Validators?type=design&node-id=1601-44644&mode=design&t=KzjgzblNnD3Ho09A-0)
[> Story](https://pioneer-2-storybook-git-fork-eshark9312-update-10e9a2-joystream.vercel.app/?path=/story/pages-validators-validatorlist--statistics-and-lists)


chrlschwb (approved):

packages/ui/src/validators/components/ValidatorsList.tsx
```diff
@@ -19,8 +19,10 @@ export const ValidatorsList = ({ validators }: ValidatorsListProps) => {
         <ListHeader>Validator</ListHeader>
         <ListHeader>Verification</ListHeader>
```
thesan, line 20: Please add a tooltip next to the "Verification" column header: image

The copy could be something like:

> The profile of Verified validator has been entirely verified by the [Membership working group](https://pioneerapp.xyz/#/working-groups/membership).

But it would be better to have the link point at the "About" tab and add the Working group description to this tab. See #4654

WDYT @dmtrjsg ?

packages/ui/src/validators/components/ValidatorsList.tsx
```diff
@@ -19,8 +19,10 @@ export const ValidatorsList = ({ validators }: ValidatorsListProps) => {
...
         <ListHeader>State</ListHeader>
-        <ListHeader>Total Reward</ListHeader>
-        <ListHeader>APR</ListHeader>
+        <ListHeader>Own Stake</ListHeader>
+        <ListHeader>Total Stake</ListHeader>
+        <ListHeader>Expected Nom APR</ListHeader>
+        <ListHeader>Comission</ListHeader>
```
thesan, line 25: The table should be sort-able by "Expected Nom APR" and "Commission"

image

packages/ui/src/validators/components/ValidatorsList.tsx
```diff
@@ -19,8 +19,10 @@ export const ValidatorsList = ({ validators }: ValidatorsListProps) => {
         <ListHeader>Validator</ListHeader>
         <ListHeader>Verification</ListHeader>
         <ListHeader>State</ListHeader>
-        <ListHeader>Total Reward</ListHeader>
-        <ListHeader>APR</ListHeader>
+        <ListHeader>Own Stake</ListHeader>
+        <ListHeader>Total Stake</ListHeader>
+        <ListHeader>Expected Nom APR</ListHeader>
```
thesan, line 24: I know it's not in the design but personally I could really use a tooltip on "Expected Nom APR" !

WDYT @dmtrjsg ?

packages/ui/src/validators/hooks/useValidatorsList.tsx
```diff
@@ -111,6 +111,15 @@ export const useValidatorsList = () => {
           isActive: activeValidators.includes(address),
           totalRewards: rewardHistory.reduce((total: BN, data) => total.add(data.eraReward), new BN(0)),
           APR: apr,
+          commission: validatorInfo.commission.toNumber() / 10 ** 7,
```
thesan, line 114: Please add a function in `@/common/utils.ts` (next to `clamp`):
```ts
export const perbillToPercent = (perbill: BN) => perbill.toNumber() / 10 ** 7
```

and use it here to clarify the code a bit.

thesan (changes requested):
Nice job, please check my suggestions :point_down:

packages/ui/src/validators/components/ValidatorsList.tsx
dmtrjsg: Tooltip is 100% warranted. 

Text: "This column shows the expected APR for nominators who are nominating funds for the chosen validator. The APR is subject to the amount staked and have a diminishing return for higher token amounts."

packages/ui/src/validators/components/ValidatorsList.tsx
dmtrjsg: Yes its a good idea, and we could just add the description to the "About" section which is exactly the same as the are on the working group cards on the WG landing page where they are shown as a list.

thesan (changes requested):
Also please mix the mocked data a bit so sorting can be tested in the story.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "editor.inlineSuggest.showToolbar": "always"
+}
```
thesan, line 3: Also please remove this file. It would be nice to have IDE specific settings relevant to the project (like formatting options) but I think this is more of a personal preference (which might not agree with someone else).

eshark9312: I missed to remove it.

packages/ui/src/validators/components/ValidatorsList.tsx
```diff
@@ -1,85 +1,90 @@
...
-        sortByCommission === 'desc' ? a.commission - b.commission : b.commission - a.commission
-      )
-    )
-  }, [sortByCommission])
+  const onSort = (key: SortKey) => {
+    if (key === sortBy) {
+      setDescending(!isDescending)
+    } else {
+      setDescending(false)
+      setSortBy(key)
```
thesan, line 39: I realized that it makes more sense to see the APR and commission by descending order by default. But the validator names should ascending by default.

WDYT of:
```suggestion
  const onSort = (key: SortKey, descendingByDefault = false) => {
    if (key === sortBy) {
      setDescending(!isDescending)
    } else {
      setDescending(descendingByDefault)
      setSortBy(key)
```

packages/ui/src/validators/components/ValidatorsList.tsx
```diff
@@ -1,85 +1,90 @@
...
         <ListHeader>Own Stake</ListHeader>
         <ListHeader>Total Stake</ListHeader>
-        <SortHeader
-          onSort={() => {
-            setSortByApr(!sortByApr ? 'desc' : sortByApr === 'desc' ? 'asc' : undefined)
-          }}
-          isActive={!!sortByApr}
-          isDescending={sortByApr === 'desc'}
-        >
+        <SortHeader onSort={() => onSort('APR')} isActive={sortBy === 'APR'} isDescending={isDescending}>
```
thesan, line 66: And:
```suggestion
        <SortHeader onSort={() => onSort('APR', true)} isActive={sortBy === 'APR'} isDescending={isDescending}>
```

packages/ui/src/validators/components/ValidatorsList.tsx
```diff
@@ -1,85 +1,90 @@
...
           </Tooltip>
         </SortHeader>
-        <SortHeader
-          onSort={() => {
-            setSortByCommission(!sortByCommission ? 'desc' : sortByCommission === 'desc' ? 'asc' : undefined)
-          }}
-          isActive={!!sortByCommission}
-          isDescending={sortByCommission === 'desc'}
-        >
+        <SortHeader onSort={() => onSort('commission')} isActive={sortBy === 'commission'} isDescending={isDescending}>
```
thesan, line 81: And:
```suggestion
        <SortHeader onSort={() => onSort('commission', true)} isActive={sortBy === 'commission'} isDescending={isDescending}>
```

thesan (changes requested):
Great work. Just one last detail

packages/ui/src/validators/components/ValidatorsList.tsx
eshark9312: I agree.

Issue #4586, opened 2023-10-19
[USERSNAP] Signal Proposal Text Area too small


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/proposals/preview/576
**Fields**:
- Houston, we have a problem.: Signal Proposal Text Area too small
- Comment: The signal proposal text should be displayed in a modal instead of a small sidebar

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/6d460616-3dd0-48ef-9a66-c9a394342676-annotated_444820b9-649c-4415-9910-0e293ea5a7e8.png?etag=4d3c752cbede5b625030a85f0be74c45)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/6d460616-3dd0-48ef-9a66-c9a394342676-annotated_444820b9-649c-4415-9910-0e293ea5a7e8.png?etag=4d3c752cbede5b625030a85f0be74c45)*

**Browser**: Chrome 118 (Windows 10)
**Screen size**: 1920x1080
**Browser size**: 1920x963
**[Open #431 in Usersnap](https://app.usersnap.com/l/feedback/bb97912c-301e-48d3-aec1-0d3e042ee092?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
Probably actual for some cases but not sure if that's getting high up in the prioritised list against other issues. I guess the design implied not using tables here and just a few paragraphs max..

msmadeline:
@chrlschwb @thesan @dmtrjsg Hello I finished designing the correct version of the signal description side drawer. 
Here's the link ➡https://www.figma.com/file/oD3UTElt3yc4JJfPJv4dn3/Proposals?type=design&node-id=64-13353&mode=design&t=mx7G3QiEM9KclbCO-4
Here's a link to RWD ➡https://www.figma.com/file/oD3UTElt3yc4JJfPJv4dn3/Proposals?type=design&node-id=64-9140&mode=design&t=mx7G3QiEM9KclbCO-4

thesan:
[...]

PR #4676, opened 2023-12-05
modal-small

GoldWolf115:
https://github.com/Joystream/pioneer/issues/4586


thesan (changes requested):
Not all SidePane should be widen, otherwise some of them like the candidate or membership panes which were design to be narrow will look strange:
image

For now I think only the signal SidePane should be widen (maybe by passing it a property.

PR #4683, opened 2023-12-07
Blacklisted accounts

vrrayz:
Fix for https://github.com/Joystream/pioneer/issues/4138

### DIFF
--- a/packages/ui/src/common/components/Warning.tsx
+++ b/packages/ui/src/common/components/Warning.tsx
@@ -35,7 +35,11 @@ export const Warning = ({ title, content, isClosable, additionalContent, icon, i
           {title && <h5>{title}</h5>}
         </HeaderWrapper>
       )}
-      {content && <TextMedium light>{content}</TextMedium>}
+      {content && (
+        <TextMedium as="div" inter light>
+          {content}
+        </TextMedium>
+      )}
       {additionalContent}
     </WarningBlock>
   )


packages/ui/src/mocks/data/proposals.ts
```diff
@@ -242,7 +242,7 @@ export const proposalsPagesChain = (
           nextRewardPayments,
           stage: { stage: { isIdle: true }, changedAt: 123 },
         },
-        referendum: { stage: {} },
+        referendum: { stage: {}, accountsOptedOut: { keys: [] } },
```
thesan, line 245: So with the `skip` argument, this is not necessary
```suggestion
```

packages/ui/src/mocks/helpers/asChainData.ts
```diff
@@ -21,3 +22,8 @@ export const asChainData = (data: any): any => {
 }
 
 const withUnwrap = (data: Record<any, any>) => Object.defineProperty(data, 'unwrap', { value: () => data })
+// const withComputed = (data: Record<any, any>) => Object.defineProperties(data, {
+//   unwrap: { value: () => data },
+//   isSome: { value: Object.keys(data).length > 0 },
+//   keys: { value: () => of(Object.keys(data).map(entry => ({ args: [entry] }))) },
+// })
```
thesan, line 29: Please remove the commented code on this file.

You're right the value can't be returned from the `asChainData`, instead something similar could be done on `asApiMethod` but it's not urgent right now...

thesan (changes requested):

packages/ui/src/app/pages/Election/BlacklistedAccounts/BlacklistedAccounts.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+  component: BlacklistedAccounts,
+  parameters: {
+    mocks: (): MocksParameters => {
+      return {
+        chain: {
+          derive: {
+            balances: {
+              all: {},
+            },
+          },
```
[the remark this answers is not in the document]
vrrayz: > Are you sure this is needed ?

This is the error that shows up if i don't include it 
image

I think its because the `useBalance` hook is called which requires the `all()` 
image

packages/ui/src/app/pages/Election/BlacklistedAccounts/BlacklistedAccounts.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+  },
+  [allAccounts[2]]: {
+    ...mockDefaultBalance,
+    total: new BN(200_000_000_000),
+  },
+  [allAccounts[3]]: {
+    ...mockDefaultBalance,
+    total: new BN(200_000_000_000),
+  },
+}
```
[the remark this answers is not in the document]
vrrayz: > I don't think this is needed:

After removing this and implementing other changes this error shows up
image

When writing the storybook test i came across  the error earlier on and i figured mocking the locks that way would be the fix for it which was i used this. In `toBalance`  the lockdetails are been used so i thought the storybook test needed such mock data
image

thesan (changes requested):
Also it seems the balance encoding broke some tests. Please have a quick look. If there's no quick solution I'll fix it tomorrow.

Finally even if it works on storybook, this will require some actual QA before merging it. @ivanturlakov could you please create a bunch of accounts on Atlas dev transfer them some funds and Opt out of voting with them please ?
image

ivanturlakov:
> ould you please create a bunch of accounts on Atlas dev transfer them some funds and Opt out of voting with them please ?

@thesan sure, what atlas instance should I use?

thesan:
> @thesan sure, what atlas instance should I use?

@ivanturlakov I meant the Atlas dev network. However it seems that there is an opted out account already. So I guess nothing more is needed. Sorry I should have checked yesterday but I was convinced there was no opted out account on mainnet ATM.

@vrrayz please check: https://dao-git-fork-vrrayz-blacklisted-accounts-joystream.vercel.app/#/election/blacklisted-accounts there is a bug with the total counter, my guess is that it comes from: https://github.com/Joystream/pioneer/pull/4683#discussion_r1439576874 I'm surprised it didn't show on storybook (I guess the full app is triggering a lot more re-renders).

packages/ui/src/app/pages/Election/BlacklistedAccounts/BlacklistedAccount.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { TokenValue } from '@/common/components/typography'
+import { Colors } from '@/common/constants'
+import { shortenAddress } from '@/common/model/formatters'
+
+interface Prop {
+  address: string
+}
+
+export const BlacklistedAccount = ({ address }: Prop) => {
+  const balance = useBalance(address)
```
thesan, line 17: For most users `useBalance` will call `derive.balances.all`. RPC node queries are not cached and they should be reduced as much as possible. Since `useAddresses` is called on this page, this component shouldn't call `useBalance`. Instead:
```suggestion
export const BlacklistedAccount = ({ address, balance }: Prop) => {
```

packages/ui/src/app/pages/Election/BlacklistedAccounts/BlacklistedAccounts.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+interface TotalBalanceProp {
+  addresses: string[]
+}
+
+export const BlacklistedAccounts = () => {
+  const ACCOUNTS_PER_PAGE = 18
+  const [page, setPage] = useState(1)
+  const votingOptOutAccounts = useVotingOptOutAccounts()
+  const [paginatedAccounts, setPaginatedAccounts] = useState<string[] | undefined>()
```
thesan, line 29: As a result `useBalances` should be called on this component instead. Also I think the sum can be done here (no need for `TotalBalance`), and the paginated accounts should be `{ address: string, balance: BN | undefined }`:
```suggestion
  const balances = useBalances(addresses)
  const totalBalance = useMemo(
    () => balances.values().reduce((prev, value) => value?.total.add(prev) ?? prev, BN_ZERO),
    [balances]
  )
  const paginatedAccounts = useMemo(
    () =>
      votingOptOutAccounts
        .slice((page - 1) * ACCOUNTS_PER_PAGE, page * ACCOUNTS_PER_PAGE)
        .map((address) => ({ address, balance: balance.get(address)?.total }))
    [votingOptOutAccounts, page, balances])
  )
```

packages/ui/src/accounts/hooks/useBalance.ts
```diff
@@ -14,14 +18,42 @@ export const useBalance = (address: Address = ''): Balances | null => {
...
+    return isDefined(allMyBalances) && address && !myBalances ? api?.derive.balances.all(address) : undefined
+  }, [address, !myBalances])
 
   if (myBalances) {
     return myBalances
   }
 
   return balances ? toBalances(balances) : null
 }
+export const useBalances = (addresses: Address[]): Map<string, Balances | undefined> => {
```
thesan, line 31: Small detail but please skip a line here:
```suggestion

export const useBalances = (addresses: Address[]): Map<string, Balances | undefined> => {
```

packages/ui/src/accounts/hooks/useBalance.ts
```diff
@@ -14,14 +18,42 @@ export const useBalance = (address: Address = ''): Balances | null => {
...
+  const { api } = useApi()
+  const allMyBalances = useMyBalances()
+
+  const [definedBalances, setDefinedBalances] = useState(new Map<string, Balances>())
+  useEffect(() => {
+    addresses.forEach((address) => {
+      if (definedBalances.get(address)) return
```
thesan, line 38: Even smaller detail (and I know it's something I wrote) but:
```suggestion
      if (definedBalances.has(address)) return
```

thesan (changes requested):

thesan (approved):
LGTM!

I'm just waiting for #4713 to merge this

Issue #4125, opened 2023-01-26
[USERSNAP] Cant unstake from My Profile


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/election/past-votes
**Fields**:
- Houston, we have a problem.: Cant unstake from My Profile
- Comment: I managed to unstake my funds for voting in the last election successfully. Other keys do not show the unstake option, only the Show Election button. I managed to unstake only through Explorer or through Past votes.

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/74f9830b-3fb2-411e-987e-8e70fc71b8bd-annotated_26a54ecc-20e9-4d5d-bfb2-de9bb8509fb3.png?etag=18fb794f4afc0918c2e0b83b99ac62a5)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/74f9830b-3fb2-411e-987e-8e70fc71b8bd-annotated_26a54ecc-20e9-4d5d-bfb2-de9bb8509fb3.png?etag=18fb794f4afc0918c2e0b83b99ac62a5)*

**Browser**: Yandex.Browser 23 (Windows 10)
**Screen size**: 1536x864
**Browser size**: 1536x726
**[View screen recording](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/screen_recording/481b511b-f45c-4c7a-a618-756529610970_screen_recording_e0ee8ed7-3b2d-40e8-b85c-e58b150ab053.webm?etag=7e09081252d70a46da5a26c84aa232c5)**
**[Open #383 in Usersnap](https://app.usersnap.com/l/feedback/945555bc-db84-44cd-b6a2-3539f80efd11?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
Assigning high-prio label as all council-related tasks for the benefit of voters.

⚠️ In order to debug it properly, we need the exact flow on how to reproduce this issue.

igrexac:
In short: My Profile incorrectly displayed information about the stake on some of my keys.
1. In the My Profile section, I successfully returned a stake for one of the candidates from one key.
[...]

PR #4685, opened 2023-12-08
unstaking issue fix

vrrayz:
Fix for https://github.com/Joystream/pioneer/issues/4125


packages/ui/src/accounts/components/AccountItem/components/RecoverButton.tsx
```diff
@@ -16,8 +16,6 @@ interface Props {
 export const RecoverButton = React.memo(({ memberId, lock, address, isRecoverable, isSmall }: Props) => {
   const { showModal } = useModal()
   const onClick = useCallback(() => {
```
thesan, line 18: However "Voting" is the only lock that doesn't require a member id. I think the other ones will fail if it's not provided. And even though most of the time the member id should be there, if the locked staking account is present but the member controller account is not, the member id will be missing. Also it might be missing due to QN issues.

So here's what I'm proposing:
```suggestion
  const isVotingLock = lock.type === 'Voting'
  const onClick = useCallback(() => {
```

And `TransactionButton` should have a disabled state:
```tsx
<TransactionButton ... disabled={!isVoting && !memberId}
```

thesan (changes requested):

thesan (approved):
:+1:

Issue #4591, opened 2023-10-25
User profile: Add GitHub social link


mochet:
# Problem
Similar to: https://github.com/Joystream/pioneer/issues/4540
A very common question from users researching projects will be to understand who the development team is--while we have some limited info on the main website it is far more attractive for developer GitHub profiles to be viewable on their Joystream profile within Pioneer such that a user can visit the Builders WG and see the GH accounts of active builders.
Obviously the UI could be improved to showcase this data better at a later point in time, but it would be good to have the option to link to GH profiles for the purpose of verification and research.

dmtrjsg:
Can be done together with 
- #4540

thesan:
@GoldWolf115 even though everything was merged on `dev`, I tried to set my GH handle and it didn't work. So I'll reverse the changes on Pioneer until it can be debugged.

PR #4695, opened 2023-12-12
github name

GoldWolf115:
- https://github.com/Joystream/pioneer/issues/4591


packages/ui/package.json
```diff
@@ -26,9 +26,9 @@
   "dependencies": {
     "@apollo/client": "3.5.7",
     "@hcaptcha/react-hcaptcha": "^1.4.4",
-    "@joystream/js": "1.2.0",
+    "@joystream/js": "^1.6.3",
```
thesan, line 29: It looks good but please reverse the `@joystream/js` update. I'll updated it myself because there is currently a patch which should be removed as well. I think the patch is not necessary anymore but just in case I prefer to remove it in a separate PR which can be easily reversed if something goes wrong.

thesan (changes requested):

thesan (approved):

PR #4704, opened 2023-12-22
🔍 Check transaction signer addresses in interactions tests

thesan:
I realized this could be useful while debugging #4428


kdembler (approved):
LGTM

Issue #4681, opened 2023-12-07
Enable browser extension wallet on mobile


chrlschwb:
While we worked on the RWD for Pioneer, we deliberately removed the possibility of using browser extension wallets on mobile.
Before wallet connect is implemented on Joystream, we would like to enable browser extensions on mobile for signing extrinsics.

This work should be done on top of this. https://github.com/Joystream/pioneer/issues/1794

dmtrjsg:
Needs design

thesan:
This was partially addressed in #4708.

One issue to keep in mind is that unlike the other mobile wallets, the SubWallet browser doesn't reduce the size of the viewport when the keyboard appears, as result on some of the modals to be enabled later people might have to type blind.

kdembler:
Partially done as mentioned above. Closing in favor of https://github.com/Joystream/pioneer/issues/4729

PR #4712, opened 2024-01-02
🔧 Small fixes related to wallet on mobile suppport

thesan:
-  #4681

### DIFF
--- a/packages/ui/src/common/components/Modal/Modal.tsx
+++ b/packages/ui/src/common/components/Modal/Modal.tsx
@@ -211,6 +211,8 @@ export const ModalWrap = styled.section<ModalWrapProps>`
     switch (modalMaxSize) {
       case 'm':
       case 'l':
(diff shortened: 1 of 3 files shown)


packages/ui/src/common/components/Modal/Modal.tsx
```diff
@@ -211,6 +211,8 @@ export const ModalWrap = styled.section<ModalWrapProps>`
     switch (modalMaxSize) {
       case 'm':
       case 'l':
+        // HACK: Move modals to the top on mobile because SubWallet shows the keyboard on the viewport
+        // instead of reducing the viewport size (at least on Android).
```
kdembler, line 215: Is that hack already implemented and you're only adding a comment or is that a todo?

thesan: It's already implemented in #4708. It's just that I realized that even I will probably have a hard time figuring out why I added these in a few weeks, so I added these 2 comments.

kdembler (approved):
LGTM

Issue #4667, opened 2023-11-29
Votes: Past votes page is spamming the query node


mochet:
# Problem
I have many votes and old votes, in order to release stake I must visit this page, for some period of time it displays strange behavior where the page continuously refreshes.
* During the refreshing the `"recover vote"` buttons don't function properly
* It is unclear how to stop it from refreshing except to recover the votes via the `My accounts` page 
* 
![image](https://github.com/Joystream/pioneer/assets/50830161/4dd4a24a-1c8c-4049-baeb-3e310df72981)

dmtrjsg:
cc @chrlschwb

mochet:
This behavior seems to have stopped happening

mochet:
This behavior still happens and another user reported that it only happens with >100 votes (they found this out by disabling some of their addresses)

PR #4715, opened 2024-01-04
🐛 Fix continuously refreshing pages

thesan:
- #4667

### DIFF
--- a/packages/ui/src/council/components/PastVotes/PastVotesList.tsx
+++ b/packages/ui/src/council/components/PastVotes/PastVotesList.tsx
@@ -17,12 +17,12 @@ import { PastVoteColumns } from './styles'
 
 export const PastVotesList = () => {
   const { order, getSortProps } = useSort<CastVoteOrderByInput>('createdAt')
-  const { votes, isLoading, pagination } = useMyPastVotes({ order })
+  const { votes, pagination } = useMyPastVotes({ order })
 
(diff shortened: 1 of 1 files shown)


thesan:
@mochet I haven't been able to reproduce the issue so far, mostly due to the QN being very inconsistent (it sometimes doesn't work at all and it sometimes it gets pretty fast).

When you get some time could you check if it still happens on: https://dao-git-fix-page-continuous-refresh-joystream.vercel.app/ please ?

mochet:
> @mochet I haven't been able to reproduce the issue so far, mostly due to the QN being very inconsistent (it sometimes doesn't work at all and it sometimes it gets pretty fast).
> 
> When you get some time could you check if it still happens on: https://dao-git-fix-page-continuous-refresh-joystream.vercel.app/ please ?

Unfortunately I think it only works if the votes aren't recovered so I can't really see if it has fixed or not. There is an upcoming election, so I will probably be able to test this in ~7 days or so

mochet:
> @mochet I haven't been able to reproduce the issue so far, mostly due to the QN being very inconsistent (it sometimes doesn't work at all and it sometimes it gets pretty fast).
> 
> When you get some time could you check if it still happens on: https://dao-git-fix-page-continuous-refresh-joystream.vercel.app/ please ?

I just tried it now and it appears to be fixed on the vercel link that you provided

PR #4738, opened 2024-01-16
Fix endless re-rendering of the dashboard header

### LINKED ISSUE #4724: Validator dashboard: Header causes endless re-renders

eshark9312:
closes #4724

### DIFF
--- a/packages/ui/src/validators/components/statistics/Era.tsx
+++ b/packages/ui/src/validators/components/statistics/Era.tsx
@@ -12,25 +12,25 @@ import {
   formatDurationDate,
 } from '@/common/components/statistics'
 import { DurationValue } from '@/common/components/typography/DurationValue'
+import { ERA_DURATION } from '@/common/constants'
 
 interface EraProps {
   eraStartedOn: Option<u64> | undefined
-  eraDuration: number
-  now: u64 | undefined
+  now: number
   eraRewardPoints: PalletStakingEraRewardPoints | undefined
 }
 
-export const Era = ({ eraStartedOn, eraDuration, now, eraRewardPoints }: EraProps) => {
+export const Era = ({ eraStartedOn, now, eraRewardPoints }: EraProps) => {
   const { nextReward, percentage } = useMemo(() => {
-    const nextReward = now && eraStartedOn && eraDuration - (Number(now) - Number(eraStartedOn))
-    const totalDuration = Number(eraDuration)
+    const nextReward = now && eraStartedOn && ERA_DURATION - (now - Number(eraStartedOn))
+    const totalDuration = Number(ERA_DURATION)
     const percentage = nextReward ? Math.ceil(100 - (nextReward / totalDuration) * 100) : 0
     return {
       nextReward: formatDurationDate(nextReward ?? 0),
       totalDuration: formatDurationDate(totalDuration ?? 0),
       percentage,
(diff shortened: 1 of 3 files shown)


packages/ui/src/validators/hooks/useStakingStatistics.tsx
```diff
@@ -1,14 +1,14 @@
...
 
 import { useApi } from '@/api/hooks/useApi'
-import { ERA_DURATION } from '@/common/constants'
+import { useFirstObservableValue } from '@/common/hooks/useFirstObservableValue'
 import { useObservable } from '@/common/hooks/useObservable'
 
 export const useStakingStatistics = () => {
   const { api } = useApi()
-  const activeEra = useObservable(
+  const activeEra = useFirstObservableValue(
```
thesan, line 11: Actually I think this should be a `useObservable`. It should update only every 6h. I'm thinking of even passing it to `useValidatorsWithDetails` (in a followup PR) to update the queries there too.

packages/ui/src/validators/hooks/useStakingStatistics.tsx
```diff
@@ -19,44 +19,51 @@ export const useStakingStatistics = () => {
     [api?.isConnected]
   )
 
-  const now = useObservable(() => api?.query.timestamp.now(), [api?.isConnected])
-  const totalIssuance = useObservable(() => api?.query.balances.totalIssuance(), [api?.isConnected])
-  const currentStaking = useObservable(
+  const [now, setNow] = useState(Date.now())
+  setInterval(() => setNow(Date.now()), 1000)
```
thesan, line 23: This shouldn't be here. It will cause the page to re-render maybe every seconds but maybe a lot more than that because the interval is set on every render and never cleaned

packages/ui/src/validators/components/statistics/Era.tsx
```diff
@@ -12,25 +12,25 @@ import {
...
+    const nextReward = now && eraStartedOn && ERA_DURATION - (now - Number(eraStartedOn))
+    const totalDuration = Number(ERA_DURATION)
     const percentage = nextReward ? Math.ceil(100 - (nextReward / totalDuration) * 100) : 0
     return {
       nextReward: formatDurationDate(nextReward ?? 0),
       totalDuration: formatDurationDate(totalDuration ?? 0),
       percentage,
     }
-  }, [eraStartedOn, eraDuration, now])
+  }, [eraStartedOn, now])
```
thesan, line 33:
```suggestion
  const [spentDuration, setSpentDuration] = useState<number>()
  
  const { nextReward, percentage } = useMemo(() => ({
    nextReward: whenDefined(spentDuration, (d) => ERA_DURATION - d),
    percentage: spentDuration && Math.ceil(100 * ERA_DURATION / spentDuration),
  }), [spentDuration])
  
  useEffect(() => {
    if (!eraStartedOn) return

    const interval = setInterval(
      () => setSpentDuration(Math.max(0, Date.now() - Number(eraStartedOn))),
      1000
    )
    return () => clearInterval(interval)
  }, [eraStartedOn])
```

thesan (changes requested):

thesan (approved):
LGTM!

Some data looks wrong but I'll try to debug this in another PR

PR #4745, opened 2024-01-23
change openings reward display

### LINKED ISSUE #4516: Openings: Change the way rewards are displayed
# Problem
Currently, Pioneer displays rewards for openings as follows:
image

Almost all of these rewards are basically "per day" and should just be changed to say that, because:
1. Nobody knows what the blocktime of Joystream is unless they've been around for a while. Certainly even if someone knows the blocktime, 14400 will not likely register as being 1 day.
2. There is not any significant material difference in stating that a reward is per 14450 blocks vs per 14460 blocks. It would take more than a year for it to matter.

# Solution
* For all openings, display the reward as `"Reward per day"`
* Ideally, integrate the USD market rate (but that is probably a separate matter)

vrrayz:
Fixes #4516

### DIFF
--- a/packages/ui/src/app/pages/WorkingGroups/UpcomingOpening.tsx
+++ b/packages/ui/src/app/pages/WorkingGroups/UpcomingOpening.tsx
@@ -1,4 +1,3 @@
-import BN from 'bn.js'
 import React, { memo } from 'react'
 import { useParams } from 'react-router-dom'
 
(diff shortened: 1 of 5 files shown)


packages/ui/src/app/pages/WorkingGroups/UpcomingOpening.tsx
```diff
@@ -83,8 +84,8 @@ export const UpcomingOpening = () => {
                 from={opening.expectedStart}
               />
               <TokenValueStat
-                title={`Reward per ${rewardPeriod?.toString()} blocks`}
-                value={rewardPeriod?.mul(opening.rewardPerBlock)}
+                title={'Reward per week'}
+                value={rewardPeriod?.mul(opening.rewardPerBlock).mul(new BN(7))}
```
thesan, line 88: The code now makes the assumption that `opening.rewardPerBlock` is a day , which is fine but in this case there's no need to use the api `rewardPerBlock` (because if the api returns something else than `14400` then `rewardPerBlock * 7` won't be a week anymore).

So IMO this would be a lot clearer:
```suggestion
                value={opening.rewardPerBlock.muln(A_WEEK / MILLISECONDS_PER_BLOCK}
```
Or even:
```suggestion
                value={opening.rewardPerBlock.muln(BLOCKS_PER_WEEK}
```
(With `BLOCKS_PER_WEEK` defined as a constant somewhere).

thesan (changes requested):

thesan (approved):
LGTM

PR #4750, opened 2024-01-29
cancel proposal feature

### LINKED ISSUE #4694: Proposals: Cancel proposal button missing
# Problem
There used to be a button to cancel/withdraw a proposal once it was submitted that the proposal author could then use to cancel the proposal. This is now missing for some reason. I was able to cancel the proposal via an extrinsic instead.

vrrayz:
Fixes #4694

### DIFF
--- a/packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
+++ b/packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
@@ -120,7 +120,6 @@ export const ProposalPreview = () => {
             <ButtonsGroup>
               {active?.id === proposal.proposer.id &&
                 proposal.votes.length === 0 &&
-                proposal.status !== 'cancelled' &&
                 (proposal.status === 'deciding' || proposal.status === 'dormant') && (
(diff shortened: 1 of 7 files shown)


thesan (changes requested):
Please also add a happy case test in `app/pages/Proposals/ProposalPreview.stories.tsx` just to make sure the transaction was called once and with the correct parameters

packages/ui/src/app/pages/Proposals/ProposalPreview.tsx
```diff
@@ -120,7 +120,10 @@ export const ProposalPreview = () => {
             <ButtonsGroup>
               {active?.id === proposal.proposer.id &&
                 proposal.votes.length === 0 &&
-                proposal.status !== 'cancelled' && <CancelProposalButton member={active} proposalId={proposal.id} />}
+                proposal.status !== 'cancelled' &&
```
thesan, line 123: No need for this then:
```suggestion
```

thesan (changes requested):
Appart from this LGTM, but lets wait for QA before merging

thesan (approved):

Issue #4729, opened 2024-01-12
Adjust wallet connection modal for mobile


kdembler:
Part of https://github.com/Joystream/joystream/issues/5010

Blocked by https://github.com/Joystream/joystream/issues/5039

SSIA. We should allow users to select their desired wallet on mobile.

PR #4759, opened 2024-02-01
📡 Add support for WalletConnect

thesan:
Partially address:
- #4729


thesan:
This PR is already way too big so I changed the target to a new branch `wallet-connect` branch so this can be merged before this whole feature is stable, i.e wallet connect support and better mobile support.

kdembler (commented):
I will review the code tomorrow. From using, all looks great. Only issue I've noticed is the same as in Atlas PR - disconnecting a wallet in Pioneer doesn't end WC session. If I select WC option again, my previous connection is restored and I don't have an option to connect different wallet through WC

thesan:
@kdembler 
> disconnecting a wallet in Pioneer doesn't end WC session.

I just fixed this.

FYI there was a change on this PR to show the member selection modal with the "Disconnect wallet" button when you click on "Join Now". However it was bundled with a lot of UI changes so I moved everything to #4781 (to have a more manageable PR size). So until we get to #4781, if you connect with a wallet with no memberships you can disconnect the wallet by clicking on "Add New Proposal".

packages/ui/src/accounts/model/walletConnect.ts
```diff
@@ -0,0 +1,189 @@
+import { Signer } from '@polkadot/api/types'
+import { WalletConnectModal } from '@walletconnect/modal'
+import { SessionTypes } from '@walletconnect/types'
+import { UniversalProvider } from '@walletconnect/universal-provider'
+import type { UniversalProvider as Provider } from '@walletconnect/universal-provider/dist/types/UniversalProvider.d.ts'
```
kdembler, line 5: Aren't those 2 imports the same?

packages/ui/src/accounts/model/walletConnect.ts
```diff
@@ -0,0 +1,189 @@
...
+
+      disconnectSubscription.unsubscribe()
+      this._provider?.client.off('session_delete', walletDisconnectHandler)
+      this._provider?.client.off('session_expire', walletDisconnectHandler)
+      this._provider.session = undefined
+    }
+  }
+
+  public getAccounts = async (): Promise<WalletAccount[]> => {
+    return Promise.resolve(this._accounts ?? [])
```
kdembler, line 146: Nitpick, but you don't need the `Promise.resolve`, same below

packages/ui/src/app/components/ExtensionWarning.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from 'react'
 
-import { useMyAccounts } from '@/accounts/hooks/useMyAccounts'
+// import { useMyAccounts } from '@/accounts/hooks/useMyAccounts'
```
kdembler, line 3: Can be removed?

kdembler (approved):

Issue #4672, opened 2023-12-04
Show when member was created


bedeho:
When I open a membership view panel, I want to see block and timestamp for when member was created, currently there is nothing like that.

![Screenshot 2023-12-04 at 15 22 31](https://github.com/Joystream/pioneer/assets/437292/1e3075d5-4849-4e22-9034-c1630fb141ef)

dmtrjsg:
Agreed, in the meantime it is only available in the table view: 

<img width="557" alt="Screenshot 2023-12-08 at 15 12 31" src="https://github.com/Joystream/pioneer/assets/93924525/9320351e-95a6-456c-a8a5-9ee163c0977b">


## Scope
Add date created as data and block the membership was created to the side panel (member detailed view)

PR #4762, opened 2024-02-06
Show when member was created

Ilichhh:
* [Show when member was created #4672](https://github.com/Joystream/pioneer/issues/4672)

Actually this functionality was already in the code, the display just wasn't working properly.


packages/ui/src/memberships/components/MemberListItem/components/MemberCreated.tsx
```diff
@@ -13,7 +13,9 @@ export const MemberCreated = React.memo(({ member }: MemberCreatedProps) => {
     <MemeberCreated>
       <MemberCreatedDate>{new Date(member.createdAt).toLocaleDateString('en-GB')}</MemberCreatedDate>
       {member.entry && member.entry.type !== 'genesis' && (
-        <MemberCreatedBlock>{member.entry.block.number} block</MemberCreatedBlock>
+        <MemberCreatedBlock>
+          {member.entry.block.number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',')} block
+        </MemberCreatedBlock>
```
thesan, line 18: Really nice fix. But I just realized that since the  `member.createdAt` is not correct but `member.entry.block.timestamp` is, this PR could also fix the wrong date (see https://github.com/Joystream/joystream/issues/4841), with something like:
```ts
const { entry, createdAt } = member
const block = entry.type === 'genesis' ? { number: 0, timestamp: createdAt } : entry.block
```
And then:
```tsx
      <MemberCreatedDate>{new Date(block.timestamp).toLocaleDateString('en-GB')}</MemberCreatedDate>
      <MemberCreatedBlock>{block.number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',')} block</MemberCreatedBlock>
```

thesan (approved):
:100:

PR #4763, opened 2024-02-06
🔧 Fix freeze pallet proposal

thesan:
Address https://github.com/Joystream/joystream/issues/4878#issuecomment-1867715348


kdembler (approved):
LGTM, works as expected on Nara chain

Issue #4620, opened 2023-11-07
[USERSNAP] worker tags display already abandoned roles


dmtrjsg:
**Taken at**: https://pioneerapp.xyz/#/working-groups/hr
**Fields**:
- Houston, we have a problem.: worker tags display already abandoned roles
- Comment: this should be updated whenever a worker leaves a position or is fired

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/279193a1-dd8a-478c-97df-51c5c51df480-annotated_a81ab6e3-a4c9-4ecb-a434-5223be2f17e8.png?etag=1a849c1dc8ad8960feb80acd8971e698)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/279193a1-dd8a-478c-97df-51c5c51df480-annotated_a81ab6e3-a4c9-4ecb-a434-5223be2f17e8.png?etag=1a849c1dc8ad8960feb80acd8971e698)*

**Browser**: Chrome 118 (Windows 10)
**Screen size**: 1920x1080
**Browser size**: 1920x953
**[Open #433 in Usersnap](https://app.usersnap.com/l/feedback/fa804aca-07f3-40da-9694-039da89ec707?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
Looks like query is wrong on Pioneer.

Issue #4355, opened 2023-04-18
Member roles do not update when member leaves the role


dmtrjsg:
## Context

<img width="575" alt="Screenshot 2023-04-18 at 16 57 41" src="https://user-images.githubusercontent.com/93924525/232784837-e881eaca-a0a5-4f63-9445-f6af30444219.png">

## Scope
- [ ] Fix the fetching of the member role to only show current roles

PR #4766, opened 2024-02-07
Show only current roles for the member

Ilichhh:
- [Member roles do not update when member leaves the role #4355](https://github.com/Joystream/pioneer/issues/4355)
- [[USERSNAP] worker tags display already abandoned roles #4620](https://github.com/Joystream/pioneer/issues/4620)


thesan (approved):
Perfect :raised_hands: !
Thank you

Issue #4743, opened 2024-01-22
Validator dashboard QA


thesan:
## :warning: Dependencies
The QA should start only once #4742 and https://github.com/Joystream/joystream/pull/4933 get merged and `dev` is merged into validator branch one last time.

## Context

[Designs](https://www.figma.com/file/QnjZuWYtGimTEkQlN6MNlH/Validators?type=design&node-id=100-48288&mode=design&t=nWRhVkKaUSy1Vu7k-0)

Testing environment :point_right: https://dao-git-validators-joystream.vercel.app

### Dashboard itself

This should be tested at least partially on Mainnet. There has been some back and forth on how to calculate some of the values and it's important to make sure that non what is displayed is misleading to nominators. Also this page should work well in mobile:

1. Dashboard statistic headers statistic accuracy: 
  ![image](https://github.com/Joystream/pioneer/assets/6571453/fd22acc9-b35b-40fc-bc56-e39b01666ae0)
2. Dashboard list item accuracy + interaction with the filters/ordering/pagination, etc..
   ![image](https://github.com/Joystream/pioneer/assets/6571453/407f4c41-5599-46e9-8472-1a97b03e17a7)
3. Validator cards layout + data accuracy:
   ![image](https://github.com/Joystream/pioneer/assets/6571453/1e462c13-9ab7-40e0-a32d-45c298d25cfc)
4. The "Nominate" links:
   ![image](https://github.com/Joystream/pioneer/assets/6571453/88dae741-c988-4a0b-8592-8e21589c6571)
 
### Binding validator accounts to memberships

It might be easier to test this on a playground. In that case Atlas dev should be fine: https://dao-git-validators-joystream.vercel.app/#/settings?network-config=https://atlas-dev.joystream.org/network/config.json

Regardless you'll need to create validator accounts on the network used for the tests.

1. Binding a/multiple validators to a new membership.
2. Binding a/multiple validators to an existing membership.
3. Verify and unverifiy validator accounts via the CLI (https://github.com/Joystream/joystream/pull/4933) and check their statuses on the dashboard.

ivanturlakov:
### Dashboard

Tested on https://dao-git-validators-joystream.vercel.app/ Mainnet

**✅ Dashboard statistic headers statistic accuracy**

**⚠️ Dashboard list item accuracy + interaction with the filters/ordering/pagination, etc..**
- Let us make social media icons as hyperlinks to related resources
- Search validation msg (desktop & mobile)
[...]

PR #4767, opened 2024-02-08
Address create/update validator membership QA issues

eshark9312:
- #4743

### DIFF
--- a/packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
+++ b/packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
@@ -94,7 +94,6 @@ export interface MemberFormFields {
   avatarUri: File | string | null
   isReferred?: boolean
   isValidator?: boolean
-  validatorAccountCandidates?: (Account | undefined)[]
   validatorAccounts?: Account[]
   referrer?: Member
   hasTerms?: boolean
@@ -110,7 +109,7 @@ const formDefaultValues = {
   avatarUri: null,
   isReferred: false,
   isValidator: false,
-  validatorAccountCandidates: [undefined],
+  validatorAccounts: [],
   referrer: undefined,
   hasTerms: false,
   externalResources: {},
@@ -145,14 +144,14 @@ export const BuyMembershipForm = ({
     },
   })
 
-  const [handle, isReferred, isValidator, referrer, captchaToken, validatorAccountCandidates] = form.watch([
+  const [handle, isReferred, isValidator, referrer, captchaToken] = form.watch([
     'handle',
     'isReferred',
     'isValidator',
     'referrer',
     'captchaToken',
-    'validatorAccountCandidates',
   ])
+  const [validatorAccounts, setValidatorAccounts] = useState<(Account | undefined)[]>([undefined])
 
   const validators = useValidators({ skip: !isValidator ?? true })
   const validatorAddresses = useMemo(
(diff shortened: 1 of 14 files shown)


thesan (commented):
It's good so far but I agree with @ivanturlakov that the current UX is un-intuitive. TBH it append to me several times to select a validator and wonder for a moment why the button was disabled (even though I know how the design is supposed to work). And if even @ivanturlakov got confused it means that most validator will be confused too. So I think we should change the UX so it works kind of like [the application form](https://pioneer-2-storybook-joystream.vercel.app/?path=/story/pages-working-group-workinggroup--default):
image

 By default there would be one select, under the select a "Add Validator account" button which would add other selects above itself. Something like this:
image
(oc the remove field icon can remain a X)

WDYT ?

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -177,19 +177,53 @@ export const BuyMembershipForm = ({
...
+        ...validatorAccountCandidates.slice(0, index),
+        ...validatorAccountCandidates.slice(index + 1),
+      ])
+  }
+
+  const setValidatorAccounts = (accounts: (Account | undefined)[]) => {
+    form?.setValue('validatorAccountCandidates' as keyof MemberFormFields, [])
+    accounts.map((account, index) => {
+      form?.register(('validatorAccountCandidates[' + index + ']') as keyof MemberFormFields)
+      form?.setValue(('validatorAccountCandidates[' + index + ']') as keyof MemberFormFields, account)
```
thesan, line 206: I don't think React Hook Form should be used this way (registering fields again and again). Instead why not keeping `validatorAccounts` an internal state in the component ?

```ts
const [validatorAccounts, setValidatorAccounts] = useState<(Account | undefined)[]>([])
```

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -177,19 +177,53 @@ export const BuyMembershipForm = ({
...
+    form?.setValue('validatorAccountCandidate' as keyof MemberFormFields, undefined)
   }
 
   const removeValidatorAccount = (index: number) => {
-    setValidatorAccounts([...validatorAccounts.slice(0, index), ...validatorAccounts.slice(index + 1)])
+    validatorAccountCandidates &&
+      setValidatorAccounts([
+        ...validatorAccountCandidates.slice(0, index),
+        ...validatorAccountCandidates.slice(index + 1),
+      ])
```
thesan, line 199: So here maybe you can use `toSpliced`:
```suggestion
    setValidatorAccounts((accounts) => accounts.toSpliced(index, 1))
```

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -177,19 +177,53 @@ export const BuyMembershipForm = ({
...
   const isDisabled =
     type === 'onBoarding' && process.env.REACT_APP_CAPTCHA_SITE_KEY ? !captchaToken || !isFormValid : !isFormValid
 
   const addValidatorAccount = () => {
-    if (validatorAccountCandidate && isValidValidatorAccount) {
-      setValidatorAccounts([...new Set([...validatorAccounts, validatorAccountCandidate])])
-      form?.setValue('validatorAccountCandidate' as keyof MemberFormFields, undefined)
-    }
+    setValidatorAccounts([...(validatorAccountCandidates ?? []), undefined])
+    form?.setValue('validatorAccountCandidate' as keyof MemberFormFields, undefined)
```
thesan, line 191: And here to except that the value can be set:
```ts
const addValidatorAccount = (index: number, value: Account | undefined) => {
  setValidatorAccounts((accounts) => accounts.toSpliced(index, 1, value))
}

File: packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx [hunk 4]
```diff
@@ -293,54 +327,46 @@ export const BuyMembershipForm = ({
...
+                              </TextSmall>
+                              <TextSmall error>
+                                This account is neither a validator controller account nor a validator stash account.
+                              </TextSmall>
+                            </RowInline>
+                          )}
+                        </Row>
+                      ))}
+                      <RowInline justify="end">
+                        <ButtonPrimary size="small" onClick={addValidatorAccount} className="add-button">
```
thesan, line 365: Now this can be:
```suggestion
                        <ButtonPrimary
                           size="small"
                           className="add-button"
                           onClick={() => addValidatorAccount(validatorAccount.length, undefined)}
                           >
```

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -293,54 +327,46 @@ export const BuyMembershipForm = ({
...
-                          <PlusIcon />
+                      {validatorAccountCandidates?.map((account, index) => (
+                        <Row>
+                          <RowInline>
+                            <InputComponent id="select-validatorAccount" inputSize="l">
+                              <SelectAccount
+                                id="select-validatorAccount"
+                                name={`validatorAccountCandidates[${index}]`}
+                                filter={(account) => validatorAccountSelectorFilter(index, account)}
+                              />
```
thesan, line 338: But for all this to work this should become a controlled input:
```suggestion
                              <BaseSelectAccount
                                value={validatorAccounts[index]}
                                onChange={(account)) => addValidatorAccount(index, account)}
                                filter={(account) => validatorAccountSelectorFilter(index, account)}
                              />
```

Also please don't use a static id in loops, it makes invalid html and react complains about that.

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -94,7 +94,7 @@ export interface MemberFormFields {
   avatarUri: File | string | null
   isReferred?: boolean
   isValidator?: boolean
-  validatorAccountCandidate?: Account
+  validatorAccountCandidates?: (Account | undefined)[]
```
thesan, line 97: Finally this should be removed since the state is internal to the component

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -177,19 +177,53 @@ export const BuyMembershipForm = ({
...
+    (!validatorAccountCandidates ||
+      ![...validatorAccountCandidates.slice(0, index), ...validatorAccountCandidates.slice(index + 1)].find(
+        (accountOrUndefined) => accountOrUndefined?.address === account.address
+      )) &&
+    !!validatorAddresses?.includes(account.address)
+
+  const submit = () => {
+    const validatorAccounts = (validatorAccountCandidates?.filter(
+      (account, index, self) => !!account && index === self.findIndex((t) => t?.address === account.address)
+    ) ?? []) as Account[]
```
thesan, line 220:
```suggestion
    const accounts = validatorAccounts
      .filter((account) => !!account)
      .filter((account, index, self) => self.slice(index + 1).every(({ address }) => address !== account.address)) as Account[]
```

packages/ui/src/memberships/modals/BuyMembershipModal/BuyMembershipFormModal.tsx
```diff
@@ -177,19 +177,53 @@ export const BuyMembershipForm = ({
...
+    validatorAccounts.forEach((account, index) => {
+      form?.register(('validatorAccounts[' + index + ']') as keyof MemberFormFields)
+      form?.setValue(('validatorAccounts[' + index + ']') as keyof MemberFormFields, account)
+    })
```
thesan, line 224: Then I think this should work:
```suggestion
    form?.setValue('validatorAccounts', accounts)
```

thesan (changes requested):

thesan (approved):
LGTM!

ivanturlakov:
Tested on https://dao-92au0ua0u-joystream.vercel.app/#/settings?network-config=https://37.27.43.23.nip.io/network/config.json

Thanks for the update 👍 This version looks more intuitive and user friendly
Screenshot 2024-02-14 at 15 04 33


✅ Binding 2 validators to a new membership
✅ Binding 2 validators to an existing membership
✅ Add +1 validator to an existing membership

Issue #4654, opened 2023-11-24
The working "About" tab is empty (for all working groups)


thesan:
## Context
![image](https://github.com/Joystream/pioneer/assets/6571453/38cde41f-c0b1-4955-a37d-e857f3efa9ad)

## Scope

- [ ] Show the description from the [Working groups page](https://pioneerapp.xyz/#/working-groups)
   ![image](https://github.com/Joystream/pioneer/assets/6571453/e57def7e-b1e2-4541-80c8-31733d8b24ed)

- [ ] Allow linking to this tab (and the 2 other) separately. E.g so it can be used on tooltips here: https://github.com/Joystream/pioneer/pull/4643

PR #4771, opened 2024-02-13
Working group "About" tab description and routing

Ilichhh:
- [x] Add default description for WG if metadata is not set by WG Lead
- [x] Fix broken Handbook links
- [x] Add routing to WG page

[The working "About" tab is empty (for all working groups) #4654](https://github.com/Joystream/pioneer/issues/4654)


thesan (approved):
My suggestion [here](https://github.com/Joystream/pioneer/pull/4771#discussion_r1487607750) was wrong since the `WorkingGroup` component is now actually the About tab. What you implemented instead works but it duplicates a lot of code for testing just one page.

Instead I replaced `WorkingGroup` by `WorkingGroupsModule` which works more like the original suggestion. But I wanted to make sure it worked first so I committed it myself.

The rest LGTM !

Ilichhh:
> My suggestion [here](https://github.com/Joystream/pioneer/pull/4771#discussion_r1487607750) was wrong since the `WorkingGroup` component is now actually the About tab. What you implemented instead works but it duplicates a lot of code for testing just one page.
> 
> Instead I replaced `WorkingGroup` by `WorkingGroupsModule` which works more like the original suggestion. But I wanted to make sure it worked first so I committed it myself.
> 
> The rest LGTM !

Cool, thanks!  
But you just removed some `gql.queries` that I added, and wornings on `History` and `Openings` tabs appears again. I don't think they're really needed since they contain empty data. Plus other stories also have a lot of missing gueries and there are wornings. Just want to clarify this point for better understanding for the future.

PR #4786, opened 2024-02-21
add link to staking video in nomination modal

kdembler:


packages/ui/src/validators/modals/NominatingRedirectModal/NominatingRedirectModal.tsx
```diff
@@ -7,6 +7,7 @@ import { TextBig, TextInlineBig } from '@/common/components/typography'
 import { useModal } from '@/common/hooks/useModal'
 
 import { NominatingRedirectModalCall } from './types'
+import { Link } from '@/common/components/Link'
```
thesan, line 10: The linting issue should be about the import order. I'm not sure if `yarn lint:fix` fixes it, but it's just that `@/common/components/Link` should be moved before `@/common/components/Modal`

thesan (approved):
Hopefully we can link a more adapted video soon

PR #4818, opened 2024-03-26
⚓ Integrate set era payout damping factor proposal

thesan:
Addresses #4819

~~This PR uses a type patch and schema incompatible with #4816. So it shouldn't be merged yet.~~

It now includes the changes from #4822, so #4822 should be merged first


packages/ui/src/proposals/modals/AddNewProposal/components/SpecificParameters/SetEraPayoutDampingFactor.tsx
```diff
@@ -0,0 +1,39 @@
...
+            tight
+            units="%"
+            required
+          >
+            <InputNumber
+              id="damping-factor-input"
+              name="setEraPayoutDampingFactor.dampingFactor"
+              maxAllowedValue={100}
+              placeholder="0"
+            />
```
thesan, line 33: First this field will take values between 0 and 100 is that correct ?

packages/ui/src/proposals/model/proposalDescriptions.ts
```diff
@@ -37,6 +37,7 @@ export const proposalDescriptions: ProposalDescriptions = {
     'Proposal to submit the Channel Incentives Payout payload, update the min/max claimable amounts and block/allow claiming of awarded JOYs by channels.',
   runtimeUpgrade: 'Proposal to upgrade version to the new runtime.',
   updatePalletFrozenStatus: 'Enabling and disabling various features on the blockchain level.',
+  setEraPayoutDampingFactor: 'TODO',
```
thesan, line 40: Secondly could you point me somewhere I could get a clear explanation of how will the era payout damping factor affect the validator payouts. I need this to write this short description of the proposal and I might also add a longer description to the `SetEraPayoutDampingFactor` component.

thesan (commented):
@ignazio-bovo, @kdembler:

kdembler (approved):
LGTM

PR #4829, opened 2024-04-10
⏲️ Remove the voting period from `minTriggerBlock`

thesan:


thesan:
Closing as the runtime validate that the execution block is at least: `now + grace_period + voting_period`

Issue #4823, opened 2024-04-02
Fix past council Total Spent


thesan:
The query should only return actually executed funding requests, and maybe just those approved between `$fromBlock` and `$toBlock` regardless from when they executed.

_Originally posted by @thesan in https://github.com/Joystream/pioneer/pull/4554#discussion_r1547647835_

PR #4831, opened 2024-04-11
🧾 Fix past council Total Spent

thesan:
Addresses #4823

This gets rid of the funding requests which didn't execute: rejected, slashed, failed execution, etc...

It also counts a funding requests expense in the spending of the council which approved it, regardless of when the proposal execute. If the funding requests constitutionality was to be raised, the last council would be attributed the spending. Similarly proposal which are still gracing are also included to the spending (if the proposal execution fail the spending of the council which approved it would decrease).  

Here's the result: https://dao-git-fix-past-council-total-spent-joystream.vercel.app/#/council/past-councils
Screenshot from 2024-04-11 19-18-00

@dmtrjsg do you know who could double check this ?


traumschule:
I confirmed `Spent on Proposals` is shown as 38k instead of 18.8k for term 26 starting at block 5745626 and suggest to merge this.

traumschule:
Merged in 3.9.1

Issue #4837, opened 2024-04-18
Fix the `SpecificParametersMultipleFundingRequest` test


thesan:
### Context

The `SpecificParametersMultipleFundingRequest` started failing consistently in the [Luxor release](https://github.com/Joystream/pioneer/pull/4836) after upgrading the polkadot dependencies: https://github.com/Joystream/pioneer/actions/runs/8734560529/job/23967418712

It fails on the CI only. I commented it out for now to proceed with the release: https://github.com/Joystream/pioneer/pull/4836/commits/7337a7d72dcde81370985aa159c49a1df0e8f5c6

### Scope

Figure out what's going wrong with this test and fix it.

PR #4842, opened 2024-04-24
🔥 Hotfix: Unresponsive Proxy API

### LINKED ISSUE #4839: Upgrading Polkadot dependencies break the Proxy API
Might be caused by a mismatched version of RXJS

thesan:
Fixes #4839 and #4837 (this one was due to unoptimised re-rendering in this proposal but I'm not sure why it got worse after upgrading the dependencies).


thesan:
@kdembler to demonstrate the issue I enabled the "ProxyAPI" on the `pioneer-2` Vercel deployments:
- Here's the current `main` branch: https://pioneer-2-git-main-joystream.vercel.app
- Here's this branch: https://pioneer-2-git-hotfix-luxor-app-freeze-joystream.vercel.app

The problem happens once you connect 2 or more accounts to Pioneer because it causes several queries of one type to share the same cold observable. It was probably just making the app a bit slower before Luxor, but now due to some changes in the balance object it caused an infinite loop.

I'm still not sure of why it doesn't happens on https://pioneer.joyutils.org. My guess is that the "ProxyAPI" is disabled there for some reason.

kdembler (approved):

PR #4849, opened 2024-05-16
📨 Notify users of proposal discussions

thesan:
Addresses #4281

(Also fixes some problems with the forum posts notifications).


kdembler (approved):
@thesan I have done a basic review of the code, wasn't cery thorough, but everything looks okay and I also ran a few tests and worked as expected so LGTM

PR #4859, opened 2024-07-02
Open `OnBoardingModal` for modals requiring a membership

### LINKED ISSUE #4648: Empty membership modal shown when no wallet is connected
As mentioned here https://github.com/Joystream/pioneer/pull/4641#pullrequestreview-1746485296 when opening any modal requiring a membership, Pioneer will show an empty "Select Membership" modal:

image

Instead the "Select Wallet" modal should show:

image

hlvhe:
Closes #4648

### DIFF
--- a/packages/ui/src/app/GlobalModals.tsx
+++ b/packages/ui/src/app/GlobalModals.tsx
@@ -221,7 +221,7 @@ export const MODAL_WITH_CLOSE_CONFIRMATION: ModalNames[] = [
 ]
 
 export const GlobalModals = () => {
-  const { modal, hideModal, currentModalMachine, showModal, modalData, isClosing } = useModal()
+  const { modal, hideModal, currentModalMachine, showModal, isClosing } = useModal()
   const { active: activeMember } = useMyMemberships()
   const { status } = useTransactionStatus()
   const Modal = useMemo(() => (modal && modal in modals ? memo(() => modals[modal as ModalNames]) : null), [modal])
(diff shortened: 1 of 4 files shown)


thesan (changes requested):
Hi @hlvhe. Thank you for the contribution. This approach works as expected when no wallet is connected but when a wallet is connected but no membership is selected the switch membership still should show. The way I see it the "OnBoardingModal" should not replace the "SwitchMemberModal", the 2 cases should somehow be discriminated.

thesan (commented):
Here's the flow when you don't have a wallet connected now:
[Screencast from 2024-07-29 09-31-44.webm](https://github.com/user-attachments/assets/0ffc5625-32c6-49c4-b538-178a95ad541b)

It's definitely an improvement compare to showing an empty membership switching modal. However the user still has to click the action once to pick a wallet then nothing happen until they click the action again.

Ideally they would:
1. pick there wallet
2. pick there membership
3. start the action they clicked on in the first place (e.g announce a candidacy)

To do this the "original modal" would have to be passed to the onboarding modal too, and this "original modal" should be handled there. Which might make the onboarding logic even more complex than it already is.

@kdembler WDYT ? Is this flow good enough for now ?

kdembler:
@thesan I can't get into details on this right now but if the PR behaviour is not ideal but still improves what we have currently, then that's good enough for me!

thesan (approved):

Issue #4488, opened 2023-08-08
Memberships: Update controller account tries to sign from wrong account


mochet:
## Problem
1. I am a user who wants to update my controller key on my account. I have access to my root key and want to change the controller key.
2. I go to the `My Profile` page (https://pioneerapp.xyz/#/profile/memberships) and then to `My Memberships`
3. I click the `pencil icon` to the right of my membership
4. I enter the new `controller` account and ensure that the `root` key is selected on Pioneer (it is selected by default anyway)
5. I click submit and sign the transaction--the wallet is trying to sign this transaction using the `controller` key which doesn't have access to do this extrinsic
6. The transaction fails

## Solution
* This should be signed using the root key

dmtrjsg:
@mochet thank you for raising 👍

dmtrjsg:
@chrlschwb there were a few members who struggled with this issue recently. Suggesting to prio this as part of support backlog.

PR #4916, opened 2025-11-23
use root account to change controller

traumschule:
#4488
- [x] The tx is signed with the root
- [x] Do not batch to change controller
- [ ] Update tests


traumschule:
Merged in 3.9.1

